PR #3, opened 2020-11-02
Searchable Attributes behaviors

ManyTheFish:
- Title: Searchable Attributes and Displayed Attributes behaviors
- Start Date: 2020-11-02

### DIFF
--- a/text/0002-searchable-display-attributes-behaviors.md
+++ b/text/0002-searchable-display-attributes-behaviors.md
@@ -3,7 +3,7 @@
 - specification PR: meilisearch/specifications#3
 - Meilisearch Issue: 
 
-# `searchable_attributes` & `display_attributes` behaviors
+# `searchable_attributes` & `displayed_attributes` behaviors
 
 ## First section: Feature Description and Interaction
 
--- a/text/0002-searchable-display-attributes-behaviors.md
+++ b/text/0002-searchable-display-attributes-behaviors.md
@@ -12,9 +12,9 @@ This first part has a general audience. It should be as little technical as poss
 ### Summary
 
 We wanted to change `searchable_attributes`, beacause it was confusing. We recently add wildcard in this field to select all. This last change introduced lot of edge-cases:
-- we don't know the order of searchable attributes when we use wildcards
+- we don't know the order of searchable attributes when we use wildcards.
 
-Searchable attribute have several uses:
+Searchable attributes have several uses:
 - know in which fields we can search
 - know the importance of fields for the criterion `Attributes`
 
--- a/text/0002-searchable-display-attributes-behaviors.md
+++ b/text/0002-searchable-display-attributes-behaviors.md
@@ -30,8 +30,6 @@ Is it relevant to mix this 2 information when we have wildcards?
 
 #### wildcard (*) & `searchable_attributes` order
 
-> When the value of searchable attributes is set to ["*"] the priority of the documents attributes could be different from what you expect. If the priority of your attributes matters (because you want to use the default criterion "attribute" we highly recommend you changing the default value for the priority you need. For example, if the "title" is more important than the "description" field, you should set the searchable attributes to ["title", "description"], in this order.
-
 > When the searchable attributes value is set to ["*"] the priority of the documents attributes is not defined.
 
 By default searchable attributes should be in the order of the first indexed document, for example:
--- a/text/0002-searchable-display-attributes-behaviors.md
+++ b/text/0002-searchable-display-attributes-behaviors.md
@@ -99,7 +99,35 @@ Being searchable and displayed are 2 different concept, and there should not be 
 
 Adding a new configuration `attributesPosition` that link all attributes present in a document  to a position makes it possible to keep track of the position without caring about whether a field is searchable or displayed. This also allows to keep track of all the fields that have been found in documents in the database.
 
+
+The default order is then **always** the order a field was in a document when it was first found.
+
+##### Content of `attributesPosition`
+
+The role of `attributesPosition` is to link each field to a position in a document. Hence, it contains all the attributes that have been found while indexing documents and ascociate them with a position. By default, this position is the position the field had when it was first encountered in a document.
+
+##### Addition of new fields
+
+When a new field is discovered, it is given the position it had in the document where it was found. It is ok to have two fields with the same position, and they are considered equal according to the `Attributes` criterion.
+
+e.g: In this example is makes sense that this is the default:
+
 ```json
+/// documents
+[
+	{
+		"id": 1,
+		"title": "the title"
+	},
+	{
+
+		"id": 2,
+		"titre": "le titre"
+	}
+]
+```
+```json
+/// attributesPosition
 {
 	"id": 0,
 	"title": 1,
@@ -107,7 +135,91 @@ Adding a new configuration `attributesPosition` that link all attributes present
 }
 ```
 
-The default order is then **always** the order a field was in a document when it was first found.
+##### Updating the `attributesPosition`
+
+There are 2 kinds of operations we need to consider:
+
+1) **Deletion**: `attributesPosition` are essential to the functionning of meilisearch. Their deletion should not be allowed. However they can be cleaned:
+calling `DELETE /indexes/:index_uid/attributes_position` triggers a reindexation, and only the fields present in the documents are kept. e.g:
+if my `attributesPosition` look like this: 
+```json
+/// attributesPosition
+{
+	"id": 0,
+	"title": 1,
+	"titre": 1
+}
+```
+and thoses are all of my documents are:
+```json
+/// documents
+[
+	{
+		"id": 1,
+		"title": "the title"
+	},
+	{
+
+		"id": 2,
+		"title": "the title2"
+	}
+]
+```
+
+then a call to `DELETE /indexes/:index_uid/attributes_position` will result in a re-indexation and `attributesPosition` will look like this: 
+```json
+/// attributesPosition
+{
+	"id": 0,
+	"title": 1
+}
+```
+
+2) **Updates**: As we have seen before, we don't want to allow deletion of those fields unless we are sure that this field does not exist in any documents. That would imply that trying to update the `attributesPosition` with a missing value is not allowed.
+
+Rather than returning an error, a call to `POST /indexes/:index_uid/attributes_position` will instead accept a partial dictionary, that it will merge with the current. Adding new values is allowed, missing values are ignored and existing one are updated with the new value. This results results in a user friendly experience.
+
+example:
+
+if I have this `attributesPosition`:
+```json
+/// attributesPosition
+{
+	"id": 0,
+	"title": 1,
+	"titre": 1
+}
+```
+
+and make a call to `POST /indexes/:index_uid/attributes_position` with the payload:
+
+```json
+{
+	"titre": 2,
+	"description":3
+}
+```
+
+then my `attributesPosition` will look like:
+
+```json
+{
+	"id": 0,
+	"title": 1,
+	"titre": 2,
+	"description":3
+}
+```
+
+#### Interaction with `searchableAttributes`
+
+As we've seen before, the `attributesPosition` contains an _at least_ exhaustive list of all the fields present in the documents stored in the database. This means that any existing document field that is used in the `searchableAttributes` is assured to have a position assigned. Any other field present in the `searchableAttributes` is simply ignored, for it is not present in any document. If in the future this same field appears in a document, then it will have a position associated with it, and everything will work as usually.
+
+The `["*"]` for `searchableAttibutes` and `displayedAttributes` simply means all the attributes found in `attributesPosition`, which is the same as all attributes present in the documents, as we've seen.
+
+#### (Future work) What if a really want to remove an `attributePosition`
+
+In a future work, we may allow to remove a specific `attributesPosition,` if we are **sure that it is not referenced by any documents**. Trying to remove a field that is referenced would trigger an error.
 
 ##### drawbacks:
 - new config added
--- a/text/0002-searchable-display-attributes-behaviors.md
+++ b/text/0002-searchable-display-attributes-behaviors.md
@@ -3,7 +3,7 @@
 - specification PR: meilisearch/specifications#3
 - Meilisearch Issue: 
 
-# `searchable_attributes` & `displayed_attributes` behaviors
+# `searchableAttributes` & `displayedAttributes` behaviors
 
 ## First section: Feature Description and Interaction
 
@@ -36,17 +36,17 @@ TODO
 ### Explanation
 
 For simplicity sake, we want MeiliSearch to work without configuration. To achieve this, the expected behaviour of Meiilisearch is to consider all attributes to be both `displayedAttributes` and `SearchableAttributes` by default.
-Currently, the `SearchabeAttributes` fields is order sensitive: the order of the fields in this array define the order of the fields in any documents, hence it's importance in a search with the attribute ranking rule. By default, this order is set to the order of the fields of the first document to be indexed.
+Currently, the `searchableAttributes` fields is order sensitive: the order of the fields in this array define the order of the fields in any documents, hence it's importance in a search with the attribute ranking rule. By default, this order is set to the order of the fields of the first document to be indexed.
 
 The default values for both of these attributes is `["*"]` (wildcard), to symbolize that all fields are contained.
 
 This way of doing thing comes with a few caveats, that need to be addressed here:
 
-#### wildcard (*) & `searchable_attributes` order
+#### wildcard (*) & `searchableAttributes` order
 
-The current choice of having an ordered array for `SearchableAttributes` means that when the value is set to wildcard (its default value), we loose all information about the order.
+The current choice of having an ordered array for `searchableAttributes` means that when the value is set to wildcard (its default value), we loose all information about the order.
 
-> When the searchable attributes value is set to ["*"] the priority of the documents attributes is undefined.
+> When the searchable attributes value is set to `["*"]` the priority of the documents attributes is undefined.
 
 By default searchable attributes should be in the order of the first indexed document, for example:
 
@@ -56,10 +56,10 @@ By default searchable attributes should be in the order of the first indexed doc
   "description": "it is awfull to be liked by many"
 }
 ```
-The default `searchable_attributes` order should be `["title", "description"]` where `"title" > "description"`.
+The default `searchableAttributes` order should be `["title", "description"]` where `"title" > "description"`.
 
-TBD: What happens if the order is changed and wildcard is set back to `*`?
-- undefined behaviour: the order when using `*` can't be relied upon
+TBD: What happens if the order is changed and wildcard is set back to `["*"]`?
+- undefined behaviour: the order when using `["*"]` can't be relied upon
 - some default value: which? Why?
 - Dissociate order and existence.
 
@@ -91,11 +91,11 @@ What that means is that when we index documents, we must check if a field alread
 	- same position as `title` (thus allowing attributes to exists at the same position like Algolia)
 	- added at the last position
 
-#### `searchable_attributes` & `displayed_attributes`
+#### `searchableAttributes` & `displayedAttributes`
 
 Being searchable and displayed are 2 different concept, and there should not be interactions between the two: a field can be searchable and not displayed, or displayed and not searchable, etc.
 
-#### add a new configuration `attributes_order`?
+#### add a new configuration `attributesPosition`?
 
 Adding a new configuration `attributesPosition` that link all attributes present in a document  to a position makes it possible to keep track of the position without caring about whether a field is searchable or displayed. This also allows to keep track of all the fields that have been found in documents in the database.
 
@@ -111,12 +111,12 @@ The default order is then **always** the order a field was in a document when it
 
 ##### drawbacks:
 - new config added
-- could be replaced by a small note about wildcard in `searchable_attributes`
+- could be replaced by a small note about wildcard in `searchableAttributes`
 
 
 ### Changes in meilisearch documentation
 
-Depending on what we choose, we'll have to add **default** and **wildcard** behavior of configurations, we'll potentially add a new configuration `attributes_order`. 
+Depending on what we choose, we'll have to add **default** and **wildcard** behavior of configurations, we'll potentially add a new configuration `attributesPosition`. 
 
 ## Second Section: Technical Specifications
 
--- a/text/0002-searchable-display-attributes-behaviors.md
+++ b/text/0002-searchable-display-attributes-behaviors.md
@@ -12,8 +12,8 @@
 Ever since the beginning, the design of `searchableAttributes` and `displayedAttributes` as shown flaws. The introduction of the wildcard tried to address some of those flaws, but there are still issues that this specs tries to correct.
 
 Searchable attributes have several uses:
-- know in which fields we can search.
-- know the importance of fields for the criterion `Attributes`.
+- Establish in which document fields the search can be performed.
+- Define the order and priority in which those fields are used by the criterion `Attributes`.
 
 The main issues we want to address here are:
 - The question of `searchableAttributes` order when the wildcard is used.
--- a/text/0002-searchable-display-attributes-behaviors.md
+++ b/text/0002-searchable-display-attributes-behaviors.md
@@ -30,7 +30,7 @@ Is it relevant to mix this 2 information when we have wildcards?
 
 #### wildcard (*) & `searchable_attributes` order
 
-> When the searchable attributes value is set to ["*"] the priority of the documents attributes is not defined.
+> When the searchable attributes value is set to ["*"] the priority of the documents attributes is undefined.
 
 By default searchable attributes should be in the order of the first indexed document, for example:
 ```json
--- a/text/0002-searchable-display-attributes-behaviors.md
+++ b/text/0002-searchable-display-attributes-behaviors.md
@@ -7,96 +7,116 @@
 
 ## First section: Feature Description and Interaction
 
-This first part has a general audience. It should be as little technical as possible (think user-level). This section contains 4 sub-sections:
-
 ### Summary
 
-Ever since the beginning, the design of `searchableAttributes` and `displayedAttributes` as shown flaws. The introduction of the wildcard tried to adress some of those flaws, but there are still issues that this specs tries to correct.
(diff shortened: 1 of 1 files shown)


text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+
+# `searchable_attributes` & `displayed_attributes` behaviors
+
+## First section: Feature Description and Interaction
+
+This first part has a general audience. It should be as little technical as possible (think user-level). This section contains 4 sub-sections:
+
+### Summary
+
+Ever since the beginning, the design of `searchableAttributes` and `displayedAttributes` as shown flaws. The introduction of the wildcard tried to adress some of those flaws, but there are still issues that this specs tries to correct.
```
eskombro, line 14:
```suggestion
Ever since the beginning of MeiliSearch, the design of `searchableAttributes` and `displayedAttributes` has shown flaws. The introduction of the wildcard tried to address some of those flaws, but there are still issues that this spec tries to correct.
```

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+
+Searchable attributes have several uses:
+- know in which fields we can search.
+- know the importance of fields for the criterion `Attributes`.
```
eskombro, line 18:
```suggestion
- Stablish in which document fields the search can be performed.
- Define the order and priority in which those fields are used by the criterion `Attributes`.
```

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+
+The main issues we want to adress here are:
```
eskombro, line 20:
```suggestion
The main issues we want to address here are:
```

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+- The question of `searchableAttributes` order when the wildcard is used.
```
eskombro, line 21:
```suggestion
- The behavior of `searchableAttributes` order when the wildcard is used.
```

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+- The fact that already known attributes (e.g from facets), are not indexed correctly is put before the documents (bug), need to be taken into account in the design.
```
eskombro, line 22:
```suggestion
- The fact that already known attributes (e.g from facets), are not indexed correctly if the `searchableAttributes` setting is defined before the documents indexation (bug), needs to be taken into account in the design.
```

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+
+Is it relevant to mix this 2 information when we have wildcards?
+
+### Explanation
+
+#### default values
+
+`searchable_attributes`: *   
+`displayed_attributes`: *
```
eskombro, line 31: We should choose a consistent syntax: `searchableAttributes` or `searchable_attributes` :)

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+
+#### wildcard (*) & `searchable_attributes` order
+
+> When the searchable attributes value is set to ["*"] the priority of the documents attributes is undefined.
```
eskombro, line 35: We should choose a consistent syntax: `*` or `["*"]` :)

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+
+what is the behavior for a field that:
+- > **is in `searchable_attributes` and in `displayed_attributes`:**
+  > meilisearch will use the field for a search and the field will be returned in search results
+- > **is in `searchable_attributes` but not in `displayed_attributes`:**
+  > meilisearch will use the field for a search but the field will **not** be returned in search results
+- > **is not in `searchable_attributes` but in `displayed_attributes`:**
+  > meilisearch will **not** use the field for a search but the field will be returned in search results
+- > **is not in `searchable_attributes` and not in `displayed_attributes`:**
+  > meilisearch will **not** use the field for a search and the field will **not** be returned in search results. The field is unused but should be kept in case of it is added in `searchable_attributes` or `displayed_attributes`.
```
eskombro, line 58:
```suggestion
  > meilisearch will **not** use the field for a search and the field will **not** be returned in search results. The field is unused but should be kept in case of is added in `searchable_attributes` or `displayed_attributes`, or present in a new document.
```

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+
+#### add a new configuration `attributes_order`?
+
+This new config may define the order of attributes for the criterion `Attributes`, no wildcard allowed.
```
eskombro, line 62:
```suggestion
This new Setting may define the order of attributes for the criterion `Attributes`, no wildcard allowed.
```

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+and `attributes_order` is: `["desc", "id", "title"]`
+The searchable attributes order would be:
+`"desc" > "title"` (`"id"` is not searchable so is ignored by the criterion `Attributes`).
+
+`attributes_order` is always exhaustive, If a new field is added, it is "pushed" at the end of the list.
+
+##### drawbacks:
+- new config added
+- more complexity
+- could be replace by a small note about wildcard in `searchable_attributes`
```
eskombro, line 73:
```suggestion
- could be replaced by a small note about wildcard in `searchable_attributes`
```

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,123 @@
...
+
+If the feature requires additions to the documentation or if sections of the documentation need to be updated because of this feature, it should be mentioned here. It's the role of the documentation team to point out the sections of the documentation that need to be updated. -->
+
+## Second Section: Technical Specifications
+
+<!-- This section has a much narrower audience: the developer that will implement the feature. Its goal is to make it as clear as possible to develop the feature, share knowledge, and think about the possibilities. -->
+
+### Architecture
+
+This change will impact **settings**, the **indexer** and the **search-engine** (`Attribute` criterion).
```
eskombro, line 107:
```suggestion
This change will impact **settings**, the **indexer** and the **search-engine** (`Attributes` criterion).
```

text/0002-searchable-display-attributes-behaviors.md
eskombro: This idea needs some development, it isn't clear at all what it means to replace this with a 'small note' :)

eskombro (changes requested):
As I was reading, added some suggestions for making the spec more clear 👍

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,116 @@
...
+`attributes_order` is always exhaustive, If a new field is added, it is "pushed" at the end of the list.
+
+##### drawbacks:
+- new config added
+- more complexity
+- could be replace by a small note about wildcard in `searchable_attributes`
+
+#### Changes in meilisearch documentation
+
+Depending on what we choose, we'll have to add **default** and **wildcard** behavior of configurations, we'll potentially add a new configuration `attributes_order`.
```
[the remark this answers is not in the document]
eskombro: I would like us to consider the possibility of removing the wildcard, as it seems "simple" for usage at first sight, but introduces complexity as behavior is not clear and needs to be documented/learned, and is subject to confusion.

This also would make it possible to use just one setting, stating CLEARLY that the order MATTERS! So it could solve both issues without adding a new setting!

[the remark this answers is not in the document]
MarinPostma: This is not an option. We decided to add the wildcard because without it we needed acceptNewField whose use needed an understanding of how meilisearch used new fields and was not clear at all

text/0002-searchable-display-attributes-behaviors.md
MarinPostma: domumenting strange behaviour != fixing strange behaviour, this is a non-solution to me. @bidoubiwa will tell you that people don't really read the doc in depth the first time around. What we want is discoverability + intuitiveness

text/0002-searchable-display-attributes-behaviors.md
bidoubiwa: I think you meant `Establish` @eskombro

eskombro: Danke ;)

text/0002-searchable-display-attributes-behaviors.md
bidoubiwa: maybe the capital letter on the setting is not needed? Apart from that, I agree.

text/0002-searchable-display-attributes-behaviors.md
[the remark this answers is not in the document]
eskombro: The more I reflect on this, the more I feel the need to create `attributes_order` as in this proposition. I would try hard to avoid adding a new setting, but it seems to be the only way (for now) to offer logic, consistent and clear behavior for this issue.

text/0002-searchable-display-attributes-behaviors.md
bidoubiwa: Should be lowercase if put between backticks

text/0002-searchable-display-attributes-behaviors.md
bidoubiwa: I think you tend to read the doc about a specific subject if you think you don't understand it. In the case of `searchableAttribute` the meaning is very clear in its first use `in what attributes will meilisearch search`. Because that meaning is so intuitive, you wouldn't guess that there is a second one to look after in the documentation.

So even if they in fact read the documentation, they would probably just read over that part because they think they already understand it. 

This is speculation of course, maybe this is not how users read. This is how I read and how I think people read.

text/0002-searchable-display-attributes-behaviors.md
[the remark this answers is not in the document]
eskombro: > I think that MeiliSearch being easy to use is more about the fact that you have very little configuration to do to make it work in most case than the fact that we have "few settings".

I think it is a mix of both

text/0002-searchable-display-attributes-behaviors.md
eskombro: Oh, men, I am in your team now, @MarinPostma and @bidoubiwa !

text/0002-searchable-display-attributes-behaviors.md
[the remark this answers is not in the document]
curquiza: I thought a lot about this, and I also feel really concerned about the implicit-and-not-obvious role of `searchableAttributes`, i.e the importance order of the fields.

And in this case, adding another settings might be the best option to manage the importance of the fields during the search.

Let's call this new setting `XXX`.

Here what we should be aware about if we add this new settings:
- Adding a new setting is adding an obstacle for the user if there are too many settings, even if each settings are explicit and well documentated. I had bad experience the first time I tried Algolia and I saw all the possibilities. No need to detail my experience with Elastic even if all the Elastic settings are really accurate and the docs not that bad. So, yes, adding a settings is adding complexity from my POV. We should be carefull about settings addition, but it does not mean not adding another settings at all. If `XXX` is needed, we should add it.
- The name of `XXX` should be really accurate
- We should anticipate, at the maximum, all the behaviors between `searchableAttributes` and `XXX`. What's about if there are less fields in `searchableAttributes` than in `XXX`?
- We should anticipate all the behavior possibilities when the users try to update `XXX`: if the list of attributes is not exhaustive, does MeiliSearch return an error or do an implicit behavior?
- How the `XXX` is going to be filled if the users add documents with new fields?

If we don't add `XXX` the docs should be really well explained and the users should be able to understand as soon as possible the `searchableAttributes` settings allow them to do two things.

The more I'm thinking about this, the more I would like to add a new setting, but I don't have any strong position about this. And if we add this new settings, we should be careful about a lot of behaviors.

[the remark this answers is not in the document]
curquiza: And we definitely need to talk directly about this to take a decision: adding a new settings or not 🙂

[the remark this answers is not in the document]
eskombro: Seems like time has come for a meeting! :)

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,255 @@
...
+Being searchable and displayed are 2 different concept, and there should not be interactions between the two: a field can be searchable and not displayed, or displayed and not searchable, etc.
+
+#### add a new configuration `attributesPosition`?
+
+Adding a new configuration `attributesPosition` that link all attributes present in a document  to a position makes it possible to keep track of the position without caring about whether a field is searchable or displayed. This also allows to keep track of all the fields that have been found in documents in the database.
+
+
+The default order is then **always** the order a field was in a document when it was first found.
+
+##### Content of `attributesPosition`
```
bidoubiwa, line 105: The absence of a field in `searchableAttributes` does not impact what will be added to `attributesPosition`?

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,255 @@
...
+then a call to `DELETE /indexes/:index_uid/attributes_position` will result in a re-indexation and `attributesPosition` will look like this: 
+```json
+/// attributesPosition
+{
+	"id": 0,
+	"title": 1
+}
+```
+
+2) **Updates**: As we have seen before, we don't want to allow deletion of those fields unless we are sure that this field does not exist in any documents. That would imply that trying to update the `attributesPosition` with a missing value is not allowed.
```
bidoubiwa, line 178: what is the maximum number? Can it have negative values? Can there be gaps between positions?

text/0002-searchable-display-attributes-behaviors.md
```diff
@@ -0,0 +1,255 @@
...
+- To make sure that the `searchableAttributes` and `displayedAttributes` are in sync, the must come from the same source of information; this should be the field_position.
+ TBD: What should be the position of `titre`?
+	- same position as `title` (thus allowing attributes to exists at the same position like Algolia)
+	- added at the last position
+
+#### `searchableAttributes` & `displayedAttributes`
+
+Being searchable and displayed are 2 different concept, and there should not be interactions between the two: a field can be searchable and not displayed, or displayed and not searchable, etc.
+
+#### add a new configuration `attributesPosition`?
```
bidoubiwa, line 98: Are we sure about this name? Some other suggestions: 

- `attributesImportanceOrder`
- `attributesRelevancyOrder`
- `attributesOrderOfRelevancy`
- ...

I like a clear name. I feel like `attributesPosition` is not clear in it's utility.

bidoubiwa (commented):
I really like the details for `attributesPosition`, I have the feeling that you thought about a lot of possible problems. The fact that no attribute can be removed and that it does not work with an array but an object makes for a tool that has way less possible problems.

text/0002-searchable-display-attributes-behaviors.md
MarinPostma: no it doesn't, all fields that are present in any documents are given an attribute position. This way you can add or remove an attribute position, and this will not impact it's position in the documents. No data is ever lost.

text/0002-searchable-display-attributes-behaviors.md
MarinPostma: this is TBD, it will probably be bound by the type of the value we use to store the information, this is an implementation detail I don't have the answer to now. Gaps are tolerated.

text/0002-searchable-display-attributes-behaviors.md
MarinPostma: After some research on the technical feasibility of the `attributePostition` (whatever we decide to call it in the future), I have found that it is a little bit more complicated than anticipated. 

We don't currently support attributes having the same importance ([algolia does it](https://www.algolia.com/doc/guides/managing-results/must-do/searchable-attributes/how-to/configuring-searchable-attributes-the-right-way/#searchable-attributes-with-the-same-priority)). This is a pre-requisite for this to work, and this is not easy to do with the current version (nothing impossible in my mind though).

Now, we will probably want to support attributes with the same priority at some point in time (I see no reason why we wouldn't) and this should be taken into account in the current design AND future design of MS (@Kerollmops).

### Technical reason

In order to know what position a field is in, the `DocIndex` struct contains an `attributes` which instead of being a `FieldId` is the position. This system is designed with the the position being isomorphic to the attibute id, so one position maps to only one attribute. If we allow multiple attributes to have the same importance, then we can't rely on that.
we could either:
- store the position int he `DocIndex`
- store the fieldId in the `DocIndex` and map it to a position at a later time

MarinPostma:
After some research on the technical feasibility of the `attributePostition` (whatever we decide to call it in the future), I have found that it is a little bit more complicated than anticipated. 

We don't currently support attributes having the same importance ([algolia does it](https://www.algolia.com/doc/guides/managing-results/must-do/searchable-attributes/how-to/configuring-searchable-attributes-the-right-way/#searchable-attributes-with-the-same-priority)). This is a pre-requisite for this to work, and this is not easy to do with the current version (nothing impossible in my mind though).

Now, we will probably want to support attributes with the same priority at some point in time (I see no reason why we wouldn't) and this should be taken into account in the current design AND future design of MS (@Kerollmops).

### Technical reason

In order to know what position a field is in, the `DocIndex` struct contains an `attributes` which instead of being a `FieldId` is the position. This system is designed with the the position being isomorphic to the attibute id, so one position maps to only one attribute. If we allow multiple attributes to have the same importance, then we can't rely on that.
we could either:
- store the position int he `DocIndex`
- store the fieldId in the `DocIndex` and map it to a position at a later time

bidoubiwa:
Is it not possible that we send an error message to the user if they use two times the same position ?

```json
{
 "id": 1,
 "title": 1
}
```
Would result in an error looking like this
```
Two attributes can not share the same position. Each position must be held by one attribute
```

I still think this is a better solution than the confusing setting we have now.

qdequele:
I have the impression that `displayed attributes` doesn't have a very large place in this spec. Would it be possible to remove it from the name and motivation. Make this spec really focus on the indexed fields and their order of priority?

MarinPostma:
> I have the impression that `displayed attributes` doesn't have a very large place in this spec. Would it be possible to remove it from the name and motivation. Make this spec really focus on the indexed fields and their order of priority?

MarinPostma:
> Is it not possible that we send an error message to the user if they use two times the same position ?
> 
> ```json
> {
>  "id": 1,
>  "title": 1
> }
> ```
> 
> Would result in an error looking like this
> 
> ```
> Two attributes can not share the same position. Each position must be held by one attribute
> ```
> 
> I still think this is a better solution than the confusing setting we have now.

This could be a workaround, but we now need to find an acceptable behaviour for when there is a new attribute found in an occupied position:
```{
 "id": 1,
 "title": 1
}, 
{
 "id": 1,
 "titre": 1
}
```

What should `titre` position be?

bidoubiwa:
### Pushed at the end
I would suggest adding them to the last position by default. If the user wants to change its attributes order he would be able to clearly understand by looking at the settings.

It is less sexy than adding the attribute at the position it was found but..
Jsons documents where users sort their attribute in their order of importance is not something common. I'm not sure any user would do that by default unless they are aware of how we handle their json documents. 
In which case, they are closer to resolving the issue than someone that is not aware of that and sent his documents with the attributes in whatever order he randomly made. Those are the users we're targeting with this setting. 

Users that will go to the settings, look at them, and observe that in `attributesRelevancyOrder` their attributes are in completely bad order, will realize really intuitively that there is a problem.
```json
{
  "attributesRelevancyOrder": {
      "side_node": 1,
      "id": 2,
      "boring_info": 3,
      "description": 4,
     "title": 5,
     "really_rare_attribute": 6
   }
}
```

### Keeping the smart positioning
But, even if a majority of us prefer the newfound attributes to be in their correct position we could also make way for them by pushing all attributes after the newfound by one position.

```json
{
  "attributesRelevancyOrder": {
      "side_node": 1,
      "id": 2,
      "boring_info": 3,
      "description": 4,
     "title": 5,
   }
}
```

```json
{
  "attributesRelevancyOrder": {
      "side_node": 1,
      "id": 2,
       "really_rare_attribute": 3,
      "boring_info": 4,
      "description": 5,
     "title": 6,
   }
}
```

and in the rare case of `titre` where he's competing for the last position, we could decide that an older attribute is more important than an newer. In which case it would become 
```json
{
 "id": 1,
  "title": 2,
  "titre": 3
}
```

eskombro:
I think this should be fixed by giving each attribute a priority value (0-100 for example), instead of a position. Then MeiliSearch will order by himself those attributes by priority and apply them, high priority first, low priority last. In case of collision, the first attribute ever introduced by the user is applied before an attribute that came later.  I don't know about how feasible this is, but behavior seems easy to handle and consistent, and collisions have a defined behavior, very easy to avoid. WDYT?

qdequele:
For my part, I don't like at all the proposal to assign a number to each attribute. I find that in the best of worlds it looks good but it can quickly get messy. 

An example of where it's not too complicated.

```
{
      "side_node": 1,
      "id": 2,
      "boring_info": 3,
      "description": 4,
     "title": 5,
     "really_rare_attribute": 6
}
```

An example or it is a little less clear. It is also possible to imagine it with many more fields. 

```
{
      "side_node": 4,
      "id": 1,
      "boring_info": 12,
      "description": 2,
     "title": 56,
     "really_rare_attribute": 12
}
```

In the case of:
```
{
      "side_node": 1,
      "id": 2,
      "boring_info": 3,
      "description": 4,
     "title": 5,
     "really_rare_attribute": 6
}
```

If we want to add `author` with a priority lower than `side_node` but higher than `id`. This means that we have to give it the number 2 and then push everything else lower. Super boring in many cases. 

In most cases, introducing the possibility to put unordered attributes is not enough? Should we really allow to put several attributes at the same level? Is it a requested feature?

If we must, I find that finally, the solution to put the names of attributes of the same level in the same string is ok. e.g. ["id", "title", "description", "author,producer", "note"]. It's not the best but it works. At least it's clear!

Kerollmops:
I like the solution introduced by Charlotte where the user can specify the order of the attributes by specifying the index of the attribute but I find it a little bit complex, even if it doesn't.

The solution of Samuel is kind of the same, it makes me think of an Elastic Search API where you apply a weight to each document attribute.

I prefer the initial solution to use a simple array because it brings simplicity and correctness by design, it is impossible for a user to introduce a wrong value in the system, the order of the names of the attributes is total and valid at all times. I remember seeing this kind of pattern and technics in [The Design of Everyday Things](https://en.wikipedia.org/wiki/The_Design_of_Everyday_Things) (that I highly recommend).

MarinPostma:
quoting "The design of everyday things":

>•Affordances are the possible interactions between people and the environment. Some affordances are perceivable, others are >not.
> 
> •Perceived affordances often act as signifiers, but they can be ambiguous.
> 
> •Signifiers signal things, in particular what actions are possible and how they should be done. Signifiers must be perceivable, else they fail to function.
> 
> In design, signifiers are more important than affordances, for they communicate how to use the design. A signifier can be words, a graphical illustration, or just a device whose perceived affordances are unambiguous. Creative designers incorporate the signifying part of the design into a cohesive experience. For the most part, designers can focus upon signifiers.

Particularly "Signifiers signal things, in particular what actions are possible and how they should be done. Signifiers must be perceivable, else they fail to function.". In that sense, `SearchableAttributes` does not communicate any information about the order of the attributes, according to Don Norman, what @Kerollmops suggests is, by definition, not a good design.

Kerollmops:
JSON cannot represent objects where the key is a number and if we change them into strings it could confuse users, seeing a number as a string is not a good design choice.

MarinPostma:
@Kerollmops you are right. Though I still think we should have two settings. How about that:
- we keep `SearchableAttributes` with a default value of `["*"]`
- we represent `AttributeOrder` with what you proposed (`["id", "title", "description", ["author", "producer"], "note"]`), to get an exhaustive and ordered list of attributes importance.

Doesn't that address all of the issues we have thought of so far?
- clear design
- obvious semantic
- no undefined behavior
- exhaustive list of all the fields known to the engine
- no place for errors

The corner cases are the same as with the dictionary, and are therefore treated in the same manner.

bidoubiwa:
I agree with @MarinPostma. I just want some clarity on this point
> no place for errors

What happens if the users omit an existing field in `attributeOrder`? I think we should return an error but I'm not sure if this is what you suggest when you say:

> The corner cases are the same as with the dictionary, and are therefore treated in the same manner.

MarinPostma:
@bidoubiwa yes this is what I mean, when a I no error I mean no way for the user to commit one without knowing

qdequele:
We organized an internal meeting, were present @Kerollmops, @MarinPostma, @curquiza, @bidoubiwa , @eskombro.

At the moment we cannot agree on a final solution. Two proposals have been hard discussed: 

### Proposals

#### Proposal number 1:

Split the two problems of defining searchable attributes and defining the order of priority of the attributes into two separate settings parameters. 

```json 
{
    "SearchableAttributes": ["*"],
    "AttributeOrder": ["id", "title", "description", "author", "producer", "note"]
}
```

Pro: 

- Both settings have concrete names.
    - SearchableAttributes: All searchable attributes. "*" means all.
    - AttributeOrder: The list of attributes in order.

Cons:

- A much greater source of error.
- More settings, therefore, more complexity. 


#### Proposal number 2:

Keep, as now, only one setting for both. 

```json 
{
    "SearchableAttributes": ["id", "title", "description", "author", "producer", "note"],
}
```

Pro: 

- Only one setting attribute, so a less complicated API. 
- No breaking change. 

Cons:

- Unclear name because it does the list of searchable attributes and the order at the same time. **Changeable?
- The case of the default value `*` is not explicit at all. 

### Decision

For the moment, we are still on solution number 2. We want to give ourselves more time to get more user feedback, do a usage analysis, and do more documentation.

Todo:
- Add more documentation in the references part of the documentation about the ability to choose the searchable attribute order and its importance. 
- Create a page "Understand the MeiliSearch results" that talks about the search attribute order. 
- Create a page "How to order the attribute priorities" that also talks about the search attribute order. 
Contact some users and understand if they knew how to change the attribute order, find the information on our documentation, and think it was simple.

PR #7, opened 2020-11-12
Review process

MarinPostma:
Proposition for a specification review process

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,3 +64,12 @@ Some aspects of the development will necessitate special care, they should be po
 ## Third Section: Future possibilities
 
 This last section talks about what has been thought of related to this issue, but has been decided not to be done now, and what it means regarding the feature at hand.
+
+## Review process
+
+The specification review is a two steps process:
+
+- 1) When the specification PR is opened (as draft), only ideas are discussed, until an agreement is found on the substance.
+- 2) When an agreement on the substance has been found, the PR is opened for review, and change requests to the shape are accepted.
+
+This prevents back and forth discussions between substance and shape, and keeps discussions organized and focused.


README.md
```diff
@@ -64,3 +64,12 @@ Some aspects of the development will necessitate special care, they should be po
 ## Third Section: Future possibilities
 
 This last section talks about what has been thought of related to this issue, but has been decided not to be done now, and what it means regarding the feature at hand.
+
+## Review process
+
+The specification review is a two steps process:
+
+- 1) When the specification PR is opened (as draft), only ideas are discussed, until an agreement is found on the substance.
+- 2) When an agreement on the substance has been found, the PR is opened for review, and change requests to the shape are accepted.
```
qdequele, line 73:
```suggestion
- 2) When an agreement on the substance has been found, the PR is opened for review, and change requests to the shape are accepted.
- 3) If any point is contested, the debate should not be about PR. It is therefore the duty of the organizer to call a meeting to discuss it. He will be responsible for writing in response to the thread the decisions taken. 
```

MarinPostma: I agree with @bidoubiwa, this point is general

bidoubiwa (approved):

eskombro (commented):
In my personal opinion, this is just a theoretical division. The form is the content, and the content is the form. I don't really think that in real life we can decide to discuss one while ignoring the other one. 

A great example just happened in the comment from @bidoubiwa right before, in this same PR, before this review:  she isn't sure about what @qdequele suggestion means. She suggests a modification of the sentence formulation "the form" but what's really being discussed is the meaning "the content/substance"

I know this creates a lot of back and forth movements, but from my humble point of view, this is a problem coming from the tools we use and in which we decide to discuss on. For me, both discussions (form and content) should be parallel and not linear/incremental, but definitely we could separete them :)

MarinPostma:
Is something is ambiguous, then this is substance.

qdequele:
I made another proposition in #9

qdequele:
@MarinPostma is it ok to close your PR and focus on #9?

PR #8, opened 2020-11-16
Remove frontend in production

MarinPostma:


curquiza (approved):
👍

Kerollmops (approved):

LegendreM (approved):
Ok for me, just proposing an
alternate solution with drawbacks

text/0004-optional-frontend.md
```diff
@@ -0,0 +1,45 @@
...
+
+In most production cases, there is no need for this frontend, and exposing it is a bad practice. We want to provide an opt-out option for this feature.
+
+### Prior Art and R&D
+
+N.A
+
+### Explanation
+
+On MeiliSearch startup, we would like to pass a flag `--disable-web-gui` that disables the web user interface.
```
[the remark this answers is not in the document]
MarinPostma: What do we do then? @curquiza @LegendreM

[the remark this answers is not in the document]
curquiza: I incline to the @LegendreM suggestion that automatically disables it when the production mode is set.  Since exposing the front-end is a bad practice in production, disable it automatically could help our users, otherwise all the MeiliSearch production should have both settings: `--env production` and `--disable-web-gui`. `--env production` should be enough.
Plus, ut would enforce users to use the production mode and so that we could know who is really using MeiliSearch in production 🙂

text/0004-optional-frontend.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Impact on documentation
+
+This new feature adds a new flag to the binary, that needs to be documented [here](https://docs.meilisearch.com/guides/advanced_guides/configuration.html).
```
bidoubiwa, line 28: It should also be explained:
- In the [environment documentation](https://docs.meilisearch.com/guides/advanced_guides/configuration.html#environment) 
- In the [web interface documentation](https://docs.meilisearch.com/guides/advanced_guides/web_interface.html#web-interface)

eskombro:
I would like to give my POV about this

I think I agree 100% about the fact that this 'front-end' tool (dashboard, or whatever we call it) shouldn't be available by default in the production environment, and therefore I agree with the main point of this spec.

Nevertheless, I don't see why I, as a user, can't enable it if that is what I want. So I would propose to disable it by default in production (as you propose), but to leave an option (for example a flag) to enable it explicitly :)

text/0004-optional-frontend.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Second Section: Technical Specifications
+
+### Architecture
+
+N.A
+
+### Implementation Details
+
+Pass an argument to `create_app` to conditionally enable the frontend service.
```
curquiza, line 41: Is this still needed? Or is this referred to the @eskombro comment (https://github.com/meilisearch/specifications/pull/8#issuecomment-729676988)

MarinPostma: This is just implementation detail, it only concerns the core team 😉

MarinPostma: So not related

curquiza: My baaaaad 😇

curquiza (approved):

text/0004-optional-frontend.md
```diff
@@ -0,0 +1,48 @@
...
+
+## First section: Feature Description and Interaction
+
+### Summary
+
+For testing purposes, MeiliSearch is shipped with a frontend that is served at the server's root. This provides the user with a basic GUI to perform searches on his documents. While this is useful in a development environment, this is not necessarily desirable in a production environment. This specification proposes to remove this frontend in a development environment.
+
+### Motivation
+
+This frontend was developed for testing purposes, and it should not be present in a production environment.
```
eskombro, line 16:
```suggestion
This frontend was developed for testing purposes, and it should be disabled by default in a production environment.
```

text/0004-optional-frontend.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Prior Art and R&D
+
+N.A
+
+### Explanation
+
+When the option `--env` or `MEILI_ENV` is set to `production`, the frontend is disabled.
+
```
eskombro, line 25:
```suggestion
When the option `--env` or `MEILI_ENV` is set to `production`, the frontend is disabled unless the option `--enable-gui` or `MEILI_ENABLE_GUI` is set to `true`.
```

eskombro (commented):
I updated my proposal as concrete changes to the specs.

This is just an idea for consideration

text/0004-optional-frontend.md
MarinPostma: Can't we wait until someone complains until we introduce a new option? I like @LegendreM because it doesn't introduce any change to the interface, and t probably fits in 99,9% of the usecases, wdyt @eskombro, @curquiza ?

curquiza: I'm not against Sam's suggestion, but on the other hand, I also think if a user wants the interface in production, this is not production, this is for dev purposes.
So, I'm more on Marin's side: wait for complaints and specific use-cases before adding this option. But this is just my opinion.

MarinPostma: No matter what, adding an option does not change anything to what is done here, so if we all agree on this, we can merge this spec. If adding an option is needed, it can be done in another RFC, without conflicts with this one.

Kerollmops:
Was wondering if it would be a good idea to also have a flag to enable (force) the web interface for people who would like to have this web interface in production. What do you think?

text/0004-optional-frontend.md
eskombro: > Can't we wait until someone complains

Ok, I am mainly saying this thinking on Sandbox users first (always transposing that to any other possible automation tools, like DO Image, Saas, and this kind of things) where the USER is not the OWNER. 

- I can set a production environment to give production-ready instances to someone else, it doesn't suit me to not have the dashboard as it is the only visible+interactive part of the running MeiliSerch instance in the browser (doesn't require a terminal or any other normal tool).
- I do a demo, I want to set it In production, I can still use that web GUI. 
- I do a tutorial, I want it activated to have a GUI.

Why are you taking this from me? 😢 

[ok @MarinPostma , someone has complained now haha ;) ]

eskombro: I am only asking for this if there is no implementation complexity

eskombro:
> Was wondering if it would be a good idea to also have a flag to enable (force) the web interface for people who would like to have this web interface in production. What do you think? Maybe not in this spec.

I think this is the same proposition I did just before (unless I misunderstood it)

text/0004-optional-frontend.md
MarinPostma: What's the difference between prod and dev env again?

eskombro: Actually checking the [docs](https://docs.meilisearch.com/guides/advanced_guides/configuration.html#environment) it is not quite much of a difference. But I don't know what will be built on top of that.

MarinPostma: @eskombro Is it ok for you to switch to a dev env, until we figure out what we want to do with this settings?

eskombro: Yes, let's not block this, go ahead :)

MarinPostma:
I went a bit too fast reverted this merge to make edits...

PR #13, opened 2020-11-26
Cancel pending tasks

qdequele:


qdequele:
@MarinPostma When you have the time could you please complete the technical part?

0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,151 @@
...
+
+Other proposition:
+
+- Status Code: 200
+- Response: 
+```json
+{
+    deletedUpdate: 12
+}
+```
```
curquiza, line 108: If I understand well, this action is not asynchronous, like an index deletion. So I would rather stay consistent with the `index` routes behavior, which is `204` and no body answer.

0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,151 @@
...
+
+- Add on the guide the mention of the `aborted` status. 
+- Change on the guide where we explain that it's not possible to delete an update. 
+- Add the two methods described in the HTTP API part of the documentation.
+
+## 2. Technical Specifications
+
+### Architecture
+### Implementation Details
+### Corner Cases
```
MarinPostma, line 146: don't forget to remove the value in the update after it's aborted

0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,151 @@
...
+}
+```
+
+or
+
+```json
+{
+    deletedUpdates: [12, 13, 14, 15]
+}
+```
```
curquiza, line 130: Same here as my comment above: I would rather the `204`

curquiza (commented):
Looks like a great and requested feature 👍

0013-delete-enqueued-updates.md
curquiza: (Sorry if it was not the step of the process to give my opinion. Anyway, you can add on the spec "the `204` solution" is the consistent solution with `indexes` routes 😇)

qdequele: I proposed two solutions. The first one is like you said a `204` with no content. And I think it's the best choice for this route. But the second proposition is made to be consistent with the clear all updates proposition. See the comment on your second comment.

0013-delete-enqueued-updates.md
qdequele: I'm asking myself about the fact the user could need to have the list of deleted update or the number of deleted updates. Do you think it's not useful and too much? Should we choose to return 204 in every delete/clear request? Perhaps :D

curquiza: Following this StackOverflow response, there is no hard standardization for the `DELETE` route of a RESTful API: https://stackoverflow.com/questions/25970523/restful-what-should-a-delete-response-body-contain/25970628
So we can do what we want as long as it is consistent with our other API routes.

However, IMO it's too much for a MVP, not sure this is really useful to know the list (or the total number) of the deleted objects. It could have been useful maybe for the documents, but `documents` routes are asynchronous tasks, so this is not the same context.
For the updates, I think this is too much for a first introduction, especially because it would involve changing the `DELETE indexes/` route to make it consistent, for a user need we are not even sure. Maybe should we wait for the user requests to know their use case? This way we could adapt MeiliSearch in the best possible way.

qdequele:
Must reword every `delete` by `abort`.

0013-delete-enqueued-updates.md
qdequele: 204 No Content for everything! 👍

0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,151 @@
...
+# Delete enqueued updates
+
+## 1. Feature Description and Interaction
+
+### Summary
+
+Allow to delete messages in the update queue or to empty the entire queue. A deleted update will have the status `aborted`.
+
+Questions not answered:
+- Should we delete in-process updates? If yes, what the default behavior for the clear method. What are the side effects?
```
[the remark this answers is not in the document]
qdequele: Moved to the section `Future possibilities`

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Method: DELETE
+- Route: `/indexes/:index_uid/updates`
+    - index_uid: The index uid
+- Status Code: 204
+- Response: No Response
+
+#### Potential errors
+
+- The given id is a processed id: `400` - `A processed/failed update cannot be abortd`
```
MarinPostma, line 107: not clear, what should the response look like on failed deletion attempt? What should the meili error code be? What http error code?

More generally what type of error can we encounter?

MarinPostma (changes requested):

text/0013-delete-enqueued-updates.md
qdequele: Yes agree, I'll work on that!

qdequele: Updated! Don't hesitate to give feedback on error messages and directly propose changes.

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+#### Potential errors
+
+- The given update id does not exist: 
+    - Status code: `404`
+    - Error code: `update_not_found`
+    - Error description: `The requested update can't be retrieved. Either it doesn't exist, or the database was left in an inconsistent state.`
+
+- The given update id is a processed id: 
+    - Status code: `400`
+    - Error code: `invalid_update_id`
```
curquiza, line 114:
```suggestion
    - Error code: `unabortable_update`
```

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+    - Error description: `The requested update id exist but could not be aborted. A processing/processed/failed update cannot be aborted.`
+
+- Error during the abort process:
+    - Status code: `400`
+    - Error code: `abort_update_failed`
```
curquiza, line 119:
```suggestion
    - Error code: `update_abortion_failed`
```
`abort` is a verb, I try to be consistent with our other errors like `dump_creation_failed` that use a noun before the `_failed` suffix.

text/0013-delete-enqueued-updates.md
curquiza: I felt confused with the error code and the description of the error.

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+
+- The given update id is a processed id: 
+    - Status code: `400`
+    - Error code: `unabortable_update`
+    - Error description: `The requested update id exist but could not be aborted. A processing/processed/failed update cannot be aborted.`
+
+- Error during the abort process:
+    - Status code: `400`
+    - Error code: `update_abortion_failed`
+    - Error description: `An error occurred during the abort update process. The requested abort has not been taken into account.`
```
curquiza, line 120:
```suggestion
    - Error description: `An error occurred during the update abortion process. The requested abortion has not been taken into account.`
```

curquiza (changes requested):

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+# Abort enqueued updates
+
+## 1. Feature Description and Interaction
+
+### Summary
+
+Allow to abort messages in the update queue or to empty the entire queue. An aborted update will have the status `aborted`.
+
+Questions not answered:
+- We have an attribute on update info `processedAt`. What becomes this attribute if we add the possibility to abort an update? We already use this term in the case of a failed update.
```
Kerollmops, line 18: I would say that this becomes that date time at which its status changed from pending to aborted.

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Motivation
+
+The update queue is convenient because it allows us to serialize updates and assure the [ACID property](https://en.wikipedia.org/wiki/ACID) of MeiliSearch. It also causes some problems. As the queue message is an append-only queue, it is impossible to remove an update from this queue. The updates can be large and can take a long time to index. Sometimes, it would be convenient to abort a message in this queue or empty it. 
+
+### Additional Materials
+
+As explained on the [parent spec](https://github.com/meilisearch/specifications/pull/12), It's quite rare to have this kind of update queue. Elasticsearch or Algolia, who have this kind of update queue, don't give the abort or clear their 'tasks' queues.
```
Kerollmops, line 26:
```suggestion
As explained in the [parent spec](https://github.com/meilisearch/specifications/pull/12), It's quite rare to have this kind of update queue. Elasticsearch or Algolia, who have this kind of update queue, don't give the abort or clear their 'tasks' queues. 
```

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+{
+  "status": "aborted",
+  "updateId": 12,
+  "type": {
+    "name": "DocumentsAddition",
+    "number": 4
+  },
+  "duration": 0,
+  "enqueuedAt": "2019-12-07T21:16:09.623944Z",
+  "processedAt": "2019-12-07T21:16:09.703509Z" // Date of deletion
```
Kerollmops, line 82:
```suggestion
  "processedAt": "2019-12-07T21:16:09.703509Z" // date of abortion
```

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+**Abort an update:**
+
+- Method: DELETE
+- Route: `/indexes/:index_uid/updates/:update_id`
+    - index_uid: The index uid
+    - update_id: The update identifier
+- Status code: 204
+- Response: No Body
+
+** abort all update:**
```
Kerollmops, line 97:
```suggestion
**Abort every pending updates:**
```

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+
+- Method: DELETE
+- Route: `/indexes/:index_uid/updates`
+    - index_uid: The index uid
+- Status code: 204
+- Response: No Response
```
Kerollmops, line 103:
```suggestion
- Response: No Body
```

BTW are you sure, in milli I return the ids of the aborted updates in an array, maybe that's not useful.

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+    - Error code: `update_not_found`
+    - Error description: `The requested update can't be retrieved. Either it doesn't exist, or the database was left in an inconsistent state.`
+
+- The given update id is a processed id: 
+    - Status code: `400`
+    - Error code: `unabortable_update`
+    - Error description: `The requested update id exists but could not be aborted. A processing/processed/failed update cannot be aborted.`
+
+- Error during the abort process:
+    - Status code: `400`
```
Kerollmops, line 118: I think that it must be a `500`, I am wrong?

Kerollmops (changes requested):

text/0013-delete-enqueued-updates.md
qdequele: So should we keep `processedAt` or should we change to `abortedAt` in the case of an abortion?

text/0013-delete-enqueued-updates.md
qdequele: Yes, after a discussion with @curquiza, we decided to remain consistent with other `Delete` methods and don't return deleted updates IDs.

text/0013-delete-enqueued-updates.md
qdequele: you're right
```suggestion
    - Status code: `500`
```

text/0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,142 @@
...
+    - Status code: `500`
+    - Error code: `update_abortion_failed`
+    - Error description: `An error occurred during the update abortion process. The requested abortion has not been taken into account.`
+
+
+### Impact on Documentation
+
+- Add on the guide the mention of the `aborted` status. 
+- Change on the guide where we explain that it's not possible to abort an update. 
+- Add the two methods described in the HTTP API part of the documentation.
```
curquiza, line 127:
```suggestion
- Add the two methods described in the HTTP API part of the documentation.
- Add the errors in the errors guide.
```

curquiza (approved):

0013-delete-enqueued-updates.md
```diff
@@ -0,0 +1,151 @@
+- Title: Delete enqueued updates
```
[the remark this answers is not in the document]
qdequele: Changed!

text/0013-abort-pending-updates.md
```diff
@@ -0,0 +1,143 @@
...
+# Abort pending updates
+
+## 1. Feature Description and Interaction
+
+### Summary
+
+Allow to abort messages in the update queue or to empty the entire queue. An aborted update will have the status `aborted`.
+
+Questions not answered:
+- We have an attribute on update info `processedAt`. What becomes this attribute if we add the possibility to abort an update? We already use this term in the case of a failed update.
```
eskombro, line 18: I don't see any problem or conflict with this. Aborting an update is a way of processing it, so we could just set the time it was aborted at.

eskombro (approved):
Glad to see this spec. This is very useful and I feel it is requested a lot. I personally would like this feature to be available!

So thanks! 🎉

text/0013-abort-pending-updates.md
curquiza: The issue exists because we use `processed` as a status to say "it's done". So it could be confused for our users in the context of MeiliSearch since we have chosen this status named `processed`.

bidoubiwa (commented):
This is a great spec! There is one thing that I like a little less but it could be only me. 

## Resftullness
I don't like the fact that the `delete` route is not RESTful in the sense that the delete method works as follow: 

>As the name applies, DELETE APIs are used to delete resources (identified by the Request-URI).

>DELETE operations are idempotent. If you DELETE a resource, **it’s removed from the collection of resources**. Repeatedly calling DELETE API on that resource will not change the outcome – however, calling DELETE on a resource a second time will return a 404 (NOT FOUND) since it was already removed. Some may argue that it makes the DELETE method non-idempotent. It’s a matter of discussion and personal opinion.

For us, this is not true. We only delete a resource if it a specific condition in met. If it is not, nothing is deleted and an error is returned.

## Comparison with ES async

I have looked at what elasticSearch does in their [async-search](https://www.elastic.co/guide/en/elasticsearch/reference/current/async-search.html) route. 

This is there delete route: 
> Delete async searched
You can use the delete async search API to manually delete an async search by ID. **If the search is still running, the search request will be cancelled. Otherwise, the saved search results are deleted**.

They take the party to say: oke you are using the delete route which means you want to delete this resource. 
- If it is `running`, we abord it 
- If it is `done`, we delete it

In our case it would translate the following way: 
- if it is `enqueued`, we abord/remove it
- if it is `running`, we abord it (for later I have red but it would be that)
- if it is `done`, we delete it.

But the reality of what you suggest is 
- If it is `enqueued`, we abord/remove it. 
- if it is `running`: error
- if it is `done`: error

I understand that those errors are made to make the user aware of how `DELETE /update` works, but as an end result it is less intuitive because more developers will expect it to work as a regular RESTfull delete.

## Suggestion

I would change the name of the route
- Route: `DELETE /indexes/:index_uid/updates/enqueued/:update_id`
- Route: `DELETE /indexes/:index_uid/updates/enqueued`

In this case we become restful and intuitive. The errors would also be way more understandable in this context.

## Conclusion 

So my general concern is that this route is not RESTfull and also quiet unclear as intuitively you would think that using DELETE on the update route would remove an update from the update array which is not the case.

And the only example I found (elasticsearch) is restful. 

Of course, I do not want to postpone any development. So these could be arguments for a later rework.

text/0013-abort-pending-updates.md
```diff
@@ -0,0 +1,143 @@
...
+    "name": "DocumentsAddition",
+    "number": 4
+  },
+  "duration": 0,
+  "enqueuedAt": "2019-12-07T21:16:09.623944Z",
+  "processedAt": "2019-12-07T21:16:09.703509Z" // date of abortion
+}
+```
+
+#### HTTP API
```
bidoubiwa, line 86: There is one thing that I like a little less but it could be only me. 

## Resftullness
I don't like the fact that the `delete` route is not RESTful in the sense that the delete method works as follow: 

>As the name applies, DELETE APIs are used to delete resources (identified by the Request-URI).

>DELETE operations are idempotent. If you DELETE a resource, **it’s removed from the collection of resources**. Repeatedly calling DELETE API on that resource will not change the outcome – however, calling DELETE on a resource a second time will return a 404 (NOT FOUND) since it was already removed. Some may argue that it makes the DELETE method non-idempotent. It’s a matter of discussion and personal opinion.

For us, this is not true. We only delete a resource if it a specific condition in met. If it is not, nothing is deleted and an error is returned.

## Comparison with ES async

I have looked at what elasticSearch does in their [async-search](https://www.elastic.co/guide/en/elasticsearch/reference/current/async-search.html) route. 

This is there delete route: 
> Delete async searched
You can use the delete async search API to manually delete an async search by ID. **If the search is still running, the search request will be cancelled. Otherwise, the saved search results are deleted**.

They take the party to say: oke you are using the delete route which means you want to delete this resource. 
- If it is `running`, we abord it 
- If it is `done`, we delete it

In our case it would translate the following way: 
- if it is `enqueued`, we abord/remove it
- if it is `running`, we abord it (for later I have red but it would be that)
- if it is `done`, we delete it.

But the reality of what you suggest is 
- If it is `enqueued`, we abord/remove it. 
- if it is `running`: error
- if it is `done`: error

I understand that those errors are made to make the user aware of how `DELETE /update` works, but as an end result it is less intuitive because more developers will expect it to work as a regular RESTfull delete.

## Suggestion

I would change the name of the route
- Route: `DELETE /indexes/:index_uid/updates/enqueued/:update_id`
- Route: `DELETE /indexes/:index_uid/updates/enqueued`

In this case we become restful and intuitive. The errors would also be way more understandable in this context.

## Conclusion 

So my general concern is that this route is not RESTfull and also quiet unclear as intuitively you would think that using DELETE on the update route would remove an update from the update array which is not the case.

And the only example I found (elasticsearch) is restful. 

Of course, I do not want to postpone any development. So these could be arguments for a later rework.

eskombro: Agreed!

Either we are ABORTING it (which is a status update and might be a PUT request if RESTFUL) or we are DELETING it (in which case it is a DELETE request if restful) and it won't be possible to find it afterwards!

qdequele: So in this case you find it more clear if we propose a PUT on `/indexes/:index_uid/updates/pending/:update_id/abort`?

curquiza: I agree with @eskombro, we should consider this, but without postponing the development as @bidoubiwa mentioned 👍 
I don't really like the huge route name, except if it's the only solution, but we can also think about routes like this (it might be what @eskombro suggested):
- `PUT /indexes/:index_uid/updates`
- `PUT /indexes/:index_uid/updates/:update_id`

Because here we update the `update` resource with a new status

curquiza: Btw being REST is not an obligation for everything
- our routes are not all REST: the `POST` route to update all the settings. And also we use the `PUT` route to update documents but, I read different articles that says we should use `PATCH` instead of `PUT`. I also read the contrary!
- the stripe API is not rest at all -> they only use `POST`, `GET` and `DELETE`, but they are really easy to use :)

What I want to say is not "we should not take this into account" (as I said above, I even suggest something) but we should not abandon "easy to use" for "being RESTful".

eskombro: > I agree with @eskombro, we should consider this, but without postponing the development as @bidoubiwa mentioned 👍
> I don't really like the huge route name, except if it's the only solution, but we can also think about routes like this (it might be what @eskombro suggested):
> 
>     * `PUT /indexes/:index_uid/updates`
> 
>     * `PUT /indexes/:index_uid/updates/:update_id`
> 
> 
> Because here we update the `update` resource with a new status

This is exactly what I had in mind. And yes, RESTFUL is not an obligation, just worth considering :)

qdequele: To be clear, for you, it's logical and understandable that if you do `PUT /indexes/:index_uid/updates/:update_id`, that will abort your update?

bidoubiwa: >To be clear, for you, it's logical and understandable that if you do PUT /indexes/:index_uid/updates/:update_id, that will abort your update?

No it is not, I think we should add a `/abort` to because **PUT does not imply abortion**
. 
I agree with the fact that `DELETE` implies a deletion, which is also not what we are doing, so `delete` is not a good idea.

 
So I would go with @qdequele suggestion (if it was a suggestion)

`PUT /indexes/:index_uid/updates/pending/:update_id/abort`

or, if we want to prepare for a future where tasks that are being `processed` we should go for this:

`PUT /indexes/:index_uid/updates/:update_id/abort`

This will do the following:

**v1**
- If it is **enqueued**: we change status to aborted.
- if it is **processing**: `Error: Cannot abord task that is being processed`
- if it is done: `Error: Cannot abord task that has already been processed`

**v2**
- If it is **enqueued**, We change status to aborted.
- if it is **processing**: We abord and change the status to aborted
- if it is done: `Error: Cannot abord task that has already been processed`

qdequele: 🧐. BTW I like the fact that if you DELETE an processed update/task, this one will be really deleted. What you proposed @bidoubiwa previously, delete a pending update, change it to aborted, delete a processed update really delete it.

eskombro: 1. Yes, using the PUT option would imply passing a parameter that explicitely asks for abortion. Yet, as you point out, this is not clear in the route.

2. The DELETE option is better. But if you delete an "update" that has been processed, this would add too much complications because the history wouldn't be consistent with the data.

As pointed out before, we don't need to be 100% restful and can create an /abort route.

curquiza: > The DELETE option is better. But if you delete an "update" that has been processed, this would add too much complications because the history wouldn't be consistent with the data.

MeiliSearch does not guarantee the consistency between the data and the update history otherwise the update history would not be empty when you import a dump. I mentioned this inconsistency during dump implementation and this is not something that MeiliSearch guarantees. The updates are for consultations: MeiliSearch is not introduced as a real DB and the update history cannot be compared to migration states/files. This is what we said after the meeting about the dumps.

text/0013-abort-pending-updates.md
qdequele: We will keep both!

text/0013-abort-pending-updates.md
qdequele: After an internal discussion we choose to:

- Renaming updates to tasks
The task is more generic, will allow us to put inside the Dumps, etc..

- Renaming the status enqueued to pending
But we keep the enqueuedAt

- Renaming abort to cancel
Abort means already in process/cancel means not yet started

- Use PUT to change the status of a task
Based on the Stripe API. (edited) 

- Keep always all fields in the object

ohld (approved):

text/0013-abort-pending-updates.md
```diff
@@ -0,0 +1,191 @@
...
+ "number": 4
+ },
+ "duration": 0,
+ "enqueuedAt": "2019-12-07T21:16:09.623944Z",
+ "processedAt": null,
+ "canceledAt": "2019-12-07T21:16:09.703509Z", // date of cancelation
+}
+```
+
+Until now, we will consider that every task status documents returned will have all fields. Even empty ones.
```
Kerollmops, line 84: I don't undesratnd this sentence, could you rephrase it, please?

Kerollmops (approved):

text/0013-abort-pending-updates.md
gmourier: @Kerollmops I didn't wrote this specification! I paused it when I arrived to take a new look at it after the 0.21 release :) Probably soon since I'm preparing the next releases cycle 😀

ManyTheFish (approved):

PR #15, opened 2020-12-16
Facets

qdequele:
Scopes:
- Facets on strings/tags/text and arrays. 
- How to declare a faceted attribute.
- How to retrieve facets during the search. How to request them, and what looks like the response. 
- Every parameters/setting that could be used with facets
- ❌ Not talk about filters apart when is about restricting facets values. 
- ❌ The facets on numbers will be discussed in a separate spec.

### DIFF
--- a/text/facets.txt
+++ b/text/facets.txt
@@ -1,0 +1,132 @@
+- Title: Facets
+- Start Date:
+- Specification PR:
+- MeiliSearch Issue: 
+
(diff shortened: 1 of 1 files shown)


bidoubiwa:
related: https://github.com/meilisearch/transplant/issues/65

gmourier:
Close this PR in favor of #27

PR #27, opened 2021-04-08
Facet and Filters behavior

curquiza:


text/0027-filter-and-facet-behavior.md
```diff
@@ -179,20 +180,26 @@ Here is the summary of all the breaking changes (that are detailed in the paragr
...
-// TODO
+See the previous part.
 
 ## 2. Technical Specifications
 
-// TODO
-
 ### Architecture
+
+Internal MeiliSearch engine uses an index data structure for each type encountered in an attribute. Doing this eliminate the need for the user to strictly type the attribute and permits to deal with loosely typed document at indexation. Thus, facilitating the developper experience.
```
gmourier, line 193: @Kerollmops I'm not quite 100% sure about this part. Don't hesitate to point out any errors. Thanks!

text/0027-filter-and-facet-behavior.md
```diff
@@ -179,20 +180,26 @@ Here is the summary of all the breaking changes (that are detailed in the paragr
...
 ### Impact on Documentation
 
-// TODO
+See the previous part.
 
 ## 2. Technical Specifications
 
-// TODO
-
 ### Architecture
```
gmourier, line 191: Kerollmops I'm not quite 100% sure about this part. Don't hesitate to point out any errors. Thanks!

text/0027-filter-and-facet-behavior.md
```diff
@@ -179,20 +180,26 @@ Here is the summary of all the breaking changes (that are detailed in the paragr
...
+
```
gmourier, line 192: Kerollmops I'm not quite 100% sure about the Architecture part. Don't hesitate to point out any errors. Thanks!

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+=> `fantasy` is set to `0` because the fantasy book does not have a `price` equals to `12` as required in the `facetFilters`.
+=> the `filters` and the `facetsDistribution` are related: the `facetsDistribution` is applied after the facet filters.
+
+##### Final decision for v0.21.0
+
+MeiliSearch v0.21.0 will behave as the `facetFilters` already behave in MeiliSearch v0.20.0: the `facetsDistribution` will be applied after the filters.
+
+#### TLDR; all the breaking changes
```
gmourier, line 173: @bidoubiwa @react-learner Do we need more precise indications for the documentation impact? Thanks!

bidoubiwa: Maybe juste some more information about the processing of the filters and facets. Is it done both at indexation time now? Or filters are still sorted at runtime?

bidoubiwa: Could we also have an example of an extreme filtering query? With a lot of facets, and filter, with AND and OR's? Nested in parentheses, etc

gmourier: @bidoubiwa I've added some information. Hope it helps!

gmourier:
@curqui I've added the Issues Summary part!

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+
+Once the facet values that are numbers are stored we got a list of facet values prefixed with the field id and the base level (i.e. 0). We use this base level to generate more levels, each level contains groups of 4 groups of the level below, so level 1 aggregates the ids of the documents of each group of 4 facet values of level 0. The left and right facet values are the inclusive bounds of the group, the level 0 group have equal left and right bounds.
+
+##### Querying phase
+
+Those levels are used to reduce the number of entries to run through, reducing the time it takes to answer too wide range filter queries, like duration > 0 where 80% of the entries will match. We go through each of the levels going from the higher one, the one which describes the biggest amount facet values and, we go deeper in the levels to find a better fitting bound.
+
+### II. Issues Summary
+TBD
```
curquiza, line 226:
```suggestion

In Milli:
- https://github.com/meilisearch/milli/issues/152: the main issue gathering all the internal changes

In Transplant:
- https://github.com/meilisearch/transplant/issues/140: expose the both syntaxes for `filters`
- https://github.com/meilisearch/transplant/issues/70: the push of `attributesForFaceting`. The behavior does not change between v0.20.0 and v0.21.0 but it still needs changes in Transplant.
- https://github.com/meilisearch/transplant/issues/81: only keeping `filters` and removing `facetFilters`
```

curquiza: @gmourier

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+MeiliSearch v0.21.0 will behave as the `facetFilters` already behave in MeiliSearch v0.20.0: the `facetsDistribution` will be applied after the filters.
+
+#### TLDR; all the breaking changes
+
+Here is the summary of all the breaking changes (that are detailed in the paragraphs above):
+
+- The `facetFilters` parameter during the search is removed. Only `filters` can be used.
+- The users need to set the attributes to `attributesForFaceting` to use the filters during the search via the `filters` parameters.
+- The users can now pass an attribute containing numbers (float or integer) in `attributesForFaceting`. It means they can use `filters` and `facetsDistribution` on this numeric field.
+- The `filters` parameter can accept both syntax: string (with `OR`/`AND`) and array.
```
Kerollmops, line 180:
```suggestion
- The `filters` parameter can accept both syntax: string (with `OR`/`AND`/`NOT`) and array.
```

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+
+Internal MeiliSearch engine uses an index data structure for each type encountered in an attribute. Doing this eliminate the need for the user to strictly type the attribute and permits to deal with loosely typed document at indexing time. Thus, facilitating the developper experience.
+
+As for example, let’s imagine that we want to index two documents with a price attribute. One containing `"price": "20"` and the second `"price": 20` as value. The two attribute values will be stored in two different indexes.
+
+Using `=` or `!=` operator on `price` attribute will lead the engine to query the two indexes and get matching document ids with an `UNION` operation. So it will return documents matching `"20"` or `20` as values for the `price` attribute.
+
+#### Implementation Details
+
+We have a database for the facets, the keys are prefixed by the field_id (u8), a level (u8) and, the facet value (i64, f64). The facet values don't have a level when those are strings. The data stored under those keys is the document ids that are faceted under those facets values.
```
Kerollmops, line 211:
```suggestion
We have a database for the facets, the keys are prefixed by the field_id (u8), a level (u8) and, the facet value (f64). The facet values don't have a level when the type is a string. The data stored under those keys is the document ids that are faceted under those facets values.
```

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+
+The type of the facet (i.e. i64, f64 or string) is stored in another data structure and this is by using it that we know how to read the facet value. If the facet type is a number we are able to use more operators like greater than or lower than (e.g. <, <=, >, >=, =, !=).
```
Kerollmops, line 213:
```suggestion
The type of the facet (i.e. f64 or string) is stored in another data structure and this is by using it that we know how to read the facet value. If the facet type is a number we are able to use more operators like greater than or lower than (e.g. <, <=, >, >=, =, !=).
```

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+
+Once the facet values that are numbers are stored we got a list of facet values prefixed with the field id and the base level (i.e. 0). We use this base level to generate more levels, each level contains groups of 4 groups of the level below, so level 1 aggregates the ids of the documents of each group of 4 facet values of level 0. The left and right facet values are the inclusive bounds of the group, the level 0 group have equal left and right bounds.
+
+##### Querying phase
+
+Those levels are used to reduce the number of entries to run through, reducing the time it takes to answer too wide range filter queries, like duration > 0 where 80% of the entries will match. We go through each of the levels going from the higher one, the one which describes the biggest amount facet values and, we go deeper in the levels to find a better fitting bound.
```
Kerollmops, line 223:
```suggestion
Those levels are used to reduce the number of entries to run through, reducing the time it takes to answer too wide range filter queries, like duration > 0 where 80% of the entries will match. We go through each of the levels going from the higher one, the one which describes the biggest amount of facet values and, we go deeper in the levels to find better fitting bounds.
```

Kerollmops (changes requested):

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,237 @@
...
+    }
+}
+```
+
+=> `fantasy` is set to `0` because the fantasy book does not have a `price` equals to `12` as required in the `facetFilters`.
+=> the `filters` and the `facetsDistribution` are related: the `facetsDistribution` is applied after the facet filters.
+
+##### Final decision for v0.21.0
+
+MeiliSearch v0.21.0 will behave as the `facetFilters` already behave in MeiliSearch v0.20.0: the `facetsDistribution` will be applied after the filters.
```
react-learner, line 171:
```suggestion
In MeiliSearch v0.21.0, `facetsDistribution` will behave with `filters` the same way it currently does with `facetFilters`: the `facetsDistribution` will be applied after the filters.
```

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,237 @@
...
+
+#### Implementation Details
+
+We have a database for the facets, the keys are prefixed by the field_id (u8), a level (u8) and, the facet value (f64). The facet values don't have a level when the type is a string. The data stored under those keys is the document ids that are faceted under those facets values.
+
+The type of the facet (i.e. f64 or string) is stored in another data structure and this is by using it that we know how to read the facet value. If the facet type is a number we are able to use more operators like greater than or lower than (e.g. <, <=, >, >=, =, !=).
+
+##### Indexing phase
+
+When documents come in and fields are declared as facets, we start storing the facet values in the previously described database, the key becomes the facet value (as a globally ordered byte slice) and, the entry data now contains the document id that contains this facet value. Note that if the facet value is a number we store it like [field id][level][left facet value][right facet value] where the level is 0 and if it is a string then we don't store the level.
```
react-learner, line 217: Make sure to put brackets in a `[code block]` so that MarkDown doesn't think you're trying to make a [link].

```suggestion
When documents come in and fields are declared as facets, we start storing the facet values in the previously described database, the key becomes the facet value (as a globally ordered byte slice) and, the entry data now contains the document id that contains this facet value. Note that if the facet value is a number we store it like `[field id][level][left facet value][right facet value]` where the level is 0 and if it is a string then we don't store the level.
```

react-learner (changes requested):
I would like to know more about filter normalization: would `"filters": "genres = Comedy"` return a document with `genres : comédy`, and vice versa?

I believe that `facetsDistribution` and `attributesForFaceting` will be renamed in the future, maybe add a note on that.

Really nice work on this! 🧱

react-learner:
I don't know if it's too late to suggest this, but I would strongly recommend we **change the `filters` parameter to just `filter`**. I think it's less confusing, mainly because the parameter's value will only ever be a single filter string or array. 

Even if the value can be nested allowing for complexity, it's still just a logical expression. It's not as though you can define multiple filters that are applied independently or sequentially.

I would also recommend `filterDistribution` and `filterAttributes` to replace `facetsDistribution` and `attributesForFaceting`.

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,237 @@
...
+
+- the string syntax with the `AND`/`OR`/`NOT` operators combined with parentheses
+ex:
+```json
+{
+    "filters": "(genres = Comedy OR genres = Romance) AND director = 'Mati Diop'"
+}
+```
+
+- the array syntax:
```
curquiza, line 81: @gmourier
Checked with @Kerollmops, one question: can we accept the string syntax in an element of the array syntax? Kero says it's technically possible 🙂 ex: `filters: [["price = 20 OR size < 12"], "NOT genre = comedy"]`

gmourier: > I don't know if it's too late to suggest this, but I would strongly recommend we **change the `filters` parameter to just `filter`**. I think it's less confusing, mainly because the parameter's value will only ever be a single filter string or array.
> 
> Even if the value can be nested allowing for complexity, it's still just a logical expression. It's not as though you can define multiple filters that are applied independently or sequentially.
> 
> I would also recommend `filterDistribution` and `filterAttributes` to replace `facetsDistribution` and `attributesForFaceting`.

@curquiza Can we rename `filters` to `filter` at this time? I also think it helps to use it this way as a contextual verb.

@react-learner I think we need to keep `facetsDistribution` as is because this field is used to return computed facets. The parameter name is contextually ok. On another side, I agree with you on `attributesForFaceting` since it's now used to enable filtering on those listed fields and also to ask for MeiliSearch to expose those fields in `facetsDistribution`.

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,237 @@
...
+
+#### Accepted syntaxes for `filters`
+
+Two syntaxes will be accepted for the `filters` parameter during the search:
+
+- the string syntax with the `AND`/`OR`/`NOT` operators combined with parentheses
+ex:
+```json
+{
+    "filters": "(genres = Comedy OR genres = Romance) AND director = 'Mati Diop'"
```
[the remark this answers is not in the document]
gmourier: I've added a text to point that out. Nice catch!

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,230 @@
...
+#### Operator behavior during the search
+
+During the search, the operators should behave as they already do in MeiliSearch v0.20.0.
+
+Here is a quick reminder of the v0.20.0 operator behavior:
+
+- `<`, `=<`, `>`, `=>` => only operates on numbers. MeiliSearch returns only the documents with numbers for this field.
+ex: `price > 19` does not return `"price": "20"` but returns `"price": 20` -> no document conversions are done.
+- `=`, `!=`/`NOT` => string and number operation. MeiliSearch returns the documents with numbers, string (and array of strings) for this field.
+- cannot filter on `null`, objects, arrays of "undefined elements" (ex: array of `null`)
```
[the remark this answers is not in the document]
gmourier: Which part? The one about returning documents?

gmourier:
> I don't know if it's too late to suggest this, but I would strongly recommend we **change the `filters` parameter to just `filter`**. I think it's less confusing, mainly because the parameter's value will only ever be a single filter string or array.
> 
> Even if the value can be nested allowing for complexity, it's still just a logical expression. It's not as though you can define multiple filters that are applied independently or sequentially.
> 
> I would also recommend `filterDistribution` and `filterAttributes` to replace `facetsDistribution` and `attributesForFaceting`.

@curquiza Can we rename `filters` to `filter` at this time? I also think it helps to use it this way as a contextual verb.

@react-learner I think we need to keep `facetsDistribution` since it expresses the fact that we want from MeiliSearch to distribute those facets as a search result parameter. On the other side, I agree with the fact that `attributesForFaceting` is lacking precision because it is now used to authorize filtering and faceting on those fields. We will handle that in another spec.

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,237 @@
...
+{
+    "filters": "(genres = Comedy OR genres = Romance) AND director = 'Mati Diop'"
+}
+```
+
+- the array syntax:
+ex:
+```json
+{
+    "filters": [["genres = Comedy", "genres = Romance"], "director = 'Mati Diop'"]
```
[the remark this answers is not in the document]
gmourier: You can express it like that `"filters": [["genres = Comedy", "director = 'Mati Diop'"], ["genres = Romance", "director = 'Mati Diop'"]]` since `(A && B) || C` is equal to `(A || C) && (B || C)`.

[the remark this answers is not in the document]
react-learner: I see, thanks for explaining that @gmourier 😄

bidoubiwa:
> I would also recommend filterDistribution and filterAttributes to replace facetsDistribution and attributesForFaceting.
> facetsDistribution

Personally, I think both of them may be confusing. As intuitively I would think this would give me back the "median" distribution of filters amongst all previous filters ? I don't know. It's not very clear. 

I thought of `documentDistribution` as we show the distribution of documents between each category. And also, as it clearly informs the users he is going to receive a distribution of **documents**. So even if it loses in facet context, it at least informs about the return value. 
Finally, about the word choices if we had to go for one of the mentioned option: 
- Facets: Its not a common word. Its mainly used by Algolia and was previously used by elastic search. I feel like its something we can relate to as french people because the etymology of the word is [french](https://educalingo.com/fr/dic-en/facet#:~:text=%C3%89TYMOLOGIE%20DU%20MOT%20FACET,French%20facette%20a%20little%20face.)
    `one side of something many-sided, especially of a cut gem`. The definition makes sense, but is not intuitive for non native English or french speaker. 
- Aggregate: The word that is now used by ElasticSearch. `a whole formed by combining several separate elements.`. Which is also technically accurate to describe the distribution. I still feel this is confusing and not intuitive. 
- Filter: Very common word, but filter distribution implies that we would have that kind of distribution `x > 12: 1200 documents` 

So if we make the assumption that the user will have to go to the documentation to understand the meaning of this field, I would choose 
1. Documents Distribution: As at least it is not an intimidating word
2. Facet Distribution: As that is what we already have and what algolia uses
3. Aggregate Distribution: As it is the one used by elastic search, a good chunk of our users 
But not Filter Distribution: As it is not really accurate

text/0027-filter-and-facet-behavior.md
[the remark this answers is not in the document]
gmourier: Thanks to @MarinPostma and @Kerollmops for helping me on this one 😅

react-learner:
@gmourier I agree that `facetsDistribution` describes the behavior pretty well; I can't really imagine any scenario besides building a faceted search UI where I would want to use this parameter. However, I also think it's worth noting that with the changes proposed here, **`facets` no longer has a particular meaning within MeiliSearch**. As far as the software is concerned, faceting is a specific way of using `filters`. In other words, **knowledge of the term "faceting" is optional to use `filters`, or even to make a faceted search UI w/ MeiliSearch**. So, why not make it optional throughout our software and documentation?

I recognize that "Faceted search" is one of MeiliSearch's desirable features, so I'm definitely not proposing that we eliminate the keyword from the docs—it attracts web traffic and communicates well with the search-focused crowd. However, I think we could background "facets"/"faceting"/"faceted search" by keeping these words off of parameter names, substituting more intuitive names like `categoryDistribution` or `resultDistribution`. By doing so, we can reduce friction for people who have never seen the word "facets" before (i.e. less search-focused users: hobbyists, solo devs, IT professionals) without impacting those who are more knowledgeable.

It's an extreme example, but I wouldn't want someone to search "facets meilisearch" on Google and end up on the `facetsDistribution` reference page instead of our main guide on setting up faceted search. Also, having a single software-level reference to `facets` when all others have been removed just seems inconsistent to me 🤷‍♂️

text/0027-filter-and-facet-behavior.md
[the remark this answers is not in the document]
bidoubiwa: I think the limitied filtering may be linked to this behavior in meilisearch: https://docs.meilisearch.com/reference/under_the_hood/datatypes.html

react-learner:
### Some possible names for facetsDistribution:

`hitsByField` : this one is really intuitive, since the value of the parameter should be a list of field names. Of course `numberOfHitsByField` would be more accurate, but that's pretty long. `hits` could also be replaced by `matches` or `results` depending on how we define these terms internally. 

`documentDistribution` : proposed by @bidoubiwa 

`result/hit/matchDistribution` : similar to Bidoubiwa's. The nice thing about the word `distribution` is it instantly conveys that the response will be a number (this is not as clear with `hitsByField`)

`facetDistribution` : at the very least, we must get rid of the `s` to be more Anglophone 😆

gmourier:
I've updated the spec to mention that we are renaming `filters` to `filter`.

text/0027-filter-and-facet-behavior.md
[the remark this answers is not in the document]
react-learner: I meant 
```
- cannot filter on `null`, objects, arrays of "undefined elements" (ex: array of `null`)
```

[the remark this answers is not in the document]
gmourier: I've separated that from the operators behavior part for clarity !

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,265 @@
...
+- Outer arrays elements are connected by an AND operator (e.g. ["genres:Romance", "director:Mati Diop"]).
+
+Example:
+```json
+{
+    "filter": [["genres = Comedy", "genres = Romance"], "director = 'Mati Diop'"]
+}
+```
+
+> The array syntax is limited to express a
```
curquiza, line 100: the sentence looks like it's not over

text/0027-filter-and-facet-behavior.md
```diff
@@ -0,0 +1,265 @@
...
+##### Final decision for v0.21.0
+
+In MeiliSearch v0.21.0, `facetsDistribution` will behave with `filter` the same way it currently does with `facetFilters`: the `facetsDistribution` will be applied after the filters.
+
+#### TLDR; all the breaking changes
+
+Here is the summary of all the breaking changes (that are detailed in the paragraphs above):
+
+- The `facetFilters` parameter during the search is removed. Only `filter` can be used.
+- The `filters`parameter is renamed `filter`.
```
curquiza, line 208:
```suggestion
- The `filters` parameter is renamed `filter`.
```

curquiza (commented):
Small remarks

Kerollmops (approved):
The syntax seems good to me, especially the new filter array syntax :)
Thank you!

curquiza (commented):
I can't approve because it's my PR, but I approve 😇

bidoubiwa (approved):
LGTM :fire:

gmourier (approved):
Replaces @curquiza approval

PR #9, opened 2020-11-17
Update process

qdequele:
The goal of this PR is to add more precision to the process to follow during a Spec creation.

PR #28, opened 2021-04-09
Indexing CSVs

gmourier:
The goal of this PR is to specify CSV file indexing feature.

### DIFF
--- a/text/0028-csv-indexation.md
+++ b/text/0028-csv-indexation.md
@@ -1,7 +1,7 @@
 - Title: Csv Indexation
 - Start Date: 2021-04-9
-- Specification PR: TODO
-- Meilisearch Issue: [#128](https://github.com/meilisearch/transplant/issues/128),[#1332](https://github.com/meilisearch/MeiliSearch/issues/1332)
+- Specification PR: [PR-#28](https://github.com/meilisearch/specifications/pull/28)
+- MeiliSearch Issues: [#128](https://github.com/meilisearch/transplant/issues/128),[#1332](https://github.com/meilisearch/MeiliSearch/issues/1332)
 
 ## Feature Description and Interaction
 
--- a/text/0028-csv-indexation.md
+++ b/text/0028-csv-indexation.md
@@ -107,7 +107,7 @@ curl \
 
 > - Sending a different payload than the `Content-Type` header should return a `415 unsupported_media_type` error.
 > - Too large payload according to the limit should return a `413 payload_too_large` error 
-> - Wrong file encoding should return should return a `420 unprocessable_entity` error
+> - Wrong file encoding should return a `420 unprocessable_entity` error
 > - Invalid CSV data should return a `420 unprocessable_entity` error
 
 ### Add or Update Documents [📎](https://docs.meilisearch.com/reference/api/documents.html#add-or-update-documents)
@@ -124,7 +124,7 @@ curl \
 
 > - Sending a different payload than the `Content-Type` header should return a `415 unsupported_media_type` error.
 > - Too large payload according to the limit should return a `413 payload_too_large` error 
-> - Wrong file encoding should return should return a `420 unprocessable_entity` error
+> - Wrong file encoding should return a `420 unprocessable_entity` error
 > - Invalid CSV data should return a `420 unprocessable_entity` error
 
 ### V. Impact on documentation
--- a/text/0028-csv-indexation.md
+++ b/text/0028-csv-indexation.md
@@ -3,41 +3,52 @@
 - Specification PR: [PR-#28](https://github.com/meilisearch/specifications/pull/28)
 - MeiliSearch Issues: [#128](https://github.com/meilisearch/transplant/issues/128),[#1332](https://github.com/meilisearch/MeiliSearch/issues/1332)
 
-## Feature Description and Interaction
+## 1. Feature Description and Interaction
 
-### Summary
+### I. Summary
 
 The initiation step of document indexing is to send some file matching a format to be parsed and tokenized in order to give search results to end-users. A [CSV](https://en.wikipedia.org/wiki/Comma-separated_values) file format is broadly used to store and exchange data in a simple format.
 
-Also, in order to boost write performance csv files are more suited than JSON for huge datasets.
+Also, in order to boost write performance csv files are more suited than JSON for consequent datasets.
 
-### Motivation
+### II. Motivation
 
 We want to provide our users with an always improved usage experience. Currently, the engine only accepts JSON format as a data source. We want to give users the possibility of another simple file format, well known, to use. Thus, give them more versatility at the data source choices for the indexation (add and update) step.
 
 Since most of SQL engines or SQL clients can easily dump data as a csv file, it will facilitate MeiliSearch adoption by extending the indexation step on a wider range of customer cases than before.
 
-Writing performance is also considered as a motivation since CSV parsing is less CPU and memory intensive than parsing Json.
+Writing performance is also considered as a motivation since CSV parsing is less CPU and memory intensive than parsing Json due to the streamable capability of the CSV format.
 
-### Explanation
+### III. Additional Materials
+N/A
 
-#### Csv formatting rules
+### IV.Explanation
+
+#### Csv Formatting Rules
 
 While there's [RFC 4180](https://tools.ietf.org/html/rfc4180) as a try to add a specification for csv format, we will find a lot of variations from that. MeiliSearch features capabilities requires the csv file to be formatted the proper way to be parsable by the engine.
 
-- Each csv file need to contain a first line representing the list of attributes.
-- Each of the following lines will represent a document inside Meilisearch after indexation (⚠️ Not if DISTINCT mode is enabled on an attribute?).
-- Each string value should be enclosed in double quotes tags to handle comma character presence inside.
-- Using double-quotes to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote as mentioned in [RFC 4180](https://tools.ietf.org/html/rfc4180).
-- No space between cells.
-- ⚠ Find a way to handle array of values. Using a csv to json transformer is an alternative. ⚠
+- A csv file need to contain a first line representing the list of attributes with the choosen type separated from the attribute name by `:` character. 
 
-*Example with a comma inside a cell*
+> An attribute can be specificed with two types: `string` or `number`. A missing type will be interpreted as a `string` by default.
+>
+> Valid headline example: "id:number","title:string","author","price:number"
+
+- Each of the following lines will represent a document for MeiliSearch.
+- Each value should be enclosed in double quotes tags to handle comma character presence inside.
+- Using double-quotes to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote as mentioned in [RFC 4180](https://tools.ietf.org/html/rfc4180).
+- Float value should be written with a `.` character, like `3.14`
+- CSV text should be encoded in UTF8
+- The format can't handle array cell values. We are providing `nd-json` format to deal with theses types of attribute in a easier way.
+
+##### Example with a comma inside a cell
+
+Given the csv payload
 ```
-"id","label","price","colors","description"
-1,"t-shirt",4.99,"red","Thus, you will rock at summer time."
+"id:number","label","price:number","colors","description"
+"1","t-shirt","4.99","red","Thus, you will rock at summer time."
 ```
-should be dumped to that as a search result
+the search result should be displayed as 
 ```json=
 {
   "hits": [
@@ -53,26 +64,22 @@ should be dumped to that as a search result
 }
 ```
 
-*Example with a double quote inside a cell*
+##### Example with a double quote inside a cell
+
+Given the csv payload
 ```
-"id","label","price","colors","description"
+"id:number","label","price","colors","description"
 "1","t-shirt","4.99","red","Hey, you will ""rock"" at summer time."
 ```
-
-*Example with an array of value inside a cell*
-```
-"id","label","price","colors","description"
-"1","t-shirt","4.99","red|green|blue","Hey, you will ""rock"" at summer time."
-```
-*should result to*
+the search result should be displayed as
 ```json=
 {
   "hits": [
     {
       "id": 1,
       "label": "t-shirt",
-      "price": 4.99,
-      "colors": ["red", "green", "blue"],
+      "price": "4.99",
+      "colors": "red",
       "description": "Hey, you will rock at summer time.",
     }
   ],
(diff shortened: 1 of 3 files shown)


MarinPostma:
Should this be a joint spec with json line? Support already exist, and just needs to be exposed

gmourier:
I personally advocate separating them into two distinct specs as we can use the merged pull requests to reflect what the engine is capable of as a single source of truth. Nonetheless, we need to take a decision about this statement. Do we use it as a task tracker or do we use it as a support for mapping the whole Meilisearch capabilities at a given point in time? I've added a [commit](https://github.com/meilisearch/specifications/pull/9/commits/5ab7cc365eb2264c65c7804b718415a74a5877a6) on the last specification guide proposal to elaborate a specification lifecycle (github automation can't do all this steps, so it will need a lot of manual work I think).

What do you mean by saying "expose"? Thinking about this one, I'm wondering if we need to expose it as an SDK's method to pass CSV file through http. It would be nice to have for sure especially to propose an interface for the future dashboard or DX but maybe we can just use it as a discovery tool to measure usage and things on CLI scope to iterate in the future.

MarinPostma:
What I mean by expose is that this feature is already implemented and supported by the engine, and is disabled for use by the user.

I get your point about isolation of specs, I'm just wondering if that since this is already implemented, and is basically the same thing, we should maybe ship the two at once, and avoid unecessary overhead.

As for the sdk, what they decide to implement usually follows what MeiliSearch exposes in it's api.

text/csv-indexation.md
```diff
@@ -0,0 +1,127 @@
...
+
+Also, in order to boost write performance csv files are more suited than JSON for huge datasets.
+
+### Motivation
+
+We want to provide our users with an always improved usage experience. Currently, the engine only accepts JSON format as a data source. We want to give users the possibility of another simple file format, well known, to use. Thus, give them more versatility at the data source choices for the indexation (add and update) step.
+
+Since most of SQL engines or SQL clients can easily dump data as a csv file, it will facilitate MeiliSearch adoption by extending the indexation step on a wider range of customer cases than before.
+
+Writing performance is also considered as a motivation since CSV parsing is less CPU and memory intensive than parsing Json.
```
MarinPostma, line 20: The main issue with json is that you need to hold the entirety on the deserialized payload in memory before you can start ingesting. This is a huge overhead.

text/csv-indexation.md
```diff
@@ -0,0 +1,127 @@
...
+- Should we privilege CSV ingestion at engine level (avoiding messaging transplant with http call leading to network overhead) or CSV ingestion from SDk's point of view for developper experience. Also, the last point allows building of an interface in the dashboard to send CSV file to the search engine. Should we considerate a first interation aims at feature adoption and payload size analytics? If we advocate for the discovery aspect of the feature, we should keep it as a simple and stupid (not handle the corner case of cell containing array definition)
```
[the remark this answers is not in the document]
MarinPostma: I'm not sure what you mean by "avoid messaging transplant"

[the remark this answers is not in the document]
MarinPostma: When you say ingest csv from the sdk point of view, you mean from the API?

text/csv-indexation.md
```diff
@@ -0,0 +1,127 @@
...
+Writing performance is also considered as a motivation since CSV parsing is less CPU and memory intensive than parsing Json.
+
+### Explanation
+
+#### Csv formatting rules
+
+While there's [RFC 4180](https://tools.ietf.org/html/rfc4180) as a try to add a specification for csv format, we will find a lot of variations from that. MeiliSearch features capabilities requires the csv file to be formatted the proper way to be parsable by the engine.
+
+- Each csv file need to contain a first line representing the list of attributes.
+- Each of the following lines will represent a document inside Meilisearch after indexation (⚠️ Not if DISTINCT mode is enabled on an attribute?).
```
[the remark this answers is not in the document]
gmourier: Thank you!

text/csv-indexation.md
gmourier: Thanks, I will update the specs to mention that in this part.

text/csv-indexation.md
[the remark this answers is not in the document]
gmourier: > We want to privilege a data format that is streamable and that does not require unmarshalling to disk: both csv and jsonl are supported now that allow that. The conversion to json is done at search time, when the documents are federalized from.the database

It seems that Algolia proposes a tool to convert complex CSV values (For example the cell "genres" contains values that need to be parsed as an array). To do that, they seem to convert CSV to JSON and push it to their main JSON indexing method. The main caveat of that is concerning the performance issue since we are aiming to propose a fast way of importing a bigger dataset. Does the CSV indexation method currently handle that type of cell from a CSV file?

> I'm not sure what you mean by "avoid messaging transplant"
I should have elaborate on my question more precisely. Please correct me if I'm wrong but I understand that Transplant is used to expose HTTP API Endpoints. By avoiding Transplant, I mean using the CLI on the engine side.

>When you say ingest csv from the sdk point of view, you mean from the API?
Exactly.

text/csv-indexation.md
```diff
@@ -0,0 +1,127 @@
...
+  inputFile: '',
+  batchSize: 1000,
+  delimiter: ',',
+  clearIndex: false,
+  parseArrays: ['column'],
+  arrayDelimiter: '|',
+  geoColumns: {lat: 'latCol', 'lng': 'longColumn'}
+});
+```
+- Do we need to enforce a specific encoding for the file considering that Meilisearch is aimed at supporting many languages.
```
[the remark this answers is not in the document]
gmourier: It seems to be enough for any language. Do we need to check that and throw an error if it is not an UTF8 file?

text/csv-indexation.md
```diff
@@ -0,0 +1,127 @@
...
+
+>Note that SDKs should add a specific ***Content-Type*** text/csv header on POST and PUT Documents endpoints.
+
+## Technical Specifications
+
+### Architecture
+
+### Technical Limitations
+
+- Is CSV payload size is equivalent to [JSON payload size](https://docs.meilisearch.com/reference/features/configuration.html#payload-limit-size)?
```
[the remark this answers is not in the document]
gmourier: It could be nice to have the insight to know if users change the payload size from analytics. Letting us know what we need to adjust in the future.

text/csv-indexation.md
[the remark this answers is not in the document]
MarinPostma: >  Does the CSV indexation method currently handle that type of cell from a CSV file?
No, it only support ISO csv. We use this lib under the hood: https://docs.rs/csv/1.1.6/csv/
The best thing to do would be to convert the csv into jsonl and not json, so we keep the streaming capabilities and performance. Handling such csv is a real question for me though.

> I should have elaborate on my question more precisely. Please correct me if I'm wrong but I understand that Transplant is used to expose HTTP API Endpoints.

transplant does a lot more than that, I'll explain it to you on our call this afternoon if you want :)

> Exactly.

gotcha, we usually use sdk to refer to the sdk integrations :)

text/csv-indexation.md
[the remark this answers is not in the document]
MarinPostma: it should be the case already, since it would fail to parse.This has not been encountered though, so I'm not sure what you'd get.

gmourier:
Just to follow up with you @curquiza, I will update the tech part as soon as I've updated the PR related to the change in spec definition.

gmourier:
As mentioned on https://github.com/meilisearch/transplant/issues/142, tech part needs to be updated to mention that.

gmourier:
As mentioned in https://github.com/meilisearch/transplant/issues/142 it will be a candidate for v0.22

text/0028-csv-indexation.md
```diff
@@ -0,0 +1,156 @@
+- Title: Csv Indexation
+- Start Date: 2021-04-9
+- Specification PR: [PR-#28](https://github.com/meilisearch/specifications/pull/28)
+- MeiliSearch Issues: [#128](https://github.com/meilisearch/transplant/issues/128),[#1332](https://github.com/meilisearch/MeiliSearch/issues/1332)
+
+## 1. Feature Description and Interaction
+
+### I. Summary
+
+The initiation step of document indexing is to send some file matching a format to be parsed and tokenized in order to give search results to end-users. A [CSV](https://en.wikipedia.org/wiki/Comma-separated_values) file format is broadly used to store and exchange data in a simple format.
```
curquiza, line 10: Maybe kero will explain this better than me, but I'm not sure there is any tokenization during indexation.
From my POV the indexation purpose is to sort/organize the documents of the user to make MeiliSearch able to retrieve them in the most efficient way during a search.

text/0028-csv-indexation.md
```diff
@@ -0,0 +1,156 @@
...
+  ],
+  ...
+}
+```
+
+> Note that the price attribute was not typed as a number. By default, MeiliSearch type it as a string.
+
+#### API Endpoints
+
+> Each API endpoints mentioned above will now require a `text/csv` as `Content-Type` header to process CSV data.
```
curquiza, line 94: Currently, the users don't need to use the header `Content-type: application/json` when adding documents in JSON because MeiliSearch infers they want to use JSON.
Is still `Content-type: application/json` the default header if no header is passed or should the users always pass a header with the type from now?

curquiza (changes requested):
What bout

text/0028-csv-indexation.md
gmourier: Yes, I wrote that to point out why we need to send data for indexing. I can just write that we need indexing to return those data as a search result.

text/0028-csv-indexation.md
gmourier: We could keep this default behavior and make sure that if `Content-Type: application/json` is set, we have the exact same behavior.

curquiza: We should specify this behavior in the spec (sorry if it's already present, but I don't see it). Also, this should be present in the `impact on the Documentation` section. The documentation should describe this default behavior somewhere I think. It's up to tommy to decide, but I would rather he's aware of this behavior.

gmourier: I have added that part. Nice catch! Same for NDJSON specification.

curquiza (approved):
Awesome spec! Thanks :)
I would rather Marin or Kero to review it before any merge

bidoubiwa (approved):
🔥

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+
+## 1. Feature Description and Interaction
+
+### I. Summary
+
+To index documents, the body of the add documents request has to match a specific format. That specific format is then parsed and tokenized inside MeiliSearch. After which, the documents added are in the pool of searchable and returnable documents.
+
+A [CSV](https://en.wikipedia.org/wiki/Comma-separated_values) data format is broadly used to store and exchange data in a simple format.
+
+Also, in order to boost write performance csv data format is more suited than JSON for consequent datasets.
```
Kerollmops, line 16:
```suggestion
Also, in order to boost write performance CSV data format is more suited than JSON for consequent datasets has keys are not duplicated for every document.
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+
+### II. Motivation
+
+We want to provide our users with an always improved usage experience. Currently, the engine only accepts JSON format as a data source. We want to give users the possibility of another simple data format, well known, to use. Thus, give them more versatility at the data source choices for the indexation (add and update) step.
```
Kerollmops, line 20:
```suggestion
We want to provide our users with an always improved usage experience. Currently, the engine only accepts JSON format as a data source. We want to give users the possibility of another simple data format, well known, to use. Thus, give them more versatility at the data source choices for the indexing (add and update) step.
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+
+Since most of SQL engines or SQL clients can easily dump data as csv, it will facilitate MeiliSearch adoption by extending the indexation step on a wider range of customer cases than before.
```
Kerollmops, line 22:
```suggestion
Since most SQL engines or SQL clients can easily dump data as CSV, it will facilitate MeiliSearch adoption by extending the indexing step on a wider range of customer cases than before.
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+### III. Additional Materials
+N/A
+
+### IV.Explanation
+
+#### Csv Formatting Rules
+
+While there's [RFC 4180](https://tools.ietf.org/html/rfc4180) as a try to add a specification for csv format, we will find a lot of variations from that. MeiliSearch features capabilities requires csv data to be formatted the proper way to be parsable by the engine.
+
+- Csv data format need to contain a first line representing the list of attributes with the choosen type separated from the attribute name by `:` character.
```
Kerollmops, line 35:
```suggestion
- CSV data format needs to contain a first line representing the list of attributes with the optionally chosen type separated from the attribute name by `:` character. The type is case insensitive.
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+
+> An attribute can be specificed with two types: `string` or `number`. A missing type will be interpreted as a `string` by default.
+>
+> Valid headline example: "id:number","title:string","author","price:number"
+
+- Each of the following lines will represent a document for MeiliSearch.
```
Kerollmops, line 41:
```suggestion
- The following CSV lines will represent a document for MeiliSearch.
```

I modified this because it is not exactly true, a CSV row can contain an escaped newline.

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+- Each value should be enclosed in double quotes tags to handle comma character presence inside.
```
Kerollmops, line 42:
```suggestion
- A CSV value should be enclosed in double-quotes when it contains a comma character to escape it.
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+the search result should be displayed as
+```json
+{
+  "hits": [
+    {
+      "id": 1,
+      "label": "t-shirt",
+      "price": 4.99,
+      "colors": "red",
+      "description": "Hey, you will rock at summer time.",
```
Kerollmops, line 64:
```suggestion
      "description": "Hey, you will rock at summer time."
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+  --data '
+    "id","label","price:number","colors","description"\n
+    "1","hoodie","19.99","purple","Hey, you will rock at summer time."
+  '
+```
+> Response code: 202 Accepted
+
+##### Error codes
+
+> - Sending a different payload than the `Content-Type` header should return a `415 unsupported_media_type` error.
```
Kerollmops, line 115: Hum, not sure about that, MeiliSearch will try to parse it by using the specified format and will return a bad request error with the associated parsing error.

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+> - Too large payload according to the limit should return a `413 payload_too_large` error
+> - Wrong encoding should return a `420 unprocessable_entity` error
+> - Invalid CSV data should return a `420 unprocessable_entity` error
```
Kerollmops, line 118: I think it's the same as above, [a bad request (400)](https://httpstatuses.com/400).

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Add or Update Documents [📎](https://docs.meilisearch.com/reference/api/documents.html#add-or-update-documents)
+
+```curl
+curl \
+  -X PUT 'http://localhost:7700/indexes/movies/documents' \
+  -H 'Content-Type: text/csv' \
+  --data '
```
Kerollmops, line 126:
```suggestion
  --binary-data '
```

It is highly advised to use `--binary-data` instead of `--data` as it keeps the space formatting of the body sent, where `--data` weirdly interpret it and deduplicated spaces.

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+  '
+```
+> Response code: 202 Accepted
+
+##### Errors handling
+
+> - Sending a different payload than the `Content-Type` header should return a `415 unsupported_media_type` error.
+> - Too large payload according to the limit should return a `413 payload_too_large` error
+> - Wrong encoding should return a `420 unprocessable_entity` error
+> - Invalid CSV data should return a `420 unprocessable_entity` error
```
Kerollmops, line 138: Same as the above errors I think.

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+
+This feature should impact MeiliSearch users documentation by adding mention of csv capability inside Documents scope at [Add or replace documents](https://docs.meilisearch.com/reference/api/documents.html#add-or-replace-documents) and [Add or update documents](https://docs.meilisearch.com/reference/api/documents.html#add-or-update-documents). It should also mention that a missing Content-Type will be interpreted as `application/json` since it's the current behavior. Giving an `application/json` Content-Type leads to the same behavior.
+
+We should also not only mention JSON format in `unsupported_media_type` section on the [errors page](https://docs.meilisearch.com/errors/#unsupported_media_type) and add CSV format. The documentation says "Currently, MeiliSearch supports only JSON payloads."
+
+Documentation should also guide the user in the correct way to properly format and send csv data. Adding a dedicated page for the purpose of formatting and sending CSV data should be considered.
+
+### VI. Impact on SDKs
+
+This feature should impact MeiliSearch SDK's in the future by adding the possibility to send a CSV to MeiliSearch on the previous explicited endpoints.
```
Kerollmops, line 150:
```suggestion
This feature should impact MeiliSearch SDKs in the future by adding the possibility to send a CSV to MeiliSearch on the previous explicit endpoints. Simplifying the typing of the headers could also be handled by the SDKs.
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+
+## 2. Technical Aspects
+
+### I. Technical Details
+
+⚠ A missing Content-Type will be interpreted as `application/json` since it's the current behavior. Giving an `application/json` Content-Type leads to the same behavior.
+
+## 3. Future possibilities
+
+- Provide an interface in the future dashboard to upload CSV data into an index.
```
Kerollmops, line 160:
```suggestion
- Provide an interface in the future dashboard to upload CSV data into an index and optionally provide the headers types.
```

Kerollmops (changes requested):

text/0028-indexing-csv.md
gmourier: I used the https://docs.meilisearch.com/errors/#unsupported_media_type in the documentation to state that it will be possible to do that. However, it would be nice to use more granular error codes when possible. Bad request is rather generic.

gmourier: Edit: 

I've tested to send a wrong JSON payload and I got a 400 Bad Request error.  
Sending data without a Content-Type header leads to a 415 Unsupported media type.

text/0028-indexing-csv.md
gmourier: I have updated the error codes!

text/0028-indexing-csv.md
gmourier: I have updated the error codes!

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,161 @@
...
+### III. Additional Materials
+N/A
+
+### IV.Explanation
+
+#### Csv Formatting Rules
+
+While there's [RFC 4180](https://tools.ietf.org/html/rfc4180) as a try to add a specification for csv format, we will find a lot of variations from that. MeiliSearch features capabilities requires csv data to be formatted the proper way to be parsable by the engine.
+
+- CSV data format needs to contain a first line representing the list of attributes with the optionally chosen type separated from the attribute name by `:` character. The type is case insensitive.
```
MarinPostma, line 35: is this still necessary, now that milli handles types internally?

Kerollmops: milli no more do conversions between strings and numbers therefore it interprets all the CSV values as strings, it needs the user to type the field to make sure it is interpreted as the valid type (number or string).

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,159 @@
...
+
+## 1. Feature Description and Interaction
+
+### I. Summary
+
+To index documents, the body of the add documents request has to match a specific format. That specific format is then parsed and tokenized inside MeiliSearch. After which, the documents added are in the pool of searchable and returnable documents.
+
+A [CSV](https://en.wikipedia.org/wiki/Comma-separated_values) data format is broadly used to store and exchange data in a simple format.
+
+Also, in order to boost write performance CSV data format is more suited than JSON for consequent datasets has keys are not duplicated for every document.
```
Kerollmops, line 16:
```suggestion
Also, in order to boost write performance CSV data format is more suited than JSON for consequent datasets, as keys are not duplicated for every document.
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,159 @@
...
+Writing performance is also considered as a motivation since CSV parsing is less CPU and memory intensive than parsing Json due to the streamable capability of the CSV format.
+
+### III. Additional Materials
+N/A
+
+### IV.Explanation
+
+#### Csv Formatting Rules
+
+While there's [RFC 4180](https://tools.ietf.org/html/rfc4180) as a try to add a specification for csv format, we will find a lot of variations from that. MeiliSearch features capabilities requires csv data to be formatted the proper way to be parsable by the engine.
```
Kerollmops, line 33:
```suggestion
While there's [RFC 4180](https://tools.ietf.org/html/rfc4180) as a try to add a specification for CSV format, we will find a lot of variations from that. MeiliSearch features capabilities requires CSV data to be formatted the proper way to be parsable by the engine.
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,159 @@
...
+
+- CSV data format needs to contain a first line representing the list of attributes with the optionally chosen type separated from the attribute name by `:` character. The type is case insensitive.
+
+> An attribute can be specificed with two types: `string` or `number`. A missing type will be interpreted as a `string` by default.
+>
+> Valid headline example: "id:number","title:string","author","price:number"
+
+- The following CSV lines will represent a document for MeiliSearch.
+- A CSV value should be enclosed in double-quotes when it contains a comma character to escape it.
```
Kerollmops, line 42:
```suggestion
- A CSV value should be enclosed in double-quotes when it contains a comma character or a newline to escape it.
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,159 @@
...
+- Using double-quotes to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote as mentioned in [RFC 4180](https://tools.ietf.org/html/rfc4180).
+- Float value should be written with a `.` character, like `3.14`.
+- CSV text should be encoded in UTF8.
+- The format can't handle array cell values. We are providing `nd-json` format to deal with theses types of attribute in a easier way.
+
+##### Example with a comma inside a cell
+
+Given the csv payload
```
Kerollmops, line 50:
```suggestion
Given the CSV payload
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,159 @@
...
+      "description": "Hey, you will rock at summer time."
+    }
+  ],
+  ...
+}
+```
+
+##### Example with a double quote inside a cell
+
+Given the csv payload
```
Kerollmops, line 73:
```suggestion
Given the CSV payload
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,159 @@
...
+> Note that the price attribute was not typed as a number. By default, MeiliSearch type it as a string.
+
+#### API Endpoints
+
+> Each API endpoints mentioned above will now require a `text/csv` as `Content-Type` header to process CSV data.
+> ⚠ A missing Content-Type will be interpreted as `application/json` since it's the current behavior. Giving an `application/json` Content-Type leads to the same behavior.
+
+#### Add or Replace Documents [📎](https://docs.meilisearch.com/reference/api/documents.html#add-or-replace-documents)
+
+```curl
```
Kerollmops, line 103:
```suggestion
```bash
```

text/0028-indexing-csv.md
```diff
@@ -0,0 +1,159 @@
...
+##### Error codes
+
+> - Sending a different payload than the `Content-Type` header should return a `400 bad_request` error.
+> - Too large payload according to the limit should return a `413 payload_too_large` error
+> - Wrong encoding should return a `400 bad_request` error
+> - Invalid CSV data should return a `400 bad_request` error
+
+### Add or Update Documents [📎](https://docs.meilisearch.com/reference/api/documents.html#add-or-update-documents)
+
+```curl
```
Kerollmops, line 123:
```suggestion
```bash
```

Kerollmops (changes requested):
Thank you :)

Kerollmops (approved):

PR #29, opened 2021-04-12
Indexing NDJSONs

gmourier:
The goal of this PR is to specify NDJSON indexing.


text/0029-indexing-ndjson.md
```diff
@@ -0,0 +1,132 @@
...
+- Specification PR: [PR-#29](https://github.com/meilisearch/specifications/pull/29)
+- MeiliSearch Tracking-Issues: TBD
+
+# Indexing NDJSON
+
+## 1. Feature Description and Interaction
+
+### I. Summary
+
+The initiation step of document indexing is to send some file matching a format to be parsed and tokenized in order to give search results to end-users. An [NDJSON](http://ndjson.org/) file is easier to use than a CSV file because it propose a convenient format for storing structured data.
```
bidoubiwa, line 12:
```suggestion
To index documents, the body of the add documents request has to match a specific format. That specific format is then parsed and tokenized inside MeiliSearch. After which, the documents added are in the pool of searchable and returnable documents.

An [NDJSON](http://ndjson.org/) file is easier to use than a CSV file because it propose a convenient format for storing structured data.
```

I had a hard time understanding the first part. Also the transition between the first part and the second part is confusing. What is the context of the comparison?

text/0029-indexing-ndjson.md
```diff
@@ -0,0 +1,132 @@
...
+
+### II. Motivation
+
+Currently, the engine only accepts JSON format as a data source. We want to give users the possibility of another simple file format to use. Thus, give them more versatility at the data source choices for the indexation step.
+
+Writing performance is also considered as a motivation since JSON Lines file parsing is less CPU and memory intensive than parsing standard JSON because every new lines represent separate entries, making the NDJSON file streamable. Thus, more suited for indexing a consequent data set.
```
bidoubiwa, line 18:
```suggestion
Writing performance is also a motivation since JSON Lines file parsing is less CPU and memory-intensive than parsing standard JSON. When new lines represent separate entries it makes the NDJSON file streamable, thus, more suited for indexing a consequent data set.
```

text/0029-indexing-ndjson.md
```diff
@@ -0,0 +1,132 @@
...
+      ],
+    }
+  ],
+  ...
+}
+```
+
+#### API Endpoints
+
+> Each API endpoints mentioned above will now require a `application/x-ndjson` as `Content-Type` header to process NDJSON data.
```
bidoubiwa, line 74:
```suggestion
> Each API endpoints mentioned above will now require a `application/x-ndjson` as `Content-Type` header to be processed as NDJSON data.
```

text/0029-indexing-ndjson.md
```diff
@@ -0,0 +1,132 @@
...
+  '
+```
+> Response code: 202 Accepted
+
+##### Error codes
+
+> - Sending a different payload than the `Content-Type` header should return a `415 unsupported_media_type` error.
+> - Too large payload according to the limit should return a `413 payload_too_large` error 
+> - Wrong file encoding should return a `420 unprocessable_entity` error
+> - Invalid CSV data should return a `420 unprocessable_entity` error
```
bidoubiwa, line 94: Is this the right spec to talk about CSV? Or is it for context?

text/0029-indexing-ndjson.md
```diff
@@ -0,0 +1,132 @@
...
+##### Errors handling
+
+> - Sending a different payload than the `Content-Type` header should return a `415 unsupported_media_type` error.
+> - Too large payload according to the limit should return a `413 payload_too_large` error 
+> - Wrong file encoding should return a `420 unprocessable_entity` error
+> - Invalid NDJSON data should return a `420 unprocessable_entity` error
+
+### V. Impact on documentation
+
+This feature should impact MeiliSearch users documentation by adding mention of `ndjson` capability inside Documents scope at [Add or replace documents](https://docs.meilisearch.com/reference/api/documents.html#add-or-replace-documents) and [Add or update documents](https://docs.meilisearch.com/reference/api/documents.html#add-or-update-documents).
```
bidoubiwa, line 118:
```suggestion
This feature should impact MeiliSearch users documentation by adding the possibility to use `ndjson` as a accepted format in the  Documents scope at [Add or replace documents](https://docs.meilisearch.com/reference/api/documents.html#add-or-replace-documents) and [Add or update documents](https://docs.meilisearch.com/reference/api/documents.html#add-or-update-documents).
```

text/0029-indexing-ndjson.md
```diff
@@ -0,0 +1,132 @@
...
+
+We should also not only mention JSON format in `unsupported_media_type` section on the [errors page](https://docs.meilisearch.com/errors/#unsupported_media_type) and add `ndjson` format.
```
bidoubiwa, line 120: Not sure I understand this sentence

bidoubiwa (changes requested):
I changed it in some places but then realise it was used everywhere. I suggest changing the wording `ndjson file` to `ndjson body`. As file sending implies `'Multiple-form-data'` as a content-type. 
But what we  send is a text-body in a given format (in this case `ndjson`).

text/0029-indexing-ndjson.md
gmourier: Nice catch! I wanted to write NDJSON.

text/0029-indexing-ndjson.md
gmourier: https://docs.meilisearch.com/errors/#unsupported_media_type The error part only mentions the use of the JSON format -> "Currently, MeiliSearch supports only JSON payloads."

text/0029-indexing-ndjson.md
gmourier: I've updated that part.

gmourier:
I've updated this specification regarding similar points from @Kerollmops review on CSV file support.

MarinPostma (approved):

bidoubiwa (approved):

text/0029-indexing-ndjson.md
```diff
@@ -0,0 +1,138 @@
...
+
+### VI. Impact on SDKs
+
+This feature should impact MeiliSearch SDK's in the future by adding the possibility to send ndjson data to MeiliSearch on the previous explicited endpoints.
+
+## 2. Technical Aspects
+
+### I. Technical details
+
+⚠ A missing Content-Type will be interpreted as `application/json` since it's the current behavior. Giving an `application/json` Content-Type leads to the same behavior.
```
curquiza, line 134: Is it a technical detail? The users and the documentation should be aware of this. Technical details are for internal implementation from my POV.
This information is as important as any information in `Explanation`.

gmourier: Nice point @curquiza. I will update the specification according to this feedback!

curquiza (approved):

PR #30, opened 2021-04-14
ASC / DESC Criterion

gmourier:


text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,82 @@
...
+### VI. Impact on SDKs
+N/A
+
+## 2. Technical Aspects
+
+### I. Abstract
+TBD
+
+### II. Issues Summary
+TBD
```
curquiza, line 78:
```suggestion
To apply the ranking rule, the search engine needs to create a database. This database is the same as we create to apply facets and filters on an attribute.
However, the users will not pass the attribute into `attributeForFaceting` when setting a ranking rule. It means the search engine must create this database for the related attribute.

ex:
```json
{
  "rankingRules": ["words", "typo", "asc(price)", "proximity"],
  "attributesForFaceting": ["genre"]
}
```json

means the search engine has to create facet databases for `genre` and `price`.

⚠️ Following this example, it also means the search engine would be technically able to apply filters and facet distribution on `price`, however, we should prevent this. To avoid confusion, the search engine should prevent the users to execute a filter or get facet distribution on the ranking attributes. Only the ranking rule should be available for this field.
```

curquiza (changes requested):

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,82 @@
...
+✅ We have decided that Milli will act exactly as MeiliSearch. Since the rankings rules and the attributes for faceting do not stictly meet the same needs. Moreover this could confuse the users configuring the search engine. The criterion configuration will remain on the `rankingRrules` field of the [global settings endpoint](https://docs.meilisearch.com/reference/api/settings.html#get-settings) and in the 
+[specific ranking rules setting endpoint](https://docs.meilisearch.com/reference/api/ranking_rules.html).
+
+We wondered if Milli's current behavior, which is to discard documents from search results that do not have the attribute configured as ascending and descending criteria might make sense for the purpose of a user performing a search.
+
+✅ We have decided that Milli will act exactly as MeiliSearch concerning search results. The ASC / DESC criterion will no longer discard documents which do not have the attribute.
+
+### V. Impact on documentation
+
+- It should inform users about that having custom ASC / DESC criteria on attributes that are not always defined for each document may return results that are not necessarily relevant in the end user's eye.
```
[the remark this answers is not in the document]
gmourier: Thanks!

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,82 @@
...
+- To be able to use this criterion on a date format, it must indicate that the timestamp is to be preferred since the type string is not supported.
```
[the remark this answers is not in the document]
gmourier: Do we need to track documentation issues related to a spec?

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,91 @@
...
+```json
+{
+  "rankingRules": ["words", "typo", "asc(price)", "proximity"],
+  "attributesForFaceting": ["genre"]
+}
+```json
+
+means the search engine has to create facet databases for `genre` and `price`.
+
+⚠️ Following this example, it also means the search engine would be technically able to apply filters and facet distribution on `price`, however, we should prevent this. To avoid confusion, the search engine should prevent the users to execute a filter or get facet distribution on the ranking attributes. Only the ranking rule should be available for this field.
```
gmourier, line 87: Can it be the behavior for v0.21 ?

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,82 @@
...
+- Number of words matching exactly (without typo)
+- Custom ranking
+
+Note that Algolia don't recommend to change the order of the default criteria because of the fact that it works for the vast majority of their use cases. Like MeiliSearch (v0.20), document that don’t have an attribute that’s in the customRanking list are pushed to the bottom of the search result.
+
+### IV.Explanation
+
+#### Current behavior of v0.20
+
+It returns documents that don't have the attribute set on a custom ASC / DESC rule as a search result at the bottom. Documents are not excluded and can be returned in the search result.
```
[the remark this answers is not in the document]
gmourier: I've added more clarity by pointing out this specific case.

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,91 @@
...
+
+Ranking rules are built-in rules that ensure relevancy in search results. Ranking rules are applied in a default order which can be changed in the settings. You can add or remove rules and change their order of importance.
+
+MeiliSearch allows you to create custom rules within the default rules. Custom rules are dedicated to sorting in ascending or descending order on an attribute.
+
+### II. Motivation
+
+We want to provide our users with an always improved usage experience. Relevance is essential in a search engine since it is what allows the engine to fulfill its objective. Delivering results that match user demands by allowing modification of the relevance is critical.
+
+The new search engine called Milli no longer processes this criterion of relevance as the current MeilliSearch. This specification makes it possible to state that Milli will be identical to the ISO version (v0.20) in the usage of API interfaces but also in the expected search results.
```
bidoubiwa, line 18:
```suggestion
The new search engine called Milli no longer processes this criterion of relevance the same way as the current MeilliSearch. This specification aims to make Milli identical both in API usage and in expected search results as the current ISO version (v0.20).
```

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,91 @@
...
+- Number of typos
+- Geolocation
+- Number of words in the query matching in the result
+- Filters
+- Distance between words
+- Best matching attribute in the record
+- Number of words matching exactly (without typo)
+- Custom ranking
+
+Note that Algolia don't recommend to change the order of the default criteria because of the fact that it works for the vast majority of their use cases. Like MeiliSearch (v0.20), document that don’t have an attribute that’s in the customRanking list are pushed to the bottom of the search result.
```
bidoubiwa, line 35:
```suggestion
Note that Algolia doesn't recommend changing the order of the default criteria because of the fact that it works for the vast majority of their use cases. Like MeiliSearch (v0.20), documents, where the attribute specified in the custom ranking is missing, are pushed to the bottom of the search results.
```

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### Current behavior of Milli
+
+If the attribute on which the ASC / DESC criterion is configured does not exist in all the documents of the index concerned by the search, Milli will discard them and never return them. This behavior is not necessarily what a end user would expect when searching.
+
+As v0.20, Milli can’t handle ASC / DESC criterion on string. Only numbers allow its use.
+
+#### Decisions
+
+1. We thought about allowing the use of the ASC / DESC criterion in the ranking rules settings as long as it would also be declared in the attributesForFaceting.
```
bidoubiwa, line 55: I'm not sure this is what I remember. 
Was it not: 
- You can add any attribute as `ASC / DESC` custom ranking rule, but they will not be considered filterable. 
- In order for an attribute that is present in `ASC / DESC ` custom ranking rule to be filterable, it has to be added in `attributesForFaceting` as well. 

Maybe I'm wrong :s

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,91 @@
...
+```json
+{
+  "rankingRules": ["words", "typo", "asc(price)", "proximity"],
+  "attributesForFaceting": ["genre"]
+}
+```json
+
+means the search engine has to create facet databases for `genre` and `price`.
+
+⚠️ Following this example, it also means the search engine would be technically able to apply filters and facet distribution on `price`, however, we should prevent this. To avoid confusion, the search engine should prevent the users to execute a filter or get facet distribution on the ranking attributes. Only the ranking rule should be available for this field.
```
bidoubiwa, line 87:
```suggestion
⚠️ Following this example, it also means the search engine would be technically able to apply filters and facet distribution on `price`, however, we should prevent this. To avoid confusion, the search engine should prevent the users to execute a filter or get facet distribution on the ranking attributes. Only the ranking rule should be available for this field.

If the user wants to filter on that attribute, he will have to add it in `attributesForFaceting` as well. 
```

bidoubiwa (changes requested):

text/0030-asc-desc-criterion.md
gmourier: Yes, these are the final decisions. We were wondering if it needed to be declared in attributesForFaceting to be set as a custom ranking rule. The fact that it's not filterable if the field is not set in attributesForFaceting is for clarity and convenience in the API usage.

bidoubiwa: Ah oke! I see 
```suggestion
1. We thought about allowing filtering of the attributes used in the ASC / DESC as long as it would also be declared in the attributesForFaceting.
```

This seems to be more clear but I would rephrase it like that.
```suggestion
1. Attributes used in ASC/DESC criterion doesn't make them filterable during search. They should be added in `attributesForFaceting` to filter on it.
```

What do you think?

gmourier: I am going to add the first proposition because it shows that we have thought about this subject. In addition, it follows the convention of the second question mark just below. Thanks @bidoubiwa!

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,95 @@
...
+- Number of words matching exactly (without typo)
+- Custom ranking
+
+Note that Algolia doesn't recommend changing the order of the default criteria because of the fact that it works for the vast majority of their use cases. Like MeiliSearch (v0.20), documents, where the attribute specified in the custom ranking is missing, are pushed to the bottom of the search results.
+
+### IV.Explanation
+
+#### Current behavior of v0.20
+
+In the case of certain documents that do not contain the attribute configured on a custom ASC/DESC ranking rules, it will return those documents at the bottom of the search result. Documents are not excluded and can be returned in the search result.
```
curquiza, line 41: It's not really true. As we discussed, the end of the bucket is returned, meaning you can have this behavior (for a custom ranking rule on `price`):
- 10 results with `price` attributes
- 2 results without `price`
- 8 results with `price`
The end of the bucket is returned.

curquiza, line 41: It's not really true. As we discussed, the end of the bucket is returned, meaning you can have this behavior (for a custom ranking rule on `price`):
- 10 results with `price` attributes
- 2 results without `price`
- 8 results with `price`

The end of the bucket is returned.

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,95 @@
...
+
+> In the case of several custom ASC / DESC rules configured within the rankings rules, the sorting behavior is difficult to explain and can appear undefined when the documents do not necessarily contain the attributes which are set on this rule.
```
curquiza, line 43: This behavior can also appear when there is only one custom ranking rule

curquiza (changes requested):
Sorry to have missed this!

text/0030-asc-desc-criterion.md
gmourier: I've added some examples to show the behavior given custom-ordered ranking rules set.

text/0030-asc-desc-criterion.md
gmourier: Yes! The newly added examples show that.

curquiza (approved):
Nice explanation!

bidoubiwa (approved):

text/0030-asc-desc-criterion.md
```diff
@@ -0,0 +1,247 @@
...
+#### Decisions
+
+1. We thought about allowing filtering of the attributes used in the ASC / DESC criterion as long as it would also be declared in the `attributesForFaceting` parameter.
+
+✅ We have decided that Milli will act exactly as MeiliSearch: no need to declare the ranking rule field in `attributeForFaceting` since the rankings rules and the attributes for faceting do not strictly meet the same needs. Moreover, this could confuse the users configuring the search engine. The criterion configuration will remain on the `rankingRrules` field of the [global settings endpoint](https://docs.meilisearch.com/reference/api/settings.html#get-settings) and in the
+[specific ranking rules setting endpoint](https://docs.meilisearch.com/reference/api/ranking_rules.html).
+
+2. We wondered if Milli's current behavior, which is to discard documents from search results that do not have the attribute configured as ascending and descending criteria might make sense for the purpose of a user performing a search.
+
+✅ We have decided that Milli will act exactly as MeiliSearch concerning search results. The ASC / DESC criterion will no longer discard documents which do not have the attribute.
```
qdequele, line 214: 🎉

qdequele (approved):

MarinPostma (approved):

PR #39, opened 2021-05-05
_formatted Field Behavior

gmourier:
Initialize _formatted field behavior specification.

### DIFF
--- a/text/0039-_formatted-field-behavior.md
+++ b/text/0039-_formatted-field-behavior.md
@@ -19,7 +19,7 @@ The goal of this specification is to clear the behavior of `attributesToRetrieve
 
 #### Algolia
 
-By default, Algolia returns `_hightlightResult` even if no `attributesToHighlight` are set at query time. So, by default the value is `'*'`.
+By default, Algolia returns `_hightlightResult` even if no `attributesToHighlight` are set at query time. So, by default the value is `*`.
 Setting a specific attribute in `attributesToHighlight` will only give this specific attribute in `_highlightResults`.
 
 Unlike highlighting, snippeting must be proactively enabled for each attribute to snippet, however Algolia authorize the usage of `*` to snippet all attributes.
@@ -29,46 +29,84 @@ Setting a specific attribute in `attributesToSnippet` will only give this specif
 
 #### Current MeiliSearch Behavior (0.20)
 
-The `_formatted` field appears in the search response as soon as the `attributesToHighlight` and/or `attributesToCrop` parameters are sent as a query parameter and are filled with a parameter or as an empty parameter.
-
-`attributesToRetrieve` is independent of `_formatted` and so of `attributesToHighlight` and `attributesToCrop`.
-
-`attributesToHighlight` and `attributesToCrop` have an unwanted behavior on `_formatted` result:
-
 Given a document made of two fields `title` and `poster`
-```json
+```
 {
     "title": "Prince Avalance",
     "poster": "https://image.tmdb.org/t/p/w1280/3KHiQt54usbHyIjLIMzaDAoIJNK.jpg"
 }
 ```
 
+**Example 1**
+
 Given these search parameters:
-```json
+```
 {
     "q": "prince",
-    "attributesToHighlight": ["title"]
+    "attributesToRetrieve": ["*"]
 }
 ```
 
-As a user i expect to have this response:
+As a user i get:
 
-```json
+```
+{
+    "hits": [
+        {
+            "title": "Prince Avalanche",
+            "poster": "https://image.tmdb.org/t/p/w1280/3KHiQt54usbHyIjLIMzaDAoIJNK.jpg"
+        }
+    ]
+}
+```
+
+Since `attributesToHighlight` and `attributesToCrop` are not set, `_formatted` is not computed.
+
+
+**Example 2**
+
+Given these search parameters:
+```
+{
+    "q": "prince",
+    "attributesToRetrieve": ["*"],
+    "attributesToHighlight": ["wrongFieldName"]
+}
+```
+
+As a user i get:
+
+```
 {
     "hits": [
         {
             "title": "Prince Avalanche",
             "poster": "https://image.tmdb.org/t/p/w1280/3KHiQt54usbHyIjLIMzaDAoIJNK.jpg",
             "_formatted": {
-                "title": "<em>Prince</em> Avalance"
+                "title": "Prince Avalance",
+                "poster": "https://image.tmdb.org/t/p/w1280/3KHiQt54usbHyIjLIMzaDAoIJNK.jpg"
             }
         }
     ]
 }
 ```
 
-but i get:
-```json
+The `_formatted` field appears in the search response as soon as the `attributesToHighlight` and/or `attributesToCrop` parameters are sent as a query parameter and are filled with a value representing an existent or inexistent field. Using an inexistent field is similar to setting `attribuesToHighlight`/`attributesToCrop` to `"*"`
+
+**Example 3**
+
+Given these search parameters:
+```
+{
+    "q": "prince",
+    "attributesToRetrieve": ["*"],
+    "attributesToHighlight": ["title"]
+}
+```
+
+As a user i get:
+
+```
 {
     "hits": [
         {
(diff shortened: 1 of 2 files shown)


text/0039-_formatted-field-behavior.md
```diff
@@ -0,0 +1,136 @@
...
+✅  `_formatted` response parameter **SHOULD NOT** be returned in the given cases:
+
+- `attributesToHighlight` is set but is empty AND `attributesToCrop` is set but is empty.
+- `attributesToHighlight` is set but is empty AND `attributesToCrop` isn't set.
+- `attributesToHighlight` is set but only contains a non-existent field AND `attributesToCrop` is set but only contains a non-existent field.
+- `attributesToHighlight` is set but only contains a non-existent field AND (`attributesToCrop` is not set OR `attributesToCrop` is set but is empty).
+
+✅ `_formatted` response parameter **SHOULD** be returned in any cases of:
+
+- `attributesToHighlight` is set with `*`. When `attributesToHighlight` is not set, the engine will use `*` as default parameter. So each `displayedAttributes` will be in the `_formated` result in that specific case.
```
curquiza, line 111: > `attributesToHighlight` is set with `*`. When `attributesToHighlight` is not set, the engine will use `*` as default parameter.

If `attributesToHighlight` is set, it means the users don't want highlight, so `*` should not be the default behavior, right?

curquiza (changes requested):

text/0039-_formatted-field-behavior.md
```diff
@@ -0,0 +1,136 @@
...
+- `attributesToCrop` is set with `*`.
+- `attributesToHighlight` is set with **1..N** existing attributes.
+- `attributesToCrop` is set with **1..N** existing attributes.
+
+
+> ⚠️ Note that if one of the two parameters is requested but empty or contains an attribute that does not exist but the other is valid, `_formatted` will be returned based on the valid parameter.
```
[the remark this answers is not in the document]
gmourier: `attributesToHighlight` and `attributesToCrop`in this case!

text/0039-_formatted-field-behavior.md
```diff
@@ -0,0 +1,136 @@
...
+
+> ⚠️ Note that `attributesToRetrieve` is independent of `_formatted` and so of `attributesToHighlight` and `attributesToCrop`.
```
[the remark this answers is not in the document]
gmourier: It seems to me that this is the easiest way for users to understand how to have what they decide to get as raw results and choose what they want to crop or highlight. Not mixing these parameters permits us to keep an easy way of describing the rules between each of these parameters. If the engine decides for them to mix these fields, we will need to explain each possible case and it seems complicated on so many levels. Moreover, they do not serve the same purpose.

@bidoubiwa and I have discussed that and that's why by default we want to highlight all the fields to keep an easy way for our users that do not read the documentation before using MeiliSearch to have a way to display and use results from `_formatted` field. @bidoubiwa explained to me that it will be easier for them to use MeiliSearch at first to directly get results from `_formatted` instead of merging some fields from `hits` and `formatted` by hand. By doing we allow keeping the engine an independent behavior between `attributesToRetrieve` and `_formatted` but we don't reduce the bootstrapping MeiliSearch experience.

For users that do not want to retrieve the whole possible fields, I've added a part in future possibilities to facilitate the removing of some fields instead of saying 'I want only this field and this field.. and so on' by using the `-fieldname` operator. E.g. If I want all the fields except one, I can write `*,-fieldname`. By removing highlighting on certain fields instead of typing all fields that the user may want,  it gives a better experience for the use case of there is a lot of fields in documents.

I'm happy to discuss that with you at any time for clarity!

gmourier:
I updated this specification after further testing the behavior of 0.20 and the current state of 0.21. The goal is to keep the behavior of 0.20 by eliminating the inconsistencies encountered

MarinPostma (approved):

text/0039-_formatted-field-behavior.md
```diff
@@ -0,0 +1,473 @@
...
+            "_formatted": {
+                "title": "<em>Prince</em> Avalanche"
+            }
+        }
+    ]
+}
+```
+
+Unlike v0.20, `_formatted` is only containing fields set in `attributesToHighlight` and `attributesToCrop`. Independently from `attributesToRetrieve` value.
+
```
curquiza, line 366: This means, contrary to the v0.20.0, there is no way for the front-end developer to have all the attributes (highlighted and non-highlighted) without merging `_formatted` with the other attributes. It's not complicated to do, but we clearly lose dev experience for front-end developers with this choice. I remember charlotte talked about this with you.
Is it what we want?

text/0039-_formatted-field-behavior.md
```diff
@@ -0,0 +1,473 @@
...
+                "actor": "<em>Prince</em>"
+            }
+        }
+    ]
+}
+```
+
+Same behavior as v0.20 release.
+
+**Example 5**
```
curquiza, line 337: Can you precise in this example that put `attributesToRetrieve` as parameter is equivalent to not set `attributesToRetrieve` at all since the default value of `attributesToRetrieve` is `["*"]`?
It might be obvious but it is an important thing we should not forget

text/0039-_formatted-field-behavior.md
gmourier: This is the current behavior of 0.21. I've made a proposal to change that and have the behavior that you explain here. This is in the "Expected MeiliSearch Behavior" part below. Line 402 for an explanation and Line 432 to 459 for an example.

text/0039-_formatted-field-behavior.md
gmourier: I will add that part in the Expected MeiliSearch Behavior part! Thanks

text/0039-_formatted-field-behavior.md
```diff
@@ -0,0 +1,474 @@
...
+
+## 1. Functional Specification
+
+### I. Summary
+
+`_formatted` is used in conjunction with raw search results to highlight and/or crop around the query term in the attributes of the document. `_formatted` main goal is to enhance the UI and the UX by providing a nice way to catch the user's eyes on the front-end side while searching.
+
+### II. Motivation
+
+The goal of this specification is to clear the behavior of `attributesToRetrieve`, `attributesToHighlight` and `attributesToCrop` on `_formatted` response parameter content.
```
bidoubiwa, line 16:
```suggestion
The goal of this specification is to clarify the behavior of `attributesToRetrieve`, `attributesToHighlight` and `attributesToCrop` on `_formatted` response parameter content.
```

text/0039-_formatted-field-behavior.md
```diff
@@ -0,0 +1,474 @@
...
+}
+```
+
+Unlike v0.20, `attributesToHighlight` set fields to be in `_formatted` independently from `attributesToRetrieve`.
+
+#### Expected MeiliSearch Behavior (0.21)
+
+✅ If `attributesToRetrieve` is not set as a parameter, the expected behavior is the same as if `attributesToRetrieve` is equal to `*`.
+✅ If `attributesToHighlight` and `attributesToCrop` are not set, do not return `_formatted` and don't compute highlights and crops.
+✅ If cumulated fields in `attributesToHighlight` and `attributesToCrop` resolve to only having non-existent fields, do not return `_formatted`.
```
bidoubiwa, line 401: Could we send back an empty `_formatted`?  To avoid key checking? Not sure this is the place, maybe it should be determined for a later version. 

```
{
  "title": "hello",
  "_formatted": {}
}
```

text/0039-_formatted-field-behavior.md
curquiza: I miss read! Sorry!! I thought it was "what you expect for the v0.21.0" 😁

bidoubiwa (changes requested):

text/0039-_formatted-field-behavior.md
gmourier: Thanks, @bidoubiwa! Can you explain a bit more about avoiding key checking? Btw, your proposal can be added to this specification iteration. :)

curquiza: @gmourier can you ping me if there is any change like this, I will open an issue once this spec is definitely validated

text/0039-_formatted-field-behavior.md
```diff
@@ -0,0 +1,474 @@
+- Title: _formatted Field Behavior
+- Start Date: 2021-05-05
+- Specification PR: [#39](https://github.com/meilisearch/specifications/pull/39)
+- MeiliSearch Tracking-Issues:
```
curquiza, line 4:
```suggestion
- MeiliSearch Tracking-Issues: [transplant/#203](https://github.com/meilisearch/transplant/issues/203)
```

text/0039-_formatted-field-behavior.md
```diff
@@ -0,0 +1,474 @@
...
+                "title": "<em>Prince</em> Avalanche",
+                "actor": "<em>Prince</em>,
+                "poster": "https://image.tmdb.org/t/p/w1280/3KHiQt54usbHyIjLIMzaDAoIJNK.jpg"
+            }
+        }
+    ]
+}
+```
+
+✅ If `attributesToHighlight` or `attributesToCrop` contains a field that is not declared in `attributesToRetrieve`, it his added to `_formatted` and it is highlighted and/or cropped.
```
curquiza, line 460:
```suggestion
✅ If `attributesToHighlight` or `attributesToCrop` contains a field that is not declared in `attributesToRetrieve`, it is added to `_formatted` and it is highlighted and/or cropped.
```

text/0039-_formatted-field-behavior.md
curquiza: I'm coding this part right now and this specific behavior would make the code more complex than it's already. Since the `_formatted` is currently a non-easy to maintain part in the code, I would rather we stay with "no `_formatted` if the fields are not valid".
If I found out a better solution to implement this, I'll come back to you and will update this spec :)

gmourier: Hello @curquiza! I'm sorry! I misspoke above when I said, "this can be added to the spec". I should have said that the specification was not merged and could still evolve! I was waiting for more information about the need expressed by @bidoubiwa about "avoiding key checking" before changing anything in the speicifcation! Anyway, thanks for the information 🙂

curquiza (approved):
🚀

text/0039-_formatted-field-behavior.md
bidoubiwa: Well, I cannot remember the reason why. So lets forget about it 🔥

bidoubiwa (approved):
🔥🔥🔥

PR #40, opened 2021-05-10
Pagination Convention

gmourier:
Initialize pagination specification as a convention guide for future standard paginated routes.

### DIFF
--- a/text/0040-pagination-convention.md
+++ b/text/0040-pagination-convention.md
@@ -22,7 +22,7 @@ N/A
 
 ##### Requirements
 
-- ✅ Each paginable route must return the total number of items it can render to the user at the end and the current number of items returned in the response. E.g. `total` and `size`. In the case of multiples pages, all other pages except the last one will have a `size = limit`.
+- ✅ Each paginable route must return the `total` number of items that match the query and the number of items returned in the response as `size`.
 - ✅ Each paginable route should be able to handle `offset` and `limit` to return a specific page.
 - ✅ Each paginable route should handle a default `limit`. E.g. `20`
 - ✅ Each paginable route should handle a default `offset` parameter with a value of `0`.
--- a/text/0040-pagination-convention.md
+++ b/text/0040-pagination-convention.md
@@ -28,7 +28,7 @@ E.g. [GET `/indexes/:index_uid/updates`](https://docs.meilisearch.com/reference/
 
 ##### Request
 
-`{{url}}/indexes/{{indexUID}}/updates`
+`/indexes/:index_uid/updates`
 
 > For the example, let's imagine that the default limit is `20`.
 
@@ -123,7 +123,7 @@ E.g. [GET `/indexes/:index_uid/updates`](https://docs.meilisearch.com/reference/
 
 ##### Request
 
-`{{url}}/indexes/{{indexUID}}/updates?offset=0&limit=2`
+`/indexes/:index_uid/updates?offset=0&limit=2`
 
 ##### Response
 ```
@@ -181,7 +181,7 @@ E.g. [GET `/indexes/:index_uid/updates`](https://docs.meilisearch.com/reference/
 
 ##### Request
 
-`{{url}}/indexes/{{indexUID}}/updates?offset=2&limit=2`
+`/indexes/:index_uid/updates?offset=2&limit=2`
 
 ##### Response
 
@@ -234,9 +234,9 @@ E.g. [GET `/indexes/:index_uid/updates`](https://docs.meilisearch.com/reference/
 
 ##### Request
 
-`{{url}}/indexes/{{indexUID}}/updates?status=processed&offset=50&limit=10`
+`/indexes/:index_uid/updates?status=processed&offset=50&limit=10`
 
-> For the example, let's imagine the total number of processed update is `60` and want page of `10` items.
+> For the example, let's imagine the total number of processed updates is `60` and we want page of `10` items.
 
 ##### Response
 ```
--- a/text/0040-pagination-convention.md
+++ b/text/0040-pagination-convention.md
@@ -29,77 +29,6 @@ N/A
 - ✅ `limit` and `offset` are number parameters.
 - ✅ If `limit` and `offset` can't return any results, the API should not return an error. It's preferable to return empty results in this case.
 
-##### Going further
-
-- Provide a `_links` object with `current`, `next` and `prev` elements to facilitate pagination on client side by pre-computing `next` and `prev` urls parameters given the pagination context.
-
-
-E.g. Result from `GET /indexes/myindex/updates/` at `page 1`.
-
-```json
-{
-    "hits": [
-        {...},
-        {...},
-        {...},
-        {...},
-        {...}
-    ],
-    "total": 50,
-    "size": 5,
-    "offset": 5,
-    "limit": 5,
-    "_links": {
-        "current": "https://0.0.0.0:7070/indexes/myindex/updates?offset=0&limit=5",
-        "next": "https://0.0.0.0:7070/indexes/myindex/updates?offset=5&limit=5",
-        "prev": null
-    }
-}
-```
-
-E.g. Result from `GET /indexes/myindex/updates/` at `page 2`.
-```json
-{
-    "hits": [
-        {...},
-        {...},
-        {...},
-        {...},
-        {...}
-    ],
-    "total": 50,
-    "size": 5,
-    "offset": 5,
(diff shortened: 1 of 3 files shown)


gmourier:
This specification will act as a convention guide for future standard paginations. E.g. Everything that is not a paginated search.

text/0040-pagination-convention.md
```diff
@@ -0,0 +1,118 @@
...
+This specification will be used as a convention to implement simple and standard paginations to be set up on API routes such as the list of updates for example. Since the pagination of search results is intrinsically linked to technical implementations of the search engine, this specific case is detailed in its own specification.
+
+### III. Additional Materials
+N/A
+
+### IV. Explanation
+
+##### Requirements
+
+- ✅ Each paginable route must return the total number of items it can render to the user at the end and the current number of items returned in the response. E.g. `total` and `size`. In the case of multiples pages, all other pages except the last one will have a `size = limit`.
```
bidoubiwa, line 25: I'm not sure if you are suggesting this becomes a default behavior and not an activable option?

gmourier: Can you please tell me which part makes you think of an option that can be activated on demand? Having `total` and `size` can be the default behavior on standard paginated resources in the response body. `size` represents the number of items in the search result for the current page and `total` represents the total number of items to be segmented on the whole pages. `offset` and `limit` parameters are used to move between pages.

text/0040-pagination-convention.md
```diff
@@ -0,0 +1,118 @@
...
+- ✅ Each paginable route should be able to handle `offset` and `limit` to return a specific page.
+- ✅ Each paginable route should handle a default `limit`. E.g. `20`
+- ✅ Each paginable route should handle a default `offset` parameter with a value of `0`.
+- ✅ `limit` and `offset` are number parameters.
+- ✅ If `limit` and `offset` can't return any results, the API should not return an error. It's preferable to return empty results in this case.
+
+##### Going further
+
+- Provide a `_links` object with `current`, `next` and `prev` elements to facilitate pagination on client side by pre-computing `next` and `prev` urls parameters given the pagination context.
```
qdequele, line 34: Do you think it really helps the user? I have the impression that it is not used too much by API consumers. If I take the example of the [Stripe API](https://stripe.com/docs/api/pagination), the only information they add in the response for paging is the `has_more`.

gmourier: Thinking about it, the advantage is that it could be easier to build a pagination on "simple" frontends. The links being directly accessible, the pagination rendering component would only have to take and assign the prev, next on related pagination buttons. However, this does not meet all types of pagination and some frontend frameworks may not be able to use the links directly but have to reform this preconceived url from the state variables.

As we don't have any direct feedback on this subject, I will only leave this mention within the future possibilities.

text/0040-pagination-convention.md
```diff
@@ -0,0 +1,47 @@
...
+This specification will be used as a convention to implement simple and standard paginations to be set up on API routes such as the list of updates for example. Since the pagination of search results is intrinsically linked to technical implementations of the search engine, this specific case is detailed in its own specification.
+
+### III. Additional Materials
+N/A
+
+### IV. Explanation
+
+##### Requirements
+
+- ✅ Each paginable route must return the `total` number of items that match the query and the number of items returned in the response as `size`.
```
curquiza, line 25: I don't get this one, when I read the previous spec about pagination, I understand `nbHits` will return the number total of items matching the query

text/0040-pagination-convention.md
```diff
@@ -0,0 +1,47 @@
...
+### II. Motivation
+
+This specification will be used as a convention to implement simple and standard paginations to be set up on API routes such as the list of updates for example. Since the pagination of search results is intrinsically linked to technical implementations of the search engine, this specific case is detailed in its own specification.
+
+### III. Additional Materials
+N/A
+
+### IV. Explanation
+
+##### Requirements
```
curquiza, line 23: How to trigger the pagination? I guess it's explained in the previous spec, but nothing mentions it here if I'm correct.
Why do we need 2 specs? 🙂 I'm sure there is a reason, but I feel a little bit confused at the moment.

text/0040-pagination-convention.md
curquiza, line 25: What are "each paginable route"? Get and post /search routes?

text/0040-pagination-convention.md
```diff
@@ -0,0 +1,47 @@
...
+
+## 1. Feature Description and Interaction
+
+### I. Summary
+
+Pagination allows the content to be viewed in the form of a sequence in order to facilitate the browsing of its content by not overloading the number of elements to be viewed for the user. Pagination also has the advantage of easily providing the total number of items that remain to be traversed by not overloading the network with all the content that can be loaded.
+
+### II. Motivation
+
+This specification will be used as a convention to implement simple and standard paginations to be set up on API routes such as the list of updates for example. Since the pagination of search results is intrinsically linked to technical implementations of the search engine, this specific case is detailed in its own specification.
```
curquiza, line 16: Is it here we should note the different types of pagination we could expect?
- infinite scroll: no need to know if the next page exists -> only `offset` and `limit` are useful in their "natural" usage
- "infinite" pagination: the users can access the next page and are aware of the existence of this page in front end. The developer needs to check if a next page exists by using `limit + 1`
- "finite" pagination : the users can access the next page and the last page. `forceExhaustivity` needs to be used.

curquiza: The SDK should provide these 3 ways of pagination as smoothly as possible

text/0040-pagination-convention.md
gmourier: All paginable routes except GET and POST /search. E.g. `/updates`, `/documents`. This specification concerns all routes that can be paged and that are not related to the search endpoints.

text/0040-pagination-convention.md
gmourier: `forceExhaustivity` is only available on the `search` endpoint because of the internal data structures and algorithms of the engine. The current specification is intended for routes that always return a finite number of documents and where we can compute a `total` easily. All type of pagination like infinite, page based can be handled by that convention.

text/0040-pagination-convention.md
gmourier: Pagination is enabled through the use of offset and limit. By default, we can say that the paging is always triggered since we return a limited number of results from the first call.

text/0040-pagination-convention.md
gmourier: `nbHits` is specific to `/search` routes. It cannot always return an exhaustive number, hence the use of a parameter to force its completeness in a finite pagination search context.

Here we are on different routes which do not integrate the notion of hits (or candidates). See `/documents` which returns a finite set of documents, it is enough to add `size` and `total` in the response to be able to do pagination simply as a user.

curquiza (approved):

text/0040-pagination-convention.md
curquiza, line 23: For some routes, like `/updates`, I would rather have the last update sent at the top of the list rather than at the end. But maybe I'm the only one.

bidoubiwa (approved):

text/0040-pagination-convention.md
gmourier: Yes, this example is to illustrate the pagination convention. A future specification on `/updates` will allow to define this specific behavior for the update API route. It could also be an asc/dec filter for example.

text/0040-pagination-convention.md
```diff
@@ -0,0 +1,312 @@
...
+    "offset": 50,
+    "limit": 10,
+    "total": 60,
+    "size": 10
+]
+```
+
+#### Requirements
+
+- ✅ Each paginable route must return the `total` number of items that match the query and the number of items returned in the response as `size`.
```
MarinPostma, line 290: can we consider using `count` instead of `size`?

gmourier: I hesitated between `count` and `size`. After thinking about it I thought that `count` could be misunderstood with `total`. In the context of a page, I thought that page size makes more sense than page count.

text/0040-pagination-convention.md
curquiza: Awesome! Thanks :)

gmourier:
I'm not going to merge this one right now because we should have one concerning the response format of our API. This one could evolve as a consequence of the response format. E.g. eventually integrate a block dedicated to pagination instead of outputting the pagination parameters at the first level of the response.

PR #43, opened 2021-06-02
Phrase Query

gmourier:
Initialize Phrase Query specification

### DIFF
--- a/text/0043-phrase-query.md
+++ b/text/0043-phrase-query.md
@@ -94,6 +94,8 @@ Using the standard query parameter syntax as `q` equals to `Plays and Playwright
 
 Using the Phrase Query syntax this way, with `q` equals to `"Plays and Playwrights 2002"`, will lead to have only one result because the title is exactly written like that. No typo tolerance nor prefix search were applied on the query terms at search time.
 
+> Note that it's case insensitive. So, if i search with `"plays and playwrights 2002"`, this will lead to the same result.
+
 ```
 {
     "hits": [
--- a/text/0043-phrase-query.md
+++ b/text/0043-phrase-query.md
@@ -96,6 +96,8 @@ Using the standard query parameter syntax as `q` equals to `Plays and Playwright
 }
 ```
 
+#### Phrase Query usage
+
 To use the Phrase Query syntax, simply surround the contiguous search terms with the characters `"`.
 
 Using the Phrase Query syntax this way, with `q` equals to `"Plays and Playwrights 2002"`, will lead to have only one result because the title is written exactly like that.
@@ -165,11 +167,25 @@ The query can be expressed like this: `"African American" poem`
 
 As you can see in the results, the presence or absence of one or more soft separators such as `-`, `_`, `\`, `:`, `/`, `\\`, `@`, `"`, `+`, `~`, `=`, `^`, `*`, `#` between two words does not affect the query match of the phrase and the document.
 
+#### Multiple Phrase Query usage
+
 It is possible to use multiple phrase queries within a search.
 
 E.g. `"African American" "Anthology"`
 
 With this expression, the returned documents will contain exactly the existence of both phrase queries.
+
+#### Know limitations
+
+##### Multiple hard separator case
+
+Given a document containing `David.- .- .- .- .-Bowie` as value for an attribute.
+
+This document can be matched using a phrase query such as `"David" "Bowie"` or `"David.Bowie"`. At the engine level, this is the same query. The latter is translated into the former phrase because the `.` character is a hard separator. This behavior comes from the default tokenizer, hard separators are seen as a marker for a different context or phrase.
+
+Here is the list of hard separators in the default tokenizer: `.`, `;`, `,`, `!`, `?`, `(`, `)`, `[`, `]`, `{`, `}`, `|`
+
+Multiple hard separator are treaten the same as if they were one. So `"David.-.-.-.Bowie"` will not work to match the document.
 
 ### V. Impact on Documentation
 
--- a/text/0043-phrase-query.md
+++ b/text/0043-phrase-query.md
@@ -13,7 +13,7 @@ MeiliSearch does not allow users a way to write a strict query in order to ask t
 
 ### II. Motivation
 
-This feature is driven by user needs. Indeed, let's take an example recently brought up in our community Slack.
+This feature is driven by the user's needs. Indeed, let's take an example recently brought up in our community Slack.
 
 The user in question would like to be able to retrieve specifically the document containing the unique ISBN identifier and only that one. In a UX context of type as you search, this is impossible today without impacting the UI/UX or finding a workaround.
 
--- a/text/0043-phrase-query.md
+++ b/text/0043-phrase-query.md
@@ -9,11 +9,11 @@
 
 ### I. Summary
 
-MeiliSearch does not allow users a way to write a strict query in order to ask the engine to be more drastic in its selection of candidates for search results. The Phrase Query feature adds a simple syntax available to users to require the engine to select documents that will strictly match the search field value. That is, without typography, n-gram, wordsplit, prefix, synonym, and, proximity. In addition, the expression of a Phrase Query is case insensitive.
+MeiliSearch does not allow users a way to write a strict query in order to ask the engine to be more strict in its selection of candidates for search results. The Phrase Query feature adds a simple syntax available to users to require the engine to select documents that strictly match some phrase, indicated by quotation marks. That is, without typography, n-gram, wordsplit, prefix, synonym, and, proximity. In addition, the expression of a Phrase Query is case insensitive.
 
 ### II. Motivation
 
-This feature is driven by user needs. Indeed, let's take an example recently brought up in our community slack.
+This feature is driven by user needs. Indeed, let's take an example recently brought up in our community Slack.
 
 The user in question would like to be able to retrieve specifically the document containing the unique ISBN identifier and only that one. In a UX context of type as you search, this is impossible today without impacting the UI/UX or finding a workaround.
 
@@ -29,7 +29,7 @@ The Phrase Query feature will easily solve this case but will also adapt to the 
 
 Algolia allows the use of Phrase Query syntax as long as the `advancedSyntax` parameter is set to true in the settings.
 
-As Algolia documentation said, a phrase query represents a specific sequence of terms that must be matched next to one another. A phrase query needs to be surrounded by double quotes ("). For example, the query "search engine" only returns a record if it contains “search engine” exactly in at least one attribute.
+As Algolia documentation said, a phrase query represents a specific sequence of terms that must be matched next to one another and in the given order. A phrase query needs to be surrounded by double quotes ("). For example, the query "search engine" only returns a record if it contains “search engine” exactly in at least one attribute.
 
 #### Elasticsearch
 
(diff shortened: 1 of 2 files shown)


text/0043-phrase-query.md
```diff
@@ -0,0 +1,161 @@
...
+- Specification PR: [#43](https://github.com/meilisearch/specifications/pull/43)
+- MeiliSearch Tracking-Issues:
+
+# Phrase Query
+
+## 1. Feature Description and Interaction
+
+### I. Summary
+
+Meilisearch does not allow users a way to write a strict query in order to ask the engine to be more drastic in its selection of candidates for search results. The Phrase Query feature adds a simple syntax available to users to require the engine to select documents that will strictly match the search field value. That is, without typography, without prefix and, in the order of the given terms.
```
curquiza, line 12:
```suggestion
MeiliSearch does not allow users a way to write a strict query in order to ask the engine to be more drastic in its selection of candidates for search results. The Phrase Query feature adds a simple syntax available to users to require the engine to select documents that will strictly match the search field value. That is, without typography, without prefix and, in the order of the given terms.
```

curquiza: Also, without n-gram (gather two words) and split words (trying to split the word into 2 other words)

text/0043-phrase-query.md
```diff
@@ -0,0 +1,161 @@
...
+
+### II. Motivation
+
+This feature is driven by user needs. Indeed, let's take an example recently brought up in our community slack.
+
+The user in question would like to be able to retrieve specifically the document containing the unique ISBN identifier and only that one. In an ux context of type as you search, this is impossible today without impacting the UI/UX or finding a workaround.
```
curquiza, line 18:
```suggestion
The user in question would like to be able to retrieve specifically the document containing the unique ISBN identifier and only that one. In a UX context of type as you search, this is impossible today without impacting the UI/UX or finding a workaround.
```

text/0043-phrase-query.md
```diff
@@ -0,0 +1,161 @@
...
+
+- It could have had a specific search field only dedicated to this ISBN field in order to make it a filter and using the `=` operator. Not great for the ux.
```
curquiza, line 20:
```suggestion
- It could have had a specific search field only dedicated to this ISBN field in order to make it a filter and using the `=` operator. Not great for the UX.
```

text/0043-phrase-query.md
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+}
+```
+
+### IV. Explanation
+
+To use the Phrase Query syntax, simply surround the contiguous search terms with the characters `"`.
+
+Let's say i want to search for a specific book with a title stricly containing `Plays and Playwrights 2002`.
```
curquiza, line 53:
```suggestion
Let's say I want to search for a specific book with a title strictly containing `Plays and Playwrights 2002`.
```

curquiza (changes requested):
I think this spec should clearly detail:
- what the `"` involves: it is described in the "Summary" but not clearly in "Explanation". Additionally to the examples, we should know the query word surrounded by `"` will not be impacted by: typo, n-gram, split word, prefix...  I mean an exhaustive and accurate list. You should check with Many we don't forget anything 😉
- the limitation of the query search: it is not case sensitive for example.

curquiza (approved):
🔥

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+
+The Phrase Query feature will easily solve this case but will also adapt to the needs of the user performing the search.
+
+### III. Additional Materials
+
+#### Algolia
+
+Algolia allows the use of Phrase Query syntax as long as the `advancedSyntax` parameter is set to true in the settings.
+
+As Algolia documentation said, a phrase query represents a specific sequence of terms that must be matched next to one another. A phrase query needs to be surrounded by double quotes ("). For example, the query "search engine" only returns a record if it contains “search engine” exactly in at least one attribute.
```
dichotommy, line 32:
```suggestion
As Algolia documentation said, a phrase query represents a specific sequence of terms that must be matched next to one another and in the given order. A phrase query needs to be surrounded by double quotes ("). For example, the query "search engine" only returns a record if it contains “search engine” exactly in at least one attribute.
```

dichotommy: Not 100% sure that this is true for Algolia, but it would make more sense for an "exact" search than just requiring the query terms to be next to one another.

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+    "query": "Plays and Playwrights 2002",
+    "limit": 20,
+    "offset": 0,
+    "processingTimeMs": 7
+}
+```
+
+To use the Phrase Query syntax, simply surround the contiguous search terms with the characters `"`.
+
+Using the Phrase Query syntax this way, with `q` equals to `"Plays and Playwrights 2002"`, will lead to have only one result because the title is exactly written like that.
```
dichotommy, line 95:
```suggestion
Using the Phrase Query syntax this way, with `q` equals to `"Plays and Playwrights 2002"`, will lead to have only one result because the title is written exactly like that.
```

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+
+> Note that it's case insensitive. So, if i search with `"plays and playwrights 2002"`, this will lead to the same result.
```
dichotommy, line 97:
```suggestion
> Note that it's case insensitive. So, if I search with `"plays and playwrights 2002"`, this will lead to the same result.
```

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+    "nbHits": 1,
+    "exhaustiveNbHits": false,
+    "query": "\"Plays and Playwrights 2002\"",
+    "limit": 20,
+    "offset": 0,
+    "processingTimeMs": 0
+}
+```
+
+So now let's say that i want to search title strictly mentioning `"African American"` but speaking about poetry. The Phrase query syntax can be used in conjunction with the basic syntax.
```
dichotommy, line 126:
```suggestion
So now let's say that I want to search for a title that strictly includes the phrase `"African American"` but speaking about poetry. The Phrase Query syntax can be used in conjunction with the basic syntax.
```

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+
+The query can be exprimed like that `"African American" poem`
```
dichotommy, line 128:
```suggestion
The query can be expressed like this: `"African American" poem`
```

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+            "title": "African-American Literature: A Brief Introduction and Anthology",
+            "author": "Al Young"
+        },
+        {
+            "title": "Early African American Classics (Barnes &amp; Noble Classics Series)",
+            "author": "Barnes &amp; Noble"
+        },
+```
+
+As you can see in the results, the presence or absence of one or more soft separators such as `-` between two words does not affect the query match of the phrase and the document.
```
dichotommy, line 159: What are the other "soft separators"?

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+
+### V. Impact on Documentation
+
+- Mention that new Phrase Query syntax in the documentation. Can it be on a dedicated page or on [the q search parameter](https://docs.meilisearch.com/reference/features/search_parameters.html#query-q)?
```
dichotommy, line 163: Yes, both! It should probably be mentioned on any page related to search, but also deserves a dedicated page.

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+- Specification PR: [#43](https://github.com/meilisearch/specifications/pull/43)
+- MeiliSearch Tracking-Issues:
+
+# Phrase Query
+
+## 1. Feature Description and Interaction
+
+### I. Summary
+
+MeiliSearch does not allow users a way to write a strict query in order to ask the engine to be more drastic in its selection of candidates for search results. The Phrase Query feature adds a simple syntax available to users to require the engine to select documents that will strictly match the search field value. That is, without typography, n-gram, wordsplit, prefix, synonym, and, proximity. In addition, the expression of a Phrase Query is case insensitive.
```
dichotommy, line 12:
```suggestion
MeiliSearch does not allow users a way to write a strict query in order to ask the engine to be more strict in its selection of candidates for search results. The Phrase Query feature adds a simple syntax available to users to require the engine to select documents that strictly match some phrase, indicated by quotation marks. That is, without typography, n-gram, wordsplit, prefix, synonym, and, proximity. In addition, the expression of a Phrase Query is case insensitive.
```

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+
+### II. Motivation
+
+This feature is driven by user needs. Indeed, let's take an example recently brought up in our community slack.
```
dichotommy, line 16:
```suggestion
This feature is driven by user needs. Indeed, let's take an example recently brought up in our community Slack.
```

dichotommy (approved):
Great feature specification. I think it shouldn't be a problem for the technical writing team to document this in time for v0.21.

Most of my questions are about the limitations on the exactness of the search. Which characters are treated as hard separators and which as soft? If I searched for example `"David Bowie"`, would I also find a document containing `David         Bowie`? What about a document containing `David Bowie–Space Oddity`?

text/0043-phrase-query.md
gmourier: I've updated the spec with the complete list of soft separators 🤓

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+
+This feature is driven by user needs. Indeed, let's take an example recently brought up in our community Slack.
+
+The user in question would like to be able to retrieve specifically the document containing the unique ISBN identifier and only that one. In a UX context of type as you search, this is impossible today without impacting the UI/UX or finding a workaround.
+
+- It could have had a specific search field only dedicated to this ISBN field in order to make it a filter and using the `=` operator. Not great for the UX.
+
+- It could also have kept a single search field and detected that an ISBN was entered in the search field using a regex to inject that filter at that point. Not great for the developper experience, moreover, what happens when a pattern cannot be specifically determined?
+
+The Phrase Query feature will easily solve this case but will also adapt to the needs of the user performing the search.
```
qdequele, line 24: When I read the user need here, It seems that the proposed solution is not the best way to do it. What if we asked the user to put his ISBN as `primaryKey` and to retrieve his document using the [get documents](https://docs.meilisearch.com/reference/api/documents.html#documents) route.

I think this feature goes further than that. It serves many other purposes as well. As you said above, it allows you to ignore the ranking rules for words between `"`. So it could allow, like google, to search `a green "t-shirt"` and to have at least documents with `t-shirt`. This word will not be removed by the ranking rule `Word`.

gmourier: Using the getDocuments route is a solution that will only work if the need is limited to searching for an ISBN value. If the search bar should also allow searching within the document attributes, we come back to the workaround to detect the ISBN pattern and make a filter of it or make a request on getDocuments route.

Totally agree with the second point. The Phrase Query allows for more than just the specific case cited.

text/0043-phrase-query.md
```diff
@@ -0,0 +1,172 @@
...
+
+The value between the `"` operators will be searched without:
+
+- typography
+- n-gram
+- wordsplit
+- prefix
+- synonym
+- proximity
+
```
ManyTheFish, line 107: Moreover, it ensures that all matching documents contain the words in the Phrase.

gmourier:
@dichotommy I've added a specific part about hard separators! :)

qdequele (approved):

ManyTheFish (approved):

ManyTheFish (approved):

PR #44, opened 2021-06-03
API Response Format

gmourier:
This specification is intended to write guidelines about the response formats of the different API endpoints.


gmourier:
Closed in favor of #45

PR #46, opened 2021-06-07
404/405 Http Error Codes

gmourier:
The purpose of this file is to specify the behavior of the API about the http code to return in case a resource does not exists or the http method used is not allowed on the resource.

### DIFF
--- a/text/0046-404-405-http-error-code.md
+++ b/text/0046-404-405-http-error-code.md
@@ -83,7 +83,7 @@ The message pattern is `Method {{HTTP_VERB}} is not allowed. See {{API REFERENCE
 
 ### V. Impact on Documentation
 
-- Add documentation for `method_not_allowed` and `unrecognized_request_url` errors.
+- Add `method_not_allowed` and `unrecognized_request_url` errors on the [errors documentation page](https://docs.meilisearch.com/errors/).
 
 ### VI. Impact on SDKs
 N/A
--- a/text/0046-404-405-http-error-code.md
+++ b/text/0046-404-405-http-error-code.md
@@ -1,0 +1,95 @@
+- Title: 404/405 Http Error Codes
+- Start Date: 2021-06-07
+- Specification PR: [#46](https://github.com/meilisearch/specifications/pulls/46)
+- MeiliSearch Tracking-issues:
+
+
+# 404/405 Http Error Codes
+
+## 1. Feature Description and Interaction
+
+### I. Summary
+
+The purpose of this file is to specify the behavior of the API about the http code to return in case a resource does not exists or the http method used is not allowed on the resource.
+
+This specification will be merged later into the more global API format specification that is currently being written.
+
+### II. Motivation
+
+A great API must be convenient to use. The development of a new endpoint, its improvement for developers must be done on a solid base to facilitate their work and focus on the intrinsic value of the endpoint. For users, it is also much easier and faster to use an API when endpoints share a common structure through conventions.
+
+### III. Additional Materials
+N/A
+
+### IV. Explanation
+
+### 404 Code
+
+The API must return a `404` error code when the resource does not exist or when the url used does not allow to reach a resource.
+
+E.g. Asking for an inexistent ressource.
+
+**GET `/indexes/:index_uid/documents/2`**
+
+In that case, the response body contain a `document_not_found` errorCode.
+
+E.g.
+
+```
+{
+    "message": "Document with id 2 not found",
+    "errorCode": "document_not_found",
+    "errorType": "invalid_request_error",
+    "errorLink": "https://docs.meilisearch.com/errors#document_not_found"
+}
+```
+
+E.g. Using a wrong url.
+
+**GET `/indexess/documents/2`**
+
+In that case, the response body should contain a `unrecognized_request_url` errorCode.
+
+```
+{
+    "message": "Unrecognized request URL (GET: /indexess/documents/2). See https://docs.meilisearch.com/reference/api",
+    "errorCode": "unrecognized_request_url",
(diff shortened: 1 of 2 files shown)


text/0046-404-405-http-error-code.md
```diff
@@ -0,0 +1,121 @@
...
+- MeiliSearch Tracking-issues:
+
+
+# 404/405 Http Error Codes
+
+## 1. Feature Description and Interaction
+
+### I. Summary
+
+The purpose of this file is to specify the behavior of the API about the http code to return in case a resource does not exists or the http method used is not allowed on the resource.
```
curquiza, line 13:
```suggestion
The purpose of this file is to specify the behavior of the API about the HTTP code to return in case a resource does not exist or the HTTP method used is not allowed on the resource.
```

text/0046-404-405-http-error-code.md
```diff
@@ -0,0 +1,121 @@
...
+
+| 0.21 Behaviors proposal                             | Invalid url    | Valid url but object not found          | Method not allowed |
+|-----------------------------------------------------|----------------|-----------------------------------------|--------------------|
+| 404 HTTP response status code                       |       x        |                     x                   |                    |
+| 405 HTTP response status code                       |                |                                         |          x         |
+| Presence of a well formed reponse format            |       x        |                     x                   |          x         |
+
+#### 404 Code
+
+The API must return a `404` error code when the resource does not exist or when the url used does not allow to reach a resource.
```
curquiza, line 53:
```suggestion
The API must return a `404` error code when the resource does not exist or when the URL used does not allow to reach a resource.
```

text/0046-404-405-http-error-code.md
```diff
@@ -0,0 +1,121 @@
...
+
+E.g. Asking for an inexistent ressource.
```
curquiza, line 55:
```suggestion
E.g. asking for an inexistent resource.
```

text/0046-404-405-http-error-code.md
```diff
@@ -0,0 +1,121 @@
...
+
+**GET `/indexes/:index_uid/documents/2`**
+
+In that case, the response body contain a `document_not_found` errorCode.
```
curquiza, line 59:
```suggestion
In that case, the response body contains a `document_not_found` errorCode.
```

text/0046-404-405-http-error-code.md
```diff
@@ -0,0 +1,121 @@
...
+```
+{
+    "message": "Document with id 2 not found",
+    "errorCode": "document_not_found",
+    "errorType": "invalid_request_error",
+    "errorLink": "https://docs.meilisearch.com/errors#document_not_found"
+}
+```
+
+E.g. Using a wrong url.
```
curquiza, line 72:
```suggestion
E.g. using a wrong URL.
```

text/0046-404-405-http-error-code.md
```diff
@@ -0,0 +1,121 @@
...
+}
+```
+
+The message pattern proposal is `Unrecognized request URL ({{HTTP_VERB}}: {{REQUEST URL}}). See https://docs.meilisearch.com/reference/api`
+
+#### 405 Code
+
+The API must return a `405` error code when the resource is reachable but the HTTP verb is not handled by the resource.
+
+E.g. Using a http verb that is not allowed
```
curquiza, line 93:
```suggestion
E.g. using a HTTP verb that is not allowed
```

text/0046-404-405-http-error-code.md
```diff
@@ -0,0 +1,121 @@
...
+
+E.g. Using a wrong url.
+
+**GET `/indexess/documents/2`**
+
+In that case, the response body should contain a `unrecognized_request_url` errorCode.
+
+```
+{
+    "message": "Unrecognized request URL (GET: /indexess/documents/2). See https://docs.meilisearch.com/reference/api",
```
curquiza, line 80: Not sure the message needs to refer to another link than `errorLink` in the message. The docs link should be explicit and re-link to the right resources.

text/0046-404-405-http-error-code.md
```diff
@@ -0,0 +1,121 @@
...
+
+E.g. Using a http verb that is not allowed
+
+**DELETE `/indexes/movies/search`**
+
+In that case, the response body should contain a `method_not_allowed` errorCode.
+
+```
+{
+    "message": "Method DELETE is not allowed. See https://docs.meilisearch.com/reference/api/search.html#search",
```
curquiza, line 101: Same, the `errorLink` should already redirect the user.

text/0046-404-405-http-error-code.md
gmourier: On this specific error, I can't think of a specific link! I was inspired by the error messages of Stripe API for this type of case which offers quite clear messages.

text/0046-404-405-http-error-code.md
gmourier: If the `errorLink` redirect to the search page (because it was what the user intended) it's ok...but we are loosing the link to the specific error page. If we redirect to the error page instead of the endpoint page, I think the user will lose the error context. That's why I added a contextualized link in the message part.

curquiza (dismissed):
Will not be done in v0.21.0, see: https://github.com/meilisearch/transplant/issues/201
But ok for next versions :)

irevoire (approved):

curquiza (approved):

Kerollmops (commented):
I am not sure I am enough informed on the subject and will probably ask to @MarinPostma about what are the default behaviors with actix-web, which are probably the expected ones, the ones that follow the HTTP spec.

PR #47, opened 2021-06-10
Reset Stop-Words and Synonyms Settings With A Null Value

gmourier:
# Summary

All Sub Settings `POST` endpoint accepts a null value in the http request body to reset the engine configuration to the default values. This is not the case for [Stop-words](https://docs.meilisearch.com/reference/api/stop_words.html#update-stop-words) and [Synonyms](https://docs.meilisearch.com/reference/api/synonyms.html#update-synonyms) settings endpoints.

# Motivation

This specification is written to make the above mentioned routes consistent with the others.

### DIFF
--- a/text/0047-reset-stop-words-synonyms-settings-with-null.md
+++ b/text/0047-reset-stop-words-synonyms-settings-with-null.md
@@ -22,8 +22,27 @@ N/A
 
 #### Null value request body
 
+E.g. Reset Stop-Words setting with a null value on [POST Stop-Words Endpoint](https://docs.meilisearch.com/reference/api/stop_words.html#update-stop-words)
+
+Given this stop-words configuration
+
+```
+[
+    "of",
+    "the"
+]
+```
+
+As a User I send
+
 ```
 null
+```
+
+To Expect
+
+```
+[]
 ```
 
 #### [POST Stop-Words Endpoint](https://docs.meilisearch.com/reference/api/stop_words.html#update-stop-words)
--- a/text/0047-reset-stop-words-synonyms-settings-with-null.md
+++ b/text/0047-reset-stop-words-synonyms-settings-with-null.md
@@ -61,7 +61,8 @@ The default value for `Synonyms` is an empty object.
 N/A
 
 ### VI. Impact on SDKs
-N/A
+
+The SDKs must allow the `null` value for every setting.
 
 ## 2. Technical Aspects
 
--- a/text/0047-reset-stop-words-synonyms-settings-with-null.md
+++ b/text/0047-reset-stop-words-synonyms-settings-with-null.md
@@ -30,13 +30,13 @@ null
 
 Sending a `null` value in the request body like the given request body above should reset the `Stop-Words` setting to the engine default values.
 
-Default value for `Stop-Words` is an empty object.
+The default value for `Stop-Words` is an empty array.
 
 #### [POST Synonyms Endpoint](https://docs.meilisearch.com/reference/api/synonyms.html#update-synonyms)
 
 Sending a `null` value in the request body like the given request body above should reset the `Synonyms` setting to the engine default values.
 
-Default value for `Synonyms` is an empty array.
+The default value for `Synonyms` is an empty object.
 
 ### V. Impact on Documentation
 N/A
--- a/text/0047-reset-stop-words-synonyms-settings-with-null.md
+++ b/text/0047-reset-stop-words-synonyms-settings-with-null.md
@@ -1,7 +1,7 @@
 - Title: Reset Stop-words and Synonyms settings with null value
 - Start Date: 2021-06-08
 - Specification PR: [#47](https://github.com/meilisearch/specifications/pull/47)
-- MeiliSearch Tracking-issues:
+- MeiliSearch Tracking-issues: [transplant/#217](https://github.com/meilisearch/transplant/issues/217)
 
 # Reset Stop-words and Synonyms settings with a null value.
 
--- a/text/0047-reset-stop-words-synonyms-settings-with-null.md
+++ b/text/0047-reset-stop-words-synonyms-settings-with-null.md
@@ -20,7 +20,7 @@ N/A
 
 ### IV. Explanation
 
-#### Null value request body
+#### Example
 
 E.g. Reset Stop-Words setting with a null value on [POST Stop-Words Endpoint](https://docs.meilisearch.com/reference/api/stop_words.html#update-stop-words)
 
--- a/text/0047-reset-stop-words-synonyms-settings-with-null.md
+++ b/text/0047-reset-stop-words-synonyms-settings-with-null.md
@@ -1,0 +1,49 @@
+- Title: Reset Stop-words and Synonyms settings with null value
+- Start Date: 2021-06-08
+- Specification PR: [#47](https://github.com/meilisearch/specifications/pull/47)
+- MeiliSearch Tracking-issues:
+
+# Reset Stop-words and Synonyms settings with a null value.
+
+## 1. Functional Specification
+
+### I. Summary
+
+All Sub Settings `POST` endpoint accepts a null value in the http request body to reset the engine configuration to the default values. This is not the case for [Stop-words](https://docs.meilisearch.com/reference/api/stop_words.html#update-stop-words) and [Synonyms](https://docs.meilisearch.com/reference/api/synonyms.html#update-synonyms) settings endpoints.
+
+### II. Motivation
+
+This specification is written to make the above mentioned routes consistent with the others.
+
+### III. Additional Materials
+N/A
+
+### IV. Explanation
+
+#### Null value request body
+
+```
+null
+```
+
+#### [POST Stop-Words Endpoint](https://docs.meilisearch.com/reference/api/stop_words.html#update-stop-words)
+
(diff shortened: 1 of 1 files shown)


text/0047-reset-stop-words-synonyms-settings-with-null.md
```diff
@@ -0,0 +1,49 @@
...
+N/A
+
+### IV. Explanation
+
+#### Null value request body
+
+```
+null
+```
+
```
qdequele, line 28: I don't understand this part. What do you want to explain here?

text/0047-reset-stop-words-synonyms-settings-with-null.md
```diff
@@ -0,0 +1,49 @@
...
+Sending a `null` value in the request body like the given request body above should reset the `Stop-Words` setting to the engine default values.
+
+Default value for `Stop-Words` is an empty object.
+
+#### [POST Synonyms Endpoint](https://docs.meilisearch.com/reference/api/synonyms.html#update-synonyms)
+
+Sending a `null` value in the request body like the given request body above should reset the `Synonyms` setting to the engine default values.
+
+Default value for `Synonyms` is an empty array.
```
qdequele, line 39:
```suggestion
The default value for `Synonyms` is an empty object.
```

text/0047-reset-stop-words-synonyms-settings-with-null.md
```diff
@@ -0,0 +1,49 @@
...
+
+```
+null
+```
+
+#### [POST Stop-Words Endpoint](https://docs.meilisearch.com/reference/api/stop_words.html#update-stop-words)
+
+Sending a `null` value in the request body like the given request body above should reset the `Stop-Words` setting to the engine default values.
+
+Default value for `Stop-Words` is an empty object.
```
qdequele, line 33:
```suggestion
The default value for `Stop-Words` is an empty array.
```

qdequele (changes requested):

text/0047-reset-stop-words-synonyms-settings-with-null.md
gmourier: I've added a better-detailed example to show what I mean by resetting with a `null` value.

ManyTheFish:
I don't like the idea, I don't think that `null` is considered equivalent to `[]` in everybody's mind. It could be interpreted as the absence of the field.

I would prefer that the user explicitly set the field to `[]` if he wants to reset it.

curquiza:
The other settings (for example `seachableAttributes` and `displayedAttribtutes`) already accept `null` as value to reset the settings, so to fill the settings to `[]`.
Even if this possibility is not documented, we know it is already used by some users: we saw it in the integration issues we had. So for me, this spec is a good idea:
- the users that don't have the habit to use `null` are still not impacted. The feature is not documented.
- the users who used `null` to reset can now do it consistently for synonyms and stop-words.

curquiza (approved):

gmourier:
Sorry @ManyTheFish, i forgot to answer you I got caught elsewhere. Your proposal goes in the direction of an API that can be fully documented. Personally, I'm not a fan of having hidden work-around that are not documented. The format API specification will solve this kind of case.

@curquiza , it's true for the 0.21 release, we cut in the fat to advance towards its release. However, your answer makes me wonder about two questions that we can deal with in future releases.

1) Integrations could accept a null value in the parameters of the update sub-setting method and transform it into [] or {}  or even call the DELETE method in this case.
2) If I understand correctly people don't call the reset method from integrations?  So do we need to keep these? Analytics on the feature level will tell us that to take a decision about that.

curquiza:
1. If this feature is a problem from your POV, moving it to the SDKs is just move the "problem" into 8 different repos (so duplicate the code) instead of one. Even in the SDKs, it will not be documented to avoid making the README and the docs heavier. So, from my POV, either we keep it everywhere, or not. 
2. The majority call the reset method. But some of them use the update method as well. Allowing the null value allow the user to call the general update route by updating and resetting setting at the same time. If we remove the null value to reset, it will remove this possibility. That's why I thought it's a nice feature to have for people noticing it because it's intuitive for them, without disturbing people that classically use the reset route.

gmourier:
Agreed on the fact that the first proposal will shift the effort to the X integration libs. Probably not a good idea. I don't really see a problem with having this null value in the POST even though it's something that seems unusual and might surprise some developers as well. As I said earlier, making some deviations to keep a good DX is sometimes necessary.

We need factual stats on the use of these routes to talk about this again. The fact is that we have two routes that accomplish the same thing. Is this a problem at the moment? You tell me no from the user feedback you get! 

It's not an immediate priority anyway. Let's move forward on this specification as such :)

curquiza:
Even if your decision is made, I want to share this smart way of usage 😊
Some users (sorry I cannot find where their feedback was, since it was several months ago) use the `null` value when calling the general update route (`POST /indexes/:uid/settings` ) in order to update and reset setting at the same time!

Example:
Calling `POST /indexes/:uid/settings` with
```json
{
    "searchableAttributes": null,
    "displayedAttributes": [
        "title"
    ]
}
```
will reset the searchable attributes and will update the displayed attributes at the same HTTP call.

If we remove the possibility to use the `null` value to reset, it will remove this possibility.

ManyTheFish:
> Even if your decision is made, I want to share this smart way of usage 😊 It might be intersting.
> Some users (sorry I cannot find where their feedback was, since it was several months ago) use the `null` value when calling the general update route (`POST /indexes/:uid/settings` ) in order to update and reset setting at the same time!
> 
> Example:
> Calling `POST /indexes/:uid/settings` with
> 
> ```json
> {
>     "searchableAttributes": null,
>     "displayedAttributes": [
>         "title"
>     ]
> }
> ```
> 
> will reset the searchable attributes and will update the displayed attributes at the same HTTP call.
> 
> If we remove the possibility to use the `null` value to reset, it will remove this possibility.

What about:

Calling POST /indexes/:uid/settings with
```
{
    "searchableAttributes": [],
    "displayedAttributes": [
        "title"
    ]
}
```

This would be similar to your example.

curquiza:
> Calling POST /indexes/:uid/settings with
> 
> ```
> {
>     "searchableAttributes": [],
>     "displayedAttributes": [
>         "title"
>     ]
> }
> ```
> 
> This would be similar to your example.

From my POV, this means "I don't want any searchable attributes, and only `title` as displayed attributes". It's not a reset.

ManyTheFish:
My bad @curquiza, I miss understood the meaning of reset. 😅
You're right.

qdequele (approved):

text/0047-reset-stop-words-synonyms-settings-with-null.md
```diff
@@ -0,0 +1,68 @@
...
+- Specification PR: [#47](https://github.com/meilisearch/specifications/pull/47)
+- MeiliSearch Tracking-issues:
+
+# Reset Stop-words and Synonyms settings with a null value.
+
+## 1. Functional Specification
+
+### I. Summary
+
+All Sub Settings `POST` endpoint accepts a null value in the http request body to reset the engine configuration to the default values. This is not the case for [Stop-words](https://docs.meilisearch.com/reference/api/stop_words.html#update-stop-words) and [Synonyms](https://docs.meilisearch.com/reference/api/synonyms.html#update-synonyms) settings endpoints.
```
bidoubiwa, line 12: whats happens if you send null today?

curquiza: It resets the settings. Except for stop-words and synonyms

gmourier: If you send null today on the route you get an error about the request body format

text/0047-reset-stop-words-synonyms-settings-with-null.md
```diff
@@ -0,0 +1,68 @@
...
+
+Sending a `null` value in the request body like the given request body above should reset the `Synonyms` setting to the engine default values.
+
+The default value for `Synonyms` is an empty object.
+
+### V. Impact on Documentation
+N/A
+
+### VI. Impact on SDKs
+N/A
```
curquiza, line 64:
```suggestion
### VI. Impact on SDKs
The SDKs might allow the `null` value for every settings. Already implemented, but the tests are not done for `synonyms` and `stop-words`.
```

text/0047-reset-stop-words-synonyms-settings-with-null.md
```diff
@@ -0,0 +1,68 @@
+- Title: Reset Stop-words and Synonyms settings with null value
+- Start Date: 2021-06-08
+- Specification PR: [#47](https://github.com/meilisearch/specifications/pull/47)
+- MeiliSearch Tracking-issues:
```
curquiza, line 4:
```suggestion
- MeiliSearch Tracking-issues: [transplant/#217](https://github.com/meilisearch/transplant/issues/217)
```

curquiza (changes requested):

text/0047-reset-stop-words-synonyms-settings-with-null.md
curquiza: (it might be to precise for your spec, feel free to not accept the second part of my comment @gmourier)

bidoubiwa (approved):

PR #52, opened 2021-07-07
Update 0027-filter-and-facet-behavior.md

dichotommy:
Add missing triple-tick to end code block


gmourier:
Thanks, @dichotommy! I'm merging that right away!

PR #58, opened 2021-07-26
Patch 0.21 OAS

gmourier:


open-api.yaml
```diff
@@ -519,35 +519,75 @@ components:
...
+        type: string
+        example: back to the future
+        default: '""'
     attributesToRetrieve:
       name: attributesToRetrieve
       in: query
-      description: |
-        Comma-separated list of attributes whose fields will be present in the returned documents.
-        Defaults to to the [displayedAttributes list](/reference/features/settings.html#displayed-attributes) which contains by default all attributes found in the documents.
+      description: 'Comma-separated list of attributes whose fields will be present in the returned documents. Defaults to to the [displayedAttributes list](/reference/features/settings.html#displayed-attributes) which contains by default all attributes found in the documents.'
```
bidoubiwa, line 533: This definition is only in the context of a `GET /search` ? I'm not I saw the method attached to it.


Same for all array parameters

open-api.yaml
```diff
@@ -1215,40 +1232,47 @@ paths:
...
                 q:
                   type: string
-                  description: Query string _(mandatory)_.
+                  description: Query string.
+                  default: '""'
+                  example: '"Back to the future"'
                 attributesToRetrieve:
                   type: array
-                  description: Attributes to display in the returned documents.
+                  description: 'Array of attributes whose fields will be present in the returned documents. Defaults to the [displayedAttributes list](/reference/features/settings.html#displayed-attributes) which contains by default all attributes found in the documents.'
```
bidoubiwa, line 1240: I see this is written in the context of `post:` whereas the previous `attributesToRetrieve` was not in the context of `get:`

Same for all array parameters

open-api.yaml
```diff
@@ -1263,16 +1287,18 @@ paths:
                   q: Harry
                   offset: 0
                   limit: 20
-                  filter: '("genres:Horror" OR "genres:Mystery") AND "director:Jordan Peele"'
+                  filter: (genres = Horror AND genres = Mystery) OR release_date > 523242000
                   facetsDistribution:
-                    - string
+                    - genres
+                    - author
```
bidoubiwa, line 1293: Should the examples not be formatted in JSON?

`'["genres", "author"`]'`

Not sure! Because this is also confusing as it looks like as a string more than a JSON

open-api.yaml
gmourier: Yes! You can test it over `get`and `post` `/search` routes. GET does not handle array syntax and POST does not handle string syntax separated by `,`

open-api.yaml
gmourier: I distinguised parameters for POST and GET since they doesn't share the same syntax.

open-api.yaml
gmourier: This is automatically generated by stoplight as the open api specification. It seems that it uses a different syntax for a parameter example and a body request example.

open-api.yaml
bidoubiwa: in the yaml it is not clear that this is for the `GET` method. Unless I didn't see it

gmourier: Query parameters are declared as components so they can be reused. They are linked using #ref to the routes using it!

PR #59, opened 2021-08-02
Geosearch

gmourier:
# I. Summary

The purpose of this specification is to add a first iteration of the **geo-search** feature to give geo-filtering and geo-sorting capabilities at search time. 

This first iteration allows to filter documents in a geo circle shape and to sort them around a geographical point.

## Summary Key Points

- Documents MUST have a `_geo` reserved object to be geo-searchable.
- Filter documents by a given geo radius using the built-in filter `_geoRadius({lat}, {lng}, {distance_in_meters})`. It is possible to cumulate several geo-search filters within the `filter` field.
- Sort documents in ascending/descending order around a geo point. e.g. `_geoPoint({lat}, {lng}):asc`.
- It is possible to filter and/or sort by geographical criteria of the user's choice.
- `_geo` must be set as a filterable attribute to use geo-filtering capabilities.
- `_geo` must be set as a sortable attribute to use geo-sorting capabilities.
- There is no `geo` ranking rule that can be manipulated by the user. This one is automatically integrated in the ranking rule `sort` by default and activated by using the `_geoPoint({lat}, {lng})` built-in sort rule.
- Using `_geoPoint({lat}, {lng})` in the `sort` parameter at search leads the engine to return a `_geoDistance` within the search results. This field represents the distance in meters of the document from the specified `_geoPoint`.
- Add an `invalid_geo_field` error to handle badly formed _geo field at document indexing.

# II. Motivation

According to our user feedback, the lack of a geo-search feature is mentioned as one of the biggest deal-breakers for choosing MeiliSearch as a search engine. A search engine must offer this feature. Some use cases specifically require integrated geo-search capabilities. Moreover, a lot of direct competitors offer it. Users today must find workarounds like using `geohash` to be able to geo-search documents. We hope to better serve the needs of users by implementing this feature. It allows multiplying the use-cases to which MeiliSearch can respond.

# III. Future Considerations

- Handling array of geo points in the document object.
- Handling multiple geo formats for the `_geo` field. e.g. "{lat},{lng}", a geo-hash etc..
- Add built-in filter rules for polygon and bounding-box. e.g. `_geoPoly`, `_geoBox`.
---

### Discovery Issue

https://github.com/meilisearch/product/issues/42

### DIFF
--- a/text/0059-geo-search.md
+++ b/text/0059-geo-search.md
@@ -1,0 +1,24 @@
+- Title: Geo-search
+- Start Date: 2021-08-02
+- Specification PR: [#59](https://github.com/meilisearch/specifications/pull/59)
+- MeiliSearch Tracking-issues:
+
+# Geo-search
+
+## 1. Functional Specification
+
+### I. Summary
+
+The purpose of this specification is to add the **geo-search** first iteration feature to filter and sort the search results as an end-user. This first iteration allows to filter the documents in a geo circle shape and sort them around a geographical point.
+
+#### Summary Key points
+
+- Documents MUST have a `_geo` reserved object to be geo-searchable.
+- Filter documents by a given geo radius. It is possible to cumulate several geo-search filters within the `filter` field. `_geoRadius({lat}, {lng}, {distance_in_meters})`.
+- Sort documents in ascending/descending order around a geo point. e.g. `_geoPoint({lat}, {lng}):asc`.
+- It is possible to filter and/or sort by geographical criteria of the user's choice.
+- There is no `geo` ranking rule that can be manipulated by the user. This one is integrated in the ranking rule `sort` by default.
+
+### II. Motivation
+
+According to our user feedback, the lack of a geo-search feature is mentioned as one of the biggest deal-breakers for choosing MeiliSearch as a search engine. A search engine must offer this feature. Some use cases specifically require integrated geo-search capabilities. Moreover, a lot of direct competitors offer it. Users today must find workarounds like using `geohash` to be able to geo-search documents. We hope to better serve the needs of users by implementing this feature. It allows multiplying the use-cases to which MeiliSearch can respond.
--- a/text/0059-geo-search.md
+++ b/text/0059-geo-search.md
@@ -9,7 +9,7 @@
 
 ### I. Summary
 
-The purpose of this specification is to add the **geo-search** first iteration feature to filter and sort the search results as an end-user. This first iteration allows to filter the documents in a geo circle shape and sort them around a geographical point.
+The purpose of this specification is to add a first iteration of the **geo-search** feature to give geo-filtering and geo-sorting capabilities at search time.
 
 #### Summary Key points
 
@@ -21,4 +21,10 @@ The purpose of this specification is to add the **geo-search** first iteration f
 
 ### II. Motivation
 
-According to our user feedback, the lack of a geo-search feature is mentioned as one of the biggest deal-breakers for choosing MeiliSearch as a search engine. A search engine must offer this feature. Some use cases specifically require integrated geo-search capabilities. Moreover, a lot of direct competitors offer it. Users today must find workarounds like using `geohash` to be able to geo-search documents. We hope to better serve the needs of users by implementing this feature. It allows multiplying the use-cases to which MeiliSearch can respond.
+According to our user feedback, the lack of a geo-search feature is mentioned as one of the biggest deal-breakers for choosing MeiliSearch as a search engine. A search engine must offer this feature. Some use cases specifically require integrated geo-search capabilities. Moreover, a lot of direct competitors offer it. Users today must find workarounds like using geohash to be able to geo-search documents. We hope to better serve the needs of users by implementing this feature. It allows multiplying the use-cases to which MeiliSearch can respond.
+
+## 3. Future Possibilities
+
+- Handling array of geo points in the document object.
+- Handling multiple geo formats for the `_geo` field. e.g. "{lat},{lng}", a geo-hash etc..
+- Filter documents in polygon and bounding-box.
--- a/text/0059-geo-search.md
+++ b/text/0059-geo-search.md
@@ -114,8 +114,6 @@ csv format example
 ```
 
 > 🔴 Giving a bad formed `_geo` that do not conform to the format causes the `update` payload to fail. An `invalid_request_error` description is given in the `update` object.
->
-> The `_geo` field does not need to be referred in `filterableAttributes` and `sortableAttributes` by the developer.
 
 ---
 
--- a/text/0059-geo-search.md
+++ b/text/0059-geo-search.md
@@ -169,7 +169,7 @@ csv format example
 - Not required
 
 Following the [`sort` specification feature](https://github.com/meilisearch/specifications/pull/55):
-> The `_geo` field has to be set in `sortableAttributes` setting by the developer to activate geo fsorting capabilities at search.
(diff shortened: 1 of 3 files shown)


curquiza (changes requested):

text/0059-geo-search.md
```diff
@@ -0,0 +1,222 @@
...
+
+**`_geo` field definition**
+
+- Name: `_geo`
+- Type: Object
+- Format: `{lat:float, lng:float}`
+- Not required
+
+> 💡 if `_geo` is found in the document payload, `lat` and `lng` are required.
+> 💡 `lat` and `lng` must be of float value.
```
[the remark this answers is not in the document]
gmourier: Good catch @curqui. I think we'll have to create a new error. A bit like `missing_primary_key`which is displayed in the relative `update` object when a document could not be processed in the asynchronous task queue.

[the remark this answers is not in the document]
gmourier: The specification introduces an `invalid_geo_field` error code

curquiza (approved):
Can't wait to see this new feature in the search engine 🚀

alallema (commented):
So great!
This is off-topic and I don't know much about geo search but are there any plans to offer geo search in imperial system one day?

gmourier:
@alallema Indeed, we could propose that in the future, the question is open, it is possible that some users ask us that. At first, the work-around will be to do the conversion on the client-side to match an imperial value to a metric value.

Kerollmops (approved):
Looks good to me, there will be a lot of work to do! :)

text/0059-geo-search.md
```diff
@@ -0,0 +1,232 @@
...
+> 🔴 Giving a bad formed `_geo` that do not conform to the format causes the `update` payload to fail. A new `invalid_geo_field` error is given in the `update` object.
+
+##### `invalid_geo_field` error
+
+| field     | value                                                                                                                |
+|-----------|----------------------------------------------------------------------------------------------------------------------|
+| message   | The _geo field is invalid.                                                                                           |
+| errorCode | invalid_geo_field                                                                                                    |
+| errorType | invalid_request_error                                                                                                |
+| errorLink | *Link to the dedicated error page. This page should give further links to indicate the expected JSON and CSV format* |
```
guimachiavelli, line 126: These will eventually be changed to follow spec #61, correct?

text/0059-geo-search.md
```diff
@@ -0,0 +1,232 @@
...
+
+#### POST Search `/indexes/{indexUid}/search`
+
+```
+{
+    "filter": ["brand = Ferrari", "_geoRadius(48.862725, 2.287592, 2000)"]
+}
+```
+
+> 🔴 Specifying parameters that do not conform to the `_geoRadius` signature causes the API to return an `invalid_filter` error. The error message should indicate how `_geoRadius` should be used. See `_geoRadius` built-in filter rule definition part.
```
guimachiavelli, line 157: This means that if a user does not specify a distance (e.g. `_geoRadius(0.0, 0.0)`) they will get the `invalid_filter` error, correct?

text/0059-geo-search.md
```diff
@@ -0,0 +1,232 @@
...
+- Introduce a `_geoPoint({lat}, {lng})` function parameter to `sort` documents around a central point.
+
+**`_geoPoint` built-in sort rule definition**
+
+- Name: `_geoPoint`
+- Signature: ({lat:float}:required, {lng:float}:required)
+- Not required
+
+Following the [`sort` specification feature](https://github.com/meilisearch/specifications/pull/55):
+> The `_geo` field has to be set in `sortableAttributes` setting by the developer to activate geo fsorting capabilities at search.
```
guimachiavelli, line 172:
```suggestion
> The `_geo` field has to be set in `sortableAttributes` setting by the developer to activate geo sorting capabilities at search.
```

text/0059-geo-search.md
```diff
@@ -0,0 +1,232 @@
...
+- Introduce a `_geoDistance` parameter to the search result `hit` object.
+
+**`_geoDistance` field definition**
+
+- Name: `_geoDistance`
+- Description: Return document distance when the end-user sorts document from a `_geoPoint` in meters.
+- Type: int
+- Not required
+
+> 💡 `_geoDistance` response field is only computed and shown when the end-user have sorted documents around a `_geoPoint`. So if the end-user filter documents using a `_geoRadius` built-in filter without sorting them around a `_geoPoint`, this field `_geoDistance` will not appear in the search response.
```
guimachiavelli, line 206:
```suggestion
> 💡 `_geoDistance` response field is only computed and shown when the end-user have sorted documents around a `_geoPoint`. So if the end-user filters documents using a `_geoRadius` built-in filter without sorting them around a `_geoPoint`, this field `_geoDistance` will not appear in the search response.
```

text/0059-geo-search.md
```diff
@@ -0,0 +1,232 @@
+- Title: Geo-search
+- Start Date: 2021-08-02
+- Specification PR: [#59](https://github.com/meilisearch/specifications/pull/59)
+- Discovery Issue: [#42](https://github.com/meilisearch/product/issues/42)
+- MeiliSearch Tracking-issues:
+
+# Geo-search
```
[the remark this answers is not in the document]
gmourier: Thank you @gui for your feedback on this topic. I will change that here too.

text/0059-geo-search.md
gmourier: If the error format spec is accepted and approved first, I would take the time to update it for in-progress/ready-for-review PR specifications however spec files could also be modified on the release PR that will prepare and summarize all implemented specs.

But to answer you yes it would be changed. 

I wouldn't update all the old specs that are already implemented. The specifications are epic mainly used by us to define what should be done in a specific iteration in an open way.

text/0059-geo-search.md
gmourier: Yes! 

It could be very interesting to have dedicated errors for each of these built-in filtering/sorting rules in the future!

text/0059-geo-search.md
```diff
@@ -0,0 +1,232 @@
...
+## 1. Functional Specification
+
+### I. Summary
+
+The purpose of this specification is to add a first iteration of the **geo-search** feature to give geo-filtering and geo-sorting capabilities at search time.
+
+#### Summary Key points
+
+- Documents MUST have a `_geo` reserved object to be geo-searchable.
+- Filter documents by a given geo radius using the built-in filter `_geoRadius({lat}, {lng}, {distance_in_meters})`. It is possible to cumulate several geo-search filters within the `filter` field.
```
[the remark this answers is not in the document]
gmourier: "Is there precedence for this in other parts of the codebase?" 

We have it for the virtual fields generated in search response like `_formatted`.

However, we have added a reserved field `_geo` within the documents to indicate the geographical coordinates of these. I naturally want to keep `_` as a prefix to the names of these new rules `_geoRadius` and `_geoPoint` to mark the fact that they are special expression elements related.

If this first iteration hits the mark. We should be able to add built-in filters rules for polygons and bounding box, it will be the moment to review the naming if we don't do it during our testing phases of this first iteration. I want to test it like that at first and return to it if necessary.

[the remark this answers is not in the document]
guimachiavelli: ok, yeah, `_` having precedent makes it more solid.

Also, to add to my previous comment: python also follows the convention of `_` == private method/var/etc.

Not that I think unofficial JS/python conventions should inform our design decisions too much.

PR #60, opened 2021-08-13
Refashion Update APIs

gmourier:
# Summary

The term `update` is not the best choice as it can be confused with `document` or `setting` updates. We have chosen to replace this term with `task`, which is more generic and better reflects the meaning of this API resource.

As an additional change, we have reworked the format of an update to make it more in line with our expectations of an API that is supposed to be easily understandable and developer-oriented.

The changes we will make to the response format of the `task` object lists will make it easier to add more functionality in a future iteration. See the Future Possibilities section for a brief overview.

Two new API endpoints are added. Although quite simple, they allow consulting the list of tasks or a specific task without being forced to know the related index.

## Summary Key Points

- The `update` resource is renamed `task`. The names of existing API routes are also changed to reflect this change.
- Tasks are now also accessible as an independent resource of an index. `GET - /tasks`; `GET - /tasks/:taskUid`.
- The task `uid` is not incremented by index anymore. The sequence is generated globally.
- A `task_not_found` error is introduced.
-  The format of the `task` object is updated.
    - `updateId` becomes `uid`.
    - Attributes of an error appearing in a `failed` `task` are now contained in a dedicated `error` object.
    - `type` is no longer an object. It now becomes a string containing the values of its `name` field previously defined in the `type` object.
    - The possible values for the `type` field are reworked to be more clear and consistent with our naming rules.
    - A `details` object is added to contain specific information related to a `task` payload that was previously displayed in the `type` nested object. Previous `number` key is renamed `numberOfDocuments`.
    - An `indexUid` field is added to give information about the related index on which the task is performed.
    - `duration` format has been updated to express an `ISO 8601` duration.
    - `processed` status changes to `succeeded`.
    - `startedProcessingAt` is updated to `startedAt`.
    - `processedAt` is updated to `finishedAt`.
- `202 Accepted` requests previously returning an `updateId` are now returning a summarized `task` object.
- `MEILI_MAX_UDB_SIZE` env var is updated `MEILI_MAX_TASK_DB_SIZE`.
- `--max-udb-size` cli option is updated to `--max-task-db-size`.
- `task` object lists are now returned under a `results` array.
- Each operation on an index (creation, update, deletion) is now asynchronous and represented by a `task`.

# Motivation

The main motivation is to stabilize the current `update` resource to a version that conforms to our API convention and thus allows future evolutions on a more solid base. We would like to modify the name `update`, its format is also to be reviewed because some attributes are not immediately clear, either in the possible values or in the chosen names.

### DIFF
--- a/open-api.yaml
+++ b/open-api.yaml
@@ -27,6 +27,7 @@ components:
   schemas:
     timestamp:
       type: string
+      description: An `ISO-8601` format for date/time/duration.
       example: '2019-11-20T09:40:33.711324Z'
     index:
       type: object
@@ -253,75 +254,86 @@ components:
         - exhaustiveNbHits
         - processingTimeMs
         - query
-    update:
+    task:
       type: object
       description: |
         MeiliSearch is an asynchronous API. It means that the API does not behave as you would typically expect when handling the request's responses.
 
         Some operations are put in a queue and will be executed in turn (asynchronously). In this case, the server response contains the identifier to track the execution of the operation.
 
-        [Learn More](https://docs.meilisearch.com/learn/advanced/asynchronous_updates.html).
+        [Learn More](https://docs.meilisearch.com/learn/advanced/asynchronous_tasks.html).
       additionalProperties: false
       examples:
-        - status: processed
-          updateId: 0
-          type:
-            name: DocumentAddition
-            number: 10
-          enqueuedAt: '2019-11-20T09:40:33.711324Z'
-          processedAt: '2019-11-20T09:40:33.711324Z'
-          duration: 5
+        - uid: 0
+          indexUid: movies
+          status: succeeded
+          type: documentsAddition
+          details:
+            numberOfDocuments: 10
+          duration: PT5S
+          enqueuedAt: '2021-01-01T09:39:00.000000Z'
+          startedAt: '2021-01-01T09:40:00.000000Z'
+          finishedAt: '2021-01-01T09:40:05.000000Z'
       properties:
+        uid:
+          type: integer
+          description: The unique sequential identifier of the task
+        indexUid:
+          type: string
+          description: The unique identifier of the index where this task is operated
         status:
           type: string
-          description: The status of the operation.
+          description: The status of the task
           enum:
             - enqueued
             - processing
-            - processed
+            - succeeded
             - failed
-          example: processed
-        updateId:
-          type: integer
-          description: The id of the update.
+          example: succeeded
         type:
+          type: string
+          description: The type of the task
+          enum:
+            - documentsAdditions
+            - documentsPartial
+            - documentsDeletion
+            - settingsUpdate
+            - clearAll
+        details:
           type: object
-          additionalProperties: false
-          required:
-            - name
-          description: Type object for the update task
+          description: Details information of the task payload.
           properties:
-            name:
-              type: string
-              description: Name of operation
-              example: DocumentAddition
-              enum:
-                - DocumentAddition
-                - Settings
-                - DocumentsDeletion
-                - ClearAll
-                - DocumentsPartial
-            number:
+            numberOfDocuments:
               type: integer
-              description: Number of documents impacted when the information can be given.
+              description: Number of documents impacted when the information can be given for the task.
             settings:
               $ref: '#/components/schemas/settings'
-        enqueuedAt:
-          $ref: '#/components/schemas/timestamp'
-        processedAt:
-          $ref: '#/components/schemas/timestamp'
-        duration:
-          type: number
-          description: The number of seconds taken to complete the operation.
         error:
+          description: Error object containing error details and context when a task has a failed status.
+          $ref: '#/components/schemas/error'
+        duration:
           type: string
-          description: |
-            A string describing [the error that occurred](https://docs.meilisearch.com/errors).
+          description: Total elasped time the engine was in processing state expressed as a `ISO-8601` duration format. Default is set to `null`
+          nullable: true
+        enqueuedAt:
+          description: Represent the date and time as a `ISO-8601` format when the task has been enqueued
+          $ref: '#/components/schemas/timestamp'
+        startedAt:
+          $ref: '#/components/schemas/timestamp'
+          description: Represent the date and time as a `ISO-8601` format when the task has been dequeued and started to be processed. Default is set to `null`
+          nullable: true
+        finishedAt:
+          $ref: '#/components/schemas/timestamp'
+          description: Represent the date and time as a `ISO-8601` format when the task has failed or succeeded. Default is set to `null`
+          nullable: true
       required:
+        - uid
+        - indexUid
         - status
-        - updateId
         - type
         - enqueuedAt
+        - startedAt
+        - finishedAt
     synonyms:
       type: object
       description: List of associated words treated similarly. A word associated to an array of word as synonyms.
@@ -520,10 +532,10 @@ components:
       required: true
       schema:
         type: any
-    updateId:
-      name: updateId
+    taskUid:
+      name: taskUid
       in: path
-      description: The update unique identifier
+      description: The task unique identifier
       required: true
       schema:
         type: number
@@ -652,16 +664,32 @@ components:
           schema:
             type: object
             properties:
-              updateId:
+              uid:
                 type: number
-                description: 'This `updateId` allows you to [track the current update](https://docs.meilisearch.com/reference/api/updates.html).'
+                description: 'This `uid` allows you to [track the current task](https://docs.meilisearch.com/reference/api/tasks.html).'
+              indexUid:
+                type: string
+                description: The unique identifier of the index where this task is operated
+              status:
+                type: string
+                description: The status of the task
+                example: enqueued
+              enqueuedAt:
+                $ref: '#/components/schemas/timestamp'
+                description: Represent the date and time as `ISO-8601` format when the task has been enqueued
             additionalProperties: false
             required:
-              - updateId
+              - uid
+              - indexUid
+              - status
+              - enqueuedAt
           examples:
             Example:
               value:
-                updateId: 1
+                uid: 0
+                indexUid: movies
+                status: enqueued
+                enqueuedAt: '2021-07-19T14:31:16.920473Z'
     '204':
       description: No Content
     '401':
@@ -720,9 +748,9 @@ tags:
       * A POST route: this is the preferred route when using API authentication, as it allows [preflight request](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) caching and better performances.
       * A GET route: the usage of this route is discouraged, unless you have good reason to do otherwise (specific caching abilities for example).
       Other than the differences mentioned above, the two routes are strictly equivalent.
-  - name: Updates
+  - name: Tasks
     description: |
-      The `updates` route gives information about the progress of the [asynchronous processes](https://docs.meilisearch.com/learn/advanced/asynchronous_updates.html).
+      The `tasks` route gives information about the progress of the [asynchronous tasks](https://docs.meilisearch.com/learn/advanced/asynchronous_tasks.html).
   - name: Keys
     description: |
       Each instance of MeiliSearch has three keys: a master, a private, and a public. Each key has a given set of permissions on the API routes.
@@ -2170,14 +2198,14 @@ paths:
           description: Not Found
     parameters:
       - $ref: '#/components/parameters/indexUid'
-  '/indexes/{indexUid}/updates':
+  '/indexes/{indexUid}/tasks':
     get:
-      operationId: indexes.updates.list
-      summary: Get all update status
+      operationId: indexes.tasks.list
+      summary: Get all tasks of an index
       description: |
-        Get the status of all [updates](https://docs.meilisearch.com/learn/advanced/asynchronous_updates.html) in a given [index](https://docs.meilisearch.com/learn/core_concepts/indexes.html).
+        Get all [tasks](https://docs.meilisearch.com/learn/advanced/asynchronous_tasks.html) of an [index](https://docs.meilisearch.com/learn/core_concepts/indexes.html).
       tags:
-        - Updates
+        - Tasks
       security:
         - apiKey: []
       responses:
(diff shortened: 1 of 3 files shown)


text/0060-refashion-updates-apis.md
```diff
@@ -0,0 +1,416 @@
...
+| status  | string  | Status of the task. Possible values are `enqueued`, `processing`, `succeeded`, `failed`                                |
+| type    | string  | Type of the task. Possible values are `documentsAddition`, `documentsPartial`, `documentsDeletion`, `settingsUpdate`, `clearAll` |
+| details | object |  Details information of the task payload. See `details` object definition by `type` part. |
+| error | object | Error object containing error details and context when a task has a `failed` status. See https://github.com/meilisearch/specifications/pull/61|
+| duration | string | Total elasped time the engine was in processing state expressed as a `ISO-8601` duration format. Default is set to `null`.  |
+| enqueuedAt | string | Represent the date and time as `ISO-8601` format when the task has been enqueued |
+| startedAt | string | Represent the date and time as `ISO-8601` format when the task has been dequeued and started to be processed. Default is set to `null`|
+| finishedAt | string | Represent the date and time as `ISO-8601` format when the task has `failed` or `succeeded`. Default is set to `null` |
+
+> 💡 The order of the fields must be returned in this order.
```
guimachiavelli, line 66: Don't oppose this in any way, but wanted to know why? 

I thought that, generally speaking, many/most languages discourage programmers from ever relying on the order of object fields. Or at least super high-level languages (e.g. JS, Ruby, PHP).

text/0060-refashion-updates-apis.md
```diff
@@ -0,0 +1,416 @@
...
+
+| field   | type    | description                     |
+|---------|---------|---------------------------------|
+| uid      | integer | Unique sequential identifier           |
+| indexUid | string | Unique index identifier |
+| status  | string  | Status of the task. Possible values are `enqueued`, `processing`, `succeeded`, `failed`                                |
+| type    | string  | Type of the task. Possible values are `documentsAddition`, `documentsPartial`, `documentsDeletion`, `settingsUpdate`, `clearAll` |
+| details | object |  Details information of the task payload. See `details` object definition by `type` part. |
+| error | object | Error object containing error details and context when a task has a `failed` status. See https://github.com/meilisearch/specifications/pull/61|
+| duration | string | Total elasped time the engine was in processing state expressed as a `ISO-8601` duration format. Default is set to `null`.  |
```
guimachiavelli, line 61: sorry, couldn't help it.
```suggestion
 | duration | string | Total elapsed time the engine was in processing state expressed as a `ISO-8601` duration format. Default is set to `null`.  |
```

text/0060-refashion-updates-apis.md
```diff
@@ -0,0 +1,416 @@
...
+
+---
+
+#### 5. APIs endpoints
+
+**Get all tasks** | `GET` - `/tasks`
+
+##### Goals
+
+Allows users to list tasks globally regardless of the indexes involved. Particularly useful to visualize all the tasks.
```
guimachiavelli, line 238: So, this gives a full list of all task/update operations ever requested across an instance's whole history, right? Do we have any data on how big this can actually get? Could it become a problem if an instance has been running for a **very** long time and the resulting `task` response is huge?

guimachiavelli: Additionally: is the record of tasks/updates maintained if an instance is rebooted? Like, if the instance (or the machine in which it is running) crashes, will I be able to access previous updates/tasks?

text/0060-refashion-updates-apis.md
gmourier: Indeed, it is true that it is not recommended to rely on the order of the information of a JSON object programmatically to access the information. Access by keys is recommended and allows to be completely agnostic about the order of the fields in JSON format.

However, despite the JSON specification which does not guarantee a consistent order of fields, we have control over this information in the responses we generate on the server-side, it's considered a good practice to organize objects in a consistent way when possible. The goal here is also to make life easy for a human reading the API responses. For example, it's more pleasant to find the IDs at the top of the response when it's read by someone.

The idea is to write an API for machines but also for humans when possible.

text/0060-refashion-updates-apis.md
gmourier: Thanks @guimachiavelli 😇

By the way, what do you think of this nomenclature for the `duration` field? e.g. `PT3M6S` to express a duration of 3 minutes and 6 seconds.

text/0060-refashion-updates-apis.md
gmourier: "So, this gives a full list of all task/update operations ever requested across an instance's whole history, right?" Absolutely. 

This can indeed represent a large volume of data in a specific case. Having several indexes in which entries are made regularly. On the product side, we planned to add pagination (It should probably be keyset pagination based on consistency and performance) and a default limit very soon.

So yes it could become a problem on the specific case mentioned above while we can't paginate it yet. The advantage is that the user will always have access to the list of tasks by index, it is not removed.

I believe that if the crash occurs during a transaction, the task is lost.  @MarinPostma Is it true?
If the machine reboots to a stable state, the previous tasks are stored and therefore accessible.

irevoire (approved):
I’m ok with everything

text/0060-refashion-updates-apis.md
guimachiavelli: Hmmm…do you mean using `PT[number]M[number]S` instead of `hh:mm:ss` (which is what I understood ISO-8601 stood for)? I personally find it a bit hard to read. Like, I need a reference to understand what it stands for since it's not a standard (or not a standard I have previously encountered.) Or perhaps I could figure it out on my own, but it takes a bit more brain processing time.

To give my full (very personal) preference, I usually prefer duration in APIs to be given in seconds, as it makes it easier to calculate things. I'm a fan of UNIX epoch for the same reason. This is probably because of my background in front-end, so it is often the case you need to do stuff like e.g. compensating for the duration of a task, updating animations, etc based on time elapsed.

text/0060-refashion-updates-apis.md
guimachiavelli: Makes sense, thanks for the explanation =)

text/0060-refashion-updates-apis.md
guimachiavelli: My follow-up question was going to be about pagination, so thanks for clearing that up!

As for my second question, my apologies, I realised I actually already knew the reply: if an instance is interrupted during an update/task, the update/task will be marked as enqueued, as stated in the [async update guide](https://docs.meilisearch.com/learn/advanced/asynchronous_updates.html#terminate-meilisearch-while-a-task-is-being-processed).

text/0060-refashion-updates-apis.md
```diff
@@ -0,0 +1,426 @@
...
+        }
+    ]
+}
+```
+
+##### Requirements
+
+> 💡 `task` objects are contained in a `results` array.
+>
+> 💡 By default, objects are sorted by `desc` order on `uid` field. So the most recent tasks appear first.
```
mdubus, line 274: Sorry if this was answered elsewhere, but I'm not sure to understand correctly this sentence.
It appears to me that the two sentences `sorted by desc order on uid field` and `the most recent tasks appear first` are contradictory. 

By returning by `desc order on uid`, what if an index has a higher `uid` than the one we just updated ? 
For example, I have an index `pokemon` that I updated a lot in the past. Now, I'm working on an `movies` index that I just started. If I request the `tasks`, I will first receive the tasks of the `pokemon` index and not the most recent ones (the `movies` one), won't I ? 

Or maybe the `uid` aren't scoped to an index, meaning that the `uid` are unique and I can't have the same `uid` twice for my `pokemon` and `movies` index ?

Sorry if this was explained elsewhere, I'm doing my best to understand this 😄

gmourier: Hello @mdubus! `uid` is the identifier of a task. The identifiers will be unique and sequential in a global way. 

So a first task uid on a new index will start at `lastTaskUid + 1`. 

Thanks for bringing this up, it's not specified. I will clarify this immediately.

mdubus: Thanks a lot ! This is much clearer now ❤️

gmourier: If you are on the `tasks/` route you will have the most recent tasks first regardless of an index.

If you are on the `indexes/pokemon/tasks` route you will the most recent tasks first for the pokemon index.

open-api.yaml
```diff
@@ -2325,4 +2356,102 @@ paths:
...
+                        enqueuedAt: '2021-01-01T09:38:00.000000Z'
+                        startedAt: '2021-01-01T09:38:02.000000Z'
+                        finishedAt: '2021-01-01T09:38:07.000000Z'
+      operationId: tasks.list
+      parameters: []
+      security:
+        - apiKey: []
+  '/tasks/:taskUid':
+    get:
+      summary: Get a task
```
bidoubiwa, line 2412: Should we say `get a task status` ? As `get a task` means `get something to do`.

text/0060-refashion-updates-apis.md
```diff
@@ -0,0 +1,429 @@
...
+
+The changes we will make to the response format of the `task` object lists will make it easier to add more functionality in a future iteration. See the Future Possibilities section for a brief overview.
+
+Two new API endpoints are added. Although quite simple, they allow to consult the list of tasks or a specific task without being forced to know the related index.
+
+#### Summary Key Points
+
+- The `update` resource is renamed `task`. The names of existing API routes are also changed to reflect this change.
+- Tasks are now also accessible as an independent resource of an index. `GET - /tasks`; `GET - /tasks/:taskUid`
+- The task `uid` is not incremented by index anymore. The sequence is generated globally.
```
bidoubiwa, line 24: What where the motivation behind this choice? As they make chronology harder to see.

text/0060-refashion-updates-apis.md
```diff
@@ -0,0 +1,429 @@
...
+- A `task_not_found` error is introduced.
+- The format of the `task` object is updated.
+    - `updateId` becomes `uid`.
+    - Attributes of an error appearing in a `failed` `task` are now contained in a dedicated `error` object.
+    - `type` is no longer an object. It now becomes a string containing the values of its `name` field previously defined in the `type` object.
+    - The possible values for the `type` field are reworked to be more clear and consistent with our naming rules.
+    - A `details` object is added to contain specific information related to a `task` payload that was previously displayed in the `type` nested object. Previous `number` key is renamed `numberOfDocuments`.
```
bidoubiwa, line 31: Out of curiosity, Is it planned that  `numberOfDocuments` returns in enqueued/processingfailed status ?

bidoubiwa (commented):
Should we add a `limit` and `offset` to avoid that `GET /tasks` could potentially return an infinite number of tasks? Knowing that pagination with a cursor might be implemented later. But both are not incompatible together.

open-api.yaml
```diff
@@ -234,75 +235,86 @@ components:
...
-          description: The id of the update.
+          example: succeeded
         type:
+          type: string
+          description: The type of the task
+          enum:
+            - documentsAdditions
+            - documentsPartial
+            - documentsDeletion
+            - settingsUpdate
```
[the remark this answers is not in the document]
gmourier: Yes, this is something identified and noted on the product side! Please comment on the specification that kind of feedback. It's easier for us to manage on the product side afterward 🤓

open-api.yaml
```diff
@@ -625,16 +637,32 @@ components:
...
+              indexUid:
+                type: string
+                description: The unique identifier of the index where this task is operated
+              status:
+                type: string
+                description: The status of the task
+                example: enqueued
+              enqueuedAt:
+                $ref: '#/components/schemas/timestamp'
+                description: Represent the date and time as `ISO-8601` format when the task has been enqueued
```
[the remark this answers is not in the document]
gmourier: No, the API response is much smaller for the `202 Accepted` as mentioned in the spec!

text/0060-refashion-updates-apis.md
gmourier: This is only possible for tasks completed today because of the architecture of the update store. This is something that could be great in the future!

open-api.yaml
```diff
@@ -234,75 +235,86 @@ components:
...
-              description: Number of documents impacted when the information can be given.
+              description: Number of documents impacted when the information can be given for the task.
             settings:
               $ref: '#/components/schemas/settings'
+        error:
+          description: Error object containing error details and context when a task has a failed status.
+          $ref: '#/components/schemas/error'
+        duration:
+          type: string
+          description: Total elasped time the engine was in processing state expressed as a `ISO-8601` duration format. Default is set to `null`
```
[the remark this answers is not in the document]
gmourier: Yes, with the notation `PT2M3S` for example to say 2 minutes and 3 seconds.

open-api.yaml
gmourier: The open API spec is also driven by what the documentation decides in terms of text or explanation, that's why I'm taking up the texts later. The most important thing is to focus on the structure changes more than the text

text/0060-refashion-updates-apis.md
gmourier: Why would the chronology be more difficult to understand in this case? 

The uids are always sorted from the most recent to the oldest. It will always be possible to go to a specific index and have the tasks in a chronological way for a specific index. 

The goal was to provide a global endpoint for teams with several indexes to manage. If we had kept a sequence by index we could have done it with difficulty. 

This should also be a plus when implementing high availability because the sequence of identifiers is globalized in the whole system.

gmourier:
> Should we add a `limit` and `offset` to avoid that `GET /tasks` could potentially return an infinite number of tasks? Knowing that pagination with a cursor might be implemented later. But both are not incompatible together.

We have started to work on the pagination, a specification should arrive very soon. 🤓

`limit' is a useful parameter in this case, but it's less obvious for 'offset'. Both `offset` and `limit` are useful in the case of per-page pagination.

We think that a cursor pagination is very appropriate for updates because it allows to keep the performance throughout the browsing and especially the consistency of the results.

It will also be a plus if we decide to add the removal of enqueued or finished updates.

text/0060-refashion-updates-apis.md
bidoubiwa: > The goal was to provide a global endpoint for teams with several indexes to manage. If we had kept a sequence by index we could have done it with difficulty.

Oh I see, that's smart! Thanks :)

bidoubiwa:
### Type of Actions

For accessibility I copy my comment here:

Could there be also the sub types ofsettingsUpdate as their route exists.

The user might use different update sub routes one after the other and the task type might give clarity to which has been processed.

Ex:
`searchableAttributeUpdate`

---- 

### Duration

>  Bidoubiwa : Is duration showcasable with ISO-8601 ?

> Gmourier: Yes, with the notation PT2M3S for example to say 2 minutes and 3 seconds.

Is this often used to describe a specific amount of time? Seems a bit hard to read. It would require a parser or at least a knowledge of the syntax to understand how to read it. While DateTime is clear to read for whole dates (dd-mm-yy-h-m-s).
What was the motivation behind this decision? Was just the amount of seconds a possibility?

gmourier:
Sorry @bidoubiwa, I'm really late on this. 

The addition of specific types by type of modified settings will be done later if the need arises. This is noted as a future possibility on line 428.

---

The proposed format duration is a human-readable format at the API response level in addition to being easily parsable programmatically with a simple regex however I agree that standard libraries are not necessarily up to date on this and that's a shame.

I don't believe that returning a value representing seconds or milliseconds is a good thing for the duration field. Personally, if I get a duration value of `46384` in seconds or milliseconds I can't do the calculation in my head.

If there is a need to make it readable in a more usual way there are two solutions that SDKs or clients can implement. Code a regex or use a library to parse a value like `PT3M6S` to display it in the form `3 minutes and 6 seconds`. It is also possible to make the time difference between `startedAt` and `finishedAt`.

For very fast duration below one second, the iso 8601 standard foresees to use the syntax `0.5S` for example to express `500ms`.

On this particular subject, it would be nice if we took some time to look at what it is possible to do precisely at the level of our SDKs from several values of duration in 8601 formats that the engine could answer. We can fall back on something like `processingTimeMs` with a number of milliseconds or `processingTimeSeconds` with a number of seconds if it's restrictive to parse a duration in 8601 formats but we completely lose human readability with regard to the API response.

Looking very quickly at the code of the javascript client it is not indicated at any time that `duration` is a number of seconds. Someone wanting to use this field would not have any order of magnitude, making this field useless for a user and forcing him to read the documentation, whereas a better-named field or a standardized string would have allowed him to read this data regardless of the name of the field. https://github.com/meilisearch/meilisearch-js/blob/main/src/index.ts

---

Concerning the addition of limit/offset, we could propose it later too. We will focus first on cursor paging since this particular endpoint is a problem for users with a lot of updates (I've got feedback of nearly 2 million updates on some indexes), it is unusable for them. In the long run, cursor paging is beneficial for the users. SDKs and integrations will be able to integrate iterators to easily browse updates in this way.

In case the user wants the last update, he will just have to send `limit=1` since the tasks uid will be sorted by default from the most recent to the oldest. If one day there is a need to browse the task lists with a search with page numbers, we can introduce offset in addition. 

A proto-solution has been written for cursor paging, a specification will be written soon with all the information needed to implement it. https://github.com/meilisearch/product/issues/190#issuecomment-910502267

bidoubiwa:
Hey @gmourier ! 

Standard libraries of most languages do parser the ISO-8601 but only the onces represented as dates `YY-MM-DD:HH:mm:SS` and not the the duration representation. Which may or may not represent the popularity of the duration format  `PTxMxS`

> I don't believe that returning a value representing seconds or milliseconds is a good thing for the duration field. Personally, if I get a duration value of 46384 in seconds or milliseconds I can't do the calculation in my head.

I agree, I like the `00:03:04` format to represent `HH:mm:SS` as it offers a great human readable format and an international understandable format. This format is also very easy to parse, if not more, than the `ISO-8601` since the latter requires no more  than just a `split(":")`.

> On this particular subject, it would be nice if we took some time to look at what it is possible to do precisely at the level of our SDKs from several values of duration in 8601 formats that the engine could answer. 

This is a great question I do not have the right answer for. I would have said innocently seconds but if the user but we do introduce redundant information as if the user can parse iso-8061 into timestamps he make the difference of these Date's to know in seconds their duration. 
```
          enqueuedAt: '2021-01-01T09:39:00.000000Z'
          startedAt: '2021-01-01T09:40:00.000000Z'
          finishedAt: '2021-01-01T09:40:05.000000Z'
```

At the end of the day, all formats will do so if you want to stay with `iso-8601` for the representation of the duration it is good to me!

text/0060-refashion-updates-apis.md
```diff
@@ -0,0 +1,509 @@
...
+
+```json
+{
+    "uid": 1,
+    "indexUid": "movies",
+    "status": "enqueued",
+    "type": "indexUpdate",
+    "enqueuedAt": "2021-08-12T10:00:00.000000Z"
+}
+```
```
curquiza, line 477: Is this response the expected response for any asynchronous task (ex: add documents, add settings)? Sorry if it's already in the spec, I'm a little bit lost since they are a lot of changes, maybe it's visible in open API

gmourier: Yes, this is a summarized `task` response only for the `202 Accepted`. The goal is to give more details instead of just an identifier as a response.

curquiza (approved):
💪💪

PR #61, opened 2021-08-13
Error Format

gmourier:


curquiza (approved):

gmourier:
Status changed to `in progress` because it needs a detailed definition of all our errors.

gmourier:
I've fleshed out the specification by doing a complete error cleanup. Errors added by external specifications like ndjson, csv, geo search, and sort are also included. I have completed an overview of our known API errors to date which you can see here https://miro.com/app/board/o9J_l0i--Zo=/.

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,807 @@
...
+
+- The `:reason` is inferred when the message is generated. e.g. `Write Permission denied`
+
+---
+
+## index_not_accessible
+
+### Context
+
+This error occurs when an index can be accessed for various reasons.
```
guimachiavelli, line 605:
```suggestion
 This error occurs when an index cannot be accessed for various reasons.
```

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,807 @@
...
+
+---
+
+## dump_process_failed
+
+### Context
+
+This error occurs during the dump creation process. The dump creation was interrupted for various reasons.
+
+💡 Errors returned asynchronously in a `dump` object, does not include a definition of the http code.
```
guimachiavelli, line 675:
```suggestion
💡 Errors returned asynchronously in a `dump` object do not include a definition of the HTTP code.
```

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,807 @@
...
+```json
+{
+    "message": "The creation of the dump :dumpId failed due to :reason.",
+    "code": "dump_process_failed",
+    "type": "internal",
+    "link": "https://docs.meilisearch.com/errors#dump_process_failed"
+}
+```
+
+- The `:reason` is inferred when the message is generated.
```
guimachiavelli, line 688: What is an example of a possible `:reason`?

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,807 @@
...
+
+This error occurs when the user tries to add documents and the maximum size of the database reaches the limit. The user can correct this error by increasing the database size limit.
+
+### Error Definition
+
+HTTP code `500 Internal Server Error`
+
+```json
+{
+    "message": "The maximum size of :databaseSizeLimit for the database has been reached.",
```
guimachiavelli, line 704:
```suggestion
    "message": "Maximum database size of :databaseSizeLimit has been reached.",
```

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,807 @@
...
+
+This error occurs when the route is protected and the value of the `X-MEILI-API-KEY` header does not allow access to the resource.
+
+### Error Definition
+
+HTTP Code: `403 Forbidden`
+
+```json
+{
+    "message": "The provided API Key is invalid.",
```
guimachiavelli, line 792: FYI: @meilisearch/docs-team has been avoiding using `API Key` to avoid confusion between `master`, `private` and `public` keys. We have discussed using authentication key when any of the keys will do.

I'm not sure whether this applies here, so for the moment I'm just noting it for the moment.

text/0061-error-format-and-definitions.md
gmourier: For example, it can be a denied permission for writing to the file system. I'll add it to give more context. In general, known internal errors can occur as a result of system misconfiguration or limits being reached.

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,807 @@
...
+
+This error occurs at search time when there is a syntax error in the `filter` parameter or when an attribute expressed in the filter is not defined in the `filterableAttributes` list.
+
+### Error Definition
+
+HTTP Code: `400 Bad Request`
+
+```json
+{
+    "message": ":syntaxErrorHelper. Attribute :attribute is not filterable. Available filterable attributes are ..., ..., ...",
```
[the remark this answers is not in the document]
gmourier: "Can we have special cases when filterableAttributes is empty?". Absolutely, thanks @guimachiavelli. I completely forgot about this one.

open-api.yaml
```diff
@@ -960,9 +960,9 @@ paths:
                 Primary Key Already Present:
                   value:
                     message: A primary key is already present. It's impossible to update it
```
bidoubiwa, line 962: Missing dot `.` at the end of the sentence. Also i was thinking of the following wording:

`Primary Key's can not be updated when one is already defined in the index.`

Additionaly:
`Index should be re-created with correct primary key`

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,810 @@
...
+This specification also serves as a reference point for the complete list of API errors that the user may encounter.
+
+#### Summary Key Points
+
+- `error` is removed from the attributes name and `type`values.
+- An exhaustive list of API errors is defined. See Errors List part.
+
+### II. Motivation
+
+The main motivation is to stabilize the current `error` resource to a version that conforms to our API convention and thus allows future evolutions on a more solid base. This specification avoids adding unnecessary information in the error object's attribute names.
```
bidoubiwa, line 23: > that conforms to our API convention and thus allows future evolutions on a more solid base

Why is the mention of the word `error` something we want to remove? It seems a maybe useful  information for those checking errors using conditions from the response body instead of catching errors through HTTP code. 

The latter is of course the best practice but I was wondering why we decided to remove the word.

[Algolia](https://www.algolia.com/doc/rest-api/search/#error-handling) does not mention the word error

ElasticSearch does
exampl: 
```json
{
    "error": {
        "caused_by": {
            "reason": "json_parse_exception: Unexpected character (',' (code 44)): was expecting either valid name character (for unquoted name) or double-quote (for quoted) to start field name\n at [Source: org.elasticsearch.common.io.stream.InputStreamStreamInput@139163f; line: 3, column: 17]",
            "type": "json_parse_exception"
        },
        "reason": "failed to parse",
        "root_cause": [
            {
                "reason": "json_parse_exception: Unexpected character (',' (code 44)): was expecting either valid name character (for unquoted name) or double-quote (for quoted) to start field name\n at [Source: org.elasticsearch.common.io.stream.InputStreamStreamInput@139163f; line: 3, column: 17]",
                "type": "json_parse_exception"
            }
        ],
        "type": "mapper_parsing_exception"
    },
    "status": 400
}
```

Maybe `message` could be `error` instead, or wrap everything is an error key? 
```json
{
   "error": { "message": "woops", ... }
}
```

I did not find a consensus online on best practices so the way you propose it might be the best! I have no strong opinions on it :)

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,810 @@
...
+
+The second motivation is to describe in an exhaustive way all the errors that the user may encounter during his use of the API.
```
bidoubiwa, line 25:
```suggestion
The second motivation is to describe in an exhaustive way all the errors that the user may encounter during their use of the API.
```

text/0061-error-format-and-definitions.md
gmourier: I think there is no need to specify `errorCode` or `errorType` rather than `code` or `type` since the user understands that he is dealing with an error by reading the content of the response. The programs themselves should be based on the HTTP code. In any case, the context is clear so that there is no redundancy of `error` in the names.

Concerning the asynchronous errors that are reported directly in the task object, the error object will be displayed under an `error` key.

bidoubiwa: > Concerning the asynchronous errors that are reported directly in the task object, the error object will be displayed under an error key.

Should it be stated in this spec?

It would look like this? 

```json
{
  "status": "failed",
  "updateId": 1,
  "type": {
    "name": "DocumentsAddition",
  },
  "enqueuedAt": "2019-12-07T21:10:07.607581330Z",
  "duration": 0.000048524,
  "processedAt": "2019-12-07T21:10:20.511525620Z",
  "error": {
    "message": "You must have an authorization token",
    "code": "missing_authorization_header",
    "type": "authentication_error",
    "link": "https://docs.meilisearch.com/errors#missing_authorization_header"
  }
}
 ```


> Your last proposal is also something we can do! I've come across a few examples of APIs doing this because errors were returned with a 200 HTTP code response while some also return the status code in the error object with an equivalent HTTP code for the response. 

This might be linked indeed to how errors might pop in the API! I'm not sure this is something that might happen in MeiliSearch. 
It could also introduce the `warn` key, but it might be something that is never returned in API responses. At least i don't remember crossing one

File: open-api.yaml (hunk 7 above)
Comment (gmourier) [reply open-api.yaml:962]: The OAS is not yet up to date, the most updated messages are in the specification file 🤓

**Comment by gmourier** [comment]
A working group is in preparation to address the messages and return format.

File: text/0061-error-format-and-definitions.md [hunk 10]
```diff
@@ -0,0 +1,807 @@
...
+### Context
+
+This error occurs when the engine does not find an identifier to infer in the documents to be inserted during the first ingestion.
+
+### Error Definition
+
+```json
+{
+    "message": "",
+    "code": "missing_document_primary_key",
```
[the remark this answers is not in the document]
gmourier: This particular point will be addressed when we will deal with refashioning of the index and document resources. In the meantime, we will specify more precise messages for the user to guide him/her in the best way.

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,807 @@
...
+    "type": "invalid_request",
+    "link": "https://docs.meilisearch.com/errors#missing_document_primary_key"
+}
+```
+
+---
+
+## missing_document_id
```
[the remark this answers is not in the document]
gmourier: We will discuss this in more detail during the working group to clarify this. There is no fundamental difference between the primary key and the document_id. `document_primary_key` represents the name of the attribute that identifies the document while `document_id` represents the value for `document_primary_key`. In this case it is especially a notion of temporality which intervenes and which is confused.

text/0061-error-format-and-definitions.md
```diff
@@ -0,0 +1,901 @@
...
+    "link": "https://docs.meilisearch.com/errors#missing_document_id"
+}
+```
+
+- The `:primaryKey` is inferred when the message is generated. This is the value of the primaryKey attribute of the related index object.
+- The `:documentRepresentation` is inferred when the message is generated.
+
+---
+
+## document_fields_limit_reached
```
curquiza, line 206: Why this one is not an `internal`, like `database_size_limit_reached`?

gmourier: It's on the user side because the user can cause this error by sending a document containing `65536` fields in a payload 🤓

bidoubiwa (approved):

PR #63, opened 2021-08-16
Add definitions of the missing behaviors for the sort specification

gmourier:
- In the case where an attribute is specified as a sort criterion at search time and if this attribute is of a different type between several documents, the numeric type will always be favored first by the engine. This means that documents with numeric values for this attribute will be sorted before those with string values. This can lead to awkward sorting behavior, so the user should make sure to have the same type on the attribute he wants to sort on for all these documents.

- In the case where an attribute is specified as a sort criterion at search time and does not exist on a document, the document will be placed at the end of the ranking rule sort.


Kerollmops (approved):
Nothing to say, that's good to me! You can [see the advancement of this feature on milli](https://github.com/meilisearch/milli/pull/309).

PR #64, opened 2021-08-22
Patch error codes for csv and ndjson formats specs

gmourier:
On closer inspection, the previous error definitions for the first iteration of the specifications are insufficient.

This patch adds a more detailed error definitions concerning `content-type` and request body payload for POST/PUT routes for the `indexes/:index_uid/documents` endpoint

### DIFF
--- a/text/0028-indexing-csv.md
+++ b/text/0028-indexing-csv.md
@@ -130,14 +130,13 @@ curl \
 
 ##### Errors
 
-- 🔴 Sending a different `Content-Type` than `application/json` or `text/csv` will lead to 415 Unsupported Media Type - **unsupported_media_type** error code.
-- 🔴 Sending an empty `Content-Type` will lead to a 415 Unsupported Media Type - **unsupported_media_type** error code.
-- 🔴 Omitted `Content-Type` header will lead to a 415 Unsupported Media Type - **unsupported_media_type** error code.
-- 🔴 Sending a different payload type than the `Content-Type` header should return a 400 Bad Request - **invalid_payload_type** error code.
+- 🔴 Sending a different `Content-Type` than `application/json` or `text/csv` will lead to 415 Unsupported Media Type - **invalid_content_type** error code.
+- 🔴 Sending an empty `Content-Type` will lead to a 415 Unsupported Media Type - **invalid_content_type** error code.
+- 🔴 Omitted `Content-Type` header will lead to a 415 Unsupported Media Type - **missing_content_type** error code.
+- 🔴 Sending a different payload type than the `Content-Type` header should return a 400 Bad Request - **malformed_payload** error code.
 - 🔴 Sending a payload excessing the limit will lead to a 413 Payload Too Large - **payload_too_large** error code.
 - 🔴 Sending an empty payload will lead to a 400 Bad Request - **missing_payload** error code.
-- 🔴 Sending an invalid CSV format will lead to a 400 bad_request - **invalid_payload** error code.
-- 🔴 A payload encoding different from `utf-8` will lead to a 400 Bad Request **invalid_payload** error code.
+- 🔴 Sending an invalid CSV format will lead to a 400 bad_request - **malformed_payload** error code.
 
 ## 2. Technical details
 n/a
--- a/text/0028-indexing-csv.md
+++ b/text/0028-indexing-csv.md
@@ -138,6 +138,99 @@ curl \
 - 🔴 Sending an empty payload will lead to a 400 Bad Request - **missing_payload** error code.
 - 🔴 Sending an invalid CSV format will lead to a 400 bad_request - **malformed_payload** error code.
 
+##### Errors Definition
+
+## missing_content_type
+
+### Context
+
+This error occurs when the Content-Type header is missing.
+
+### Error Definition
+
+HTTP Code: `415 Unsupported Media Type`
+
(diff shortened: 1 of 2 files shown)


text/0028-indexing-csv.md
```diff
@@ -95,63 +98,142 @@ the search result should be displayed as
...
 
-### Add or Update Documents [📎](https://docs.meilisearch.com/reference/api/documents.html#add-or-update-documents)
+**PUT documents** `/indexes/:indexUid/documents`
 
 ```bash
 curl \
   -X PUT 'http://localhost:7700/indexes/movies/documents' \
   -H 'Content-Type: text/csv' \
-  --binary-data '
+  --data-binary '
```
bidoubiwa, line 124: Why does `--data` not work?

text/0028-indexing-csv.md
```diff
@@ -95,63 +98,142 @@ the search result should be displayed as
...
     "id","label","price:number","colors","description"\n
     "1","hoodie","19.99","purple","Hey, you will rock at summer time."
   '
 ```
-> Response code: 202 Accepted
+> 202 Accepted - Response
+
+##### Errors
+
+- 🔴 Omitted `Content-Type` header will lead to a 415 Unsupported Media Type - **missing_content_type** error code.
```
bidoubiwa, line 133: with every request?
This seems a bit overkill as:
- JSON is what we showcase everywhere and is in the notorious file types, the most  adequate to our usage. 
- JsonL is not that popular yet as it is less natively supported by the standard libraries of most languages. 
- CSV is not suited for facetting thus is not as effective
- A big part of our users may be front-end devs that need to implement a search solution. Devs from all levels start to learn the very basics of what they need. I feel like `headers` is a knowledge that most will have to acquire to use meilisearch. And thus, add complexity and friction in the usage 

I checked it out with [elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html).

> Request body
> <field>
> (Required, string) Request body contains the JSON source for the document data.

I think becoming more complex than elasticsearch may not be the direction to take.

So finally: 

Is it best to add complexity to the majority of users to, I guess, less confuse, users that will use jsonl or csv?
Would it not be better for these kind of documents to have a error code that describes that in case of another type of file than JSON we require the adequate content-type?

Obviously, these are my thoughts and they are based on a philosophy of simplicity to implement and emporewment of all devs.

gmourier: The specification only concerns the `POST` `/indexes/:uid/documents` and `PUT` `/indexes/:uid/documents`. 

However, I think that improving the error messages on other routes waiting for a `json` body improves the user experience and makes the API more consistent.

This is content-type check is already on for the `POST` `/search` route, `POST` `/indexes`, `PUT` `/indexes/:uid`, all `POST` `/settings`. We just don't specify what `content-type` is expected in the error message so it's even worse than the spec proposal which is to improve the error message.

Having an explicit error message is much more recommended for the developer experience. It's also, I think, rather recommended to have consistent behaviors throughout the API.

Elasticsearch chose to add the [Content-Type checking for several reasons](https://www.elastic.co/fr/blog/strict-content-type-checking-for-elasticsearch-rest-requests).

All our SDKs specify the content-type today, only our cURL examples add a default content-type that does not conform to the json format, cURL takes this freedom, fortunately otherwise users would get an error that does not specify the expected format. 

Beginners will first of all refer to our SDKs before reinventing the wheel. The content-type is not useless and allows to warn the server that we expect this data format. It is also useful for the developer since an explicit message avoids him to search in the documentation. Thus accelerating its use.

I haven't seen any feedback on this yet even though the content-type check is already present.

bidoubiwa: It's true that it only concerns http users. I'm not 100% convinced that beginner js users for example would rather use `meilisearch-js` than `axios` to create requests. 
Nonetheless, after reading elasticsearch article, it might indeed be a good idea to introduce this as soon as possible for the security. ElasticSearch does state that they are changing their philosophy of usage though: 

> As Elasticsearch has evolved we’ve made a conscious decision to favour reliability and predictability over leniency.

But, as the code-samples in the documentation will all showcase a content-type stating what meilisearch is expecting, I might compensate the complexity of having to search for this information everywhere (for example the API references of elastic search not showing that it is needed haha)! 


Also a decision will have to be made on the SDK side aswell. Do they only give the possibility to send documents through their "array of object" data type, or do we provide the possibility to send `jsonl` and `csv` through the SDK as well. Maybe this is a question for another topic though!

text/0028-indexing-csv.md
```diff
@@ -129,29 +126,113 @@ curl \
     "1","hoodie","19.99","purple","Hey, you will rock at summer time."
   '
 ```
-> Response code: 202 Accepted
+> 202 Accepted - Response
+
+##### Errors
+
+- 🔴 Sending a different `Content-Type` than `application/json` or `text/csv` will lead to 415 Unsupported Media Type - **invalid_content_type** error code.
+- 🔴 Sending an empty `Content-Type` will lead to a 415 Unsupported Media Type - **invalid_content_type** error code.
```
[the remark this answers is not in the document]
curquiza: Checked with @gmourier, I'm responsible for synchronizing the docs team to update the docs according to this

curquiza (approved):

PR #66, opened 2021-08-31
Patch GeoSearch specification to mention technical limit on `desc` ordering around a _geoPoint

gmourier:
As imagined during the first phases of discovery we encounter a technical difficulty to compute a descending order around a `_geoPoint`.

The technical team will have to do some preparatory work to overcome this limitation. For the time being, it is not planned to spend more time on this as it is not a primary use case. The impact may be very small, so it's not worth the effort at this point.

### Decision taken

It was decided after a discussion with @irevoire and @Kerollmops that when the user specifies the sort rule `_geoPoint(lat, lng):desc` an error `invalid_sort` will be returned with a message explaining that `:desc` is not available.

#### Why keep the `:asc` if `:desc` is not valid?

To keep consistency and not to introduce a different syntax among the `sort` search parameter. This is something we'll re-evaluate later.


gmourier:
Pinging @guimachiavelli  and @meilisearch/integration-team to share this information.

curquiza (approved):

PR #72, opened 2021-09-13
Add cursor-based pagination for `task` resource lists

gmourier:
# Summary

`task` resource lists now support cursor pagination, allowing users to browse multiple sets of `task` items.

## Summary Key Points

- Add `limit` and `after` response attributes to task list response.
- It concerns `/tasks` and `/indexes/:uid/tasks`.

# Motivation

Following the specification aiming to stabilize the `task` API resource, we want to give MeiliSearch capabilities to adapt to many use-cases, thus involving unpredictable orders of magnitude regarding `task` growth over time.

This feature should be usable by both solo developers and technical teams.

The objectives set during the discovery's exploration phase for the pagination of the task lists are:

- Performance should remain nearly constant, regardless of task volume.
- The results must be browsable in a consistent way between each call to the API, no matter how much the number of tasks is increased or reduced while the navigation is happening.
- Do not reinvent the wheel and therefore facilitate its implementation by selecting a proven paging interface.
- A subsidiary, but critical objective, is to feed our API guideline to respond quickly to such needs in the future within the company. By proposing a proven solution, we can build faster with confidence.

## [Rendered](https://github.com/meilisearch/specifications/blob/0ed4253d4656a944118fb3e9e83e1fcab861f331/text/0072-task-resource-lists-pagination.md)


MarinPostma (dismissed):

curquiza (dismissed):
💪

alallema (dismissed):
LGTM! 🎉

gmourier:
Closed in favor of https://github.com/meilisearch/specifications/pull/115

PR #75, opened 2021-09-14
Add ordering capabilities on `task` lists endpoints.

gmourier:
# Summary

Add sorting capabilities by `desc`/`asc` on `task` `uid` attribute to the `tasks` endpoints to facilitate the management of an instance and its indexes, or a specific index.

## Summary Key Points

- Add ordering capabilities on `task` `uid` for `GET` `tasks` endpoints.

# Motivation

Following the specification aiming to stabilize the `task` API resource, we want to give users the capability to browse lists of tasks according to the direction of their choice.

### DIFF
--- a/open-api.yaml
+++ b/open-api.yaml
@@ -362,7 +362,6 @@ components:
         - sort
         - exactness
         - release_date:asc
-      examples: []
     filterableAttributes:
       type: array
       description: |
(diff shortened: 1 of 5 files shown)


gmourier:
Closed for the moment, will be rewritten from this base.

PR #77, opened 2021-09-21
Patch GeoSearch specification to add descending order capability to _geoPoint built-in sort

gmourier:
We were able to find a solution to implement sorting around a `_geoPoint` in descending order. 

The specification is updated to remove the mention of this initial limitation.

@meilisearch/docs-team FYI


irevoire (approved):
lgtm :+1:

PR #78, opened 2021-09-22
Add variant message to ensure _geoPoint and _geoRadius expressions are not used as a custom ranking rule

gmourier:


irevoire (approved):
yes!

PR #83, opened 2021-10-13
GeoSearch — Support string type for `_geo` `lat` and `lng` fields

gmourier:
Allows us to improve the user experience when using GeoSearch. 

When indexing the `_geo` field as a filterable or sortable attribute, MeiliSearch should be able to parse string or numeric values rather than raise an `invalid_geo_field` error.


gmourier:
If there is no major technical disadvantage to be able to mix the type of the fields, it would be better for the user.

PR #85, opened 2021-10-15
API keys

gmourier:
# Summary

Granular management of API keys is added to MeiliSearch. It is possible to restrict the access of an API key to certain actions on specific indexes.

## Summary Key Changes

- `X-MEILI-API-KEY` header is replaced by the `Authorization` header. `API keys` must be specified with the bearer authorization method. See examples.
- `/keys` management is restricted to the master key.
- When a master key is set at MeiliSearch first-launch, we generate two pre-configured default `API Key` resources. A `Default Search API Key` restricted to the search action and a `Default Admin API Key` to handle all operations (except managing `/keys` resource) on MeiliSearch.
- If the master-key changes, all `API Keys` are re-generated.
- These default API Keys can be modified/deleted with the `/keys` endpoint but are not re-created if MeiliSearch has already generated them at some point.
- New endpoints are added to manage the `API Key` resource.
- `API keys` can have restrictions on which methods can be accessed via an `actions` list; they can also `expiresAt` a specific date time and be restricted to a specific set of `indexes`.
- There is no possibility to regenerate the value of the `key` field for an `API key` in this first iteration.
- New errors are added and the `missing_authorization_header` message is updated.
- The current state of the `API Keys` resource is propagated to snapshots and dumps.

# Motivation

To make MeiliSearch more reliable for teams, we extend the management and the possibilities of restrictions for an `API Key`.  Security is a critical need, often tricky to negotiate as the stakes are high for a company.


ManyTheFish:
Hey @gmourier could we split the `API Keys` part and the `Scoped Key` part? I think we can implement the first part without having  the second

text/0085-api-keys.md
```diff
@@ -19,7 +19,7 @@ To make MeiliSearch more reliable for teams, we extend the management and the po
 
 | Term               | Definition |
 |--------------------|------------|
-| Master Key         | This is the master key that allows you to create other API keys. The master key is defined by the user when launching MeiliSearch. |
+| Master Key         | This is the master key that allows you to create other API keys. The master key is defined by the user when launching MeiliSearch, thus give access to the `/keys` API endpoints. |
```
bidoubiwa, line 22:
```suggestion
| Master Key         | This is the master key that allows you to create other API keys. The master key is defined by the user when launching MeiliSearch. The key gives access to the `/keys` API endpoint. |
```

Or if you prefer to keep your wording: 
```suggestion
| Master Key         | This is the master key that allows you to create other API keys. The master key is defined by the user when launching MeiliSearch, thus give access to the `/keys` API endpoint. |
```

text/0085-api-keys.md
```diff
@@ -0,0 +1,450 @@
...
+
+`Anna` manages the MeiliSearch instance; she uses the master key she defined at startup to generate an `API Key` resource for `Mark` to use in his client code to communicate with the MeiliSearch instance.
+
+`Anna` can define access rights to certain indexes to define an expiration date and also authorized `actions` for an `API Key` (See API Key Actions List Definition Part).
+
+Only the master key allows managing the API keys.
+
+#### Master Key
+
+The master key exists to secure a MeiliSearch instance. As soon as a master key is set via the  `MEILI_MASTER_KEY` environment variable or the `--master-key` CLI option , the endpoint `/keys` is accessible only for the master key holder. It can be seen as a super admin key; It must be shared only with people who have to manage the security of a MeiliSearch instance.
```
bidoubiwa, line 53:
```suggestion
The master key exists to secure a MeiliSearch instance. As soon as a master key is set via the  `MEILI_MASTER_KEY` environment variable or the `--master-key` CLI option , most of the routes become protected. Using the master key you are able to access the `/keys` endpoint to fetch the authorization keys.

The master key can be seen as a super admin key; It must be shared only with people who have to manage the security of a MeiliSearch instance.
```

text/0085-api-keys.md
```diff
@@ -0,0 +1,450 @@
...
+
+This master key is not an API Key, thus is not stored and fetchable from an API endpoint. It must be seen as a runtime lock that activates the security of MeiliSearch as soon as an instance is launched with it.
+
+At the first launch of MeiliSearch with a master key, MeiliSearch automatically generate two default API keys (See the `GET - /keys` example) to cover many of the most basic needs. It generates a `Default Search API Key` dedicated to the search that can be used on the client-side and a `Default Admin API Key` to manipulate a MeiliSearch instance from a backend side.
+
+If the value of the master key changes, the previously generated `API Keys` do not change or expire. Thus, MeiliSearch does not re-generate the previous generated default API keys.
```
bidoubiwa, line 59: If we relaunch meilisearch with a new master key the private and public key remain the same?

gmourier: Yes, the `Default Search API Key` and the `Default Admin API Key` remain the same to add extra work by updating all clients using these keys.

open-api.yaml
```diff
@@ -517,6 +517,71 @@ components:
...
+            - string
+            - 'null'
+          description: Represent the date and time as ISO-8601 format when the API key has been updated. Generated by MeiliSearch.
+          default: null
+          readOnly: true
+      required:
+        - actions
+        - indexes
+        - expiresAt
+      description: API keys are stored and managed by the master key holder. These are the keys used by the technical teams to interact with MeiliSearch at the level of the client/server code. API keys can have restrictions on which methods can be accessed via an actions list; they can also expiresAt a specific date time and be restricted to a specific set of indexes.
```
curquiza, line 584:
```suggestion
      description: API keys are stored and managed by the main key holder. These are the keys used by the technical teams to interact with MeiliSearch at the level of the client/server code. API keys can have restrictions on which methods can be accessed via an actions list; they can also expiresAt a specific date time and be restricted to a specific set of indexes.
```

open-api.yaml
```diff
@@ -760,7 +825,7 @@ tags:
       The `tasks` route gives information about the progress of the [asynchronous tasks](https://docs.meilisearch.com/learn/advanced/asynchronous_tasks.html).
   - name: Keys
     description: |
-      Each instance of MeiliSearch has three keys: a master, a private, and a public. Each key has a given set of permissions on the API routes.
+      Manage api `keys` for a MeiliSearch instance. Each key has a given set of permissions.
       You must have the master key to access the `keys` route.
```
curquiza, line 829:
```suggestion
      You must have the main key to access the `keys` route.
```

open-api.yaml
```diff
@@ -2363,16 +2428,82 @@ paths:
...
+    delete:
+      summary: Delete an API key
+      operationId: keys.delete
+      tags:
+        - Keys
+      responses:
+        '204':
+          $ref: '#/components/responses/204'
+      description: ''
+    patch:
```
curquiza, line 2467: Are we sure we use patch instead of put since it's not consistent with the other route?

text/0085-api-keys.md
```diff
@@ -0,0 +1,453 @@
...
+
+At the first launch of MeiliSearch with a main key, MeiliSearch automatically generate two default API keys (See the `GET - /keys` example) to cover many of the most basic needs. It generates a `Default Search API Key` dedicated to the search that can be used on the client-side and a `Default Admin API Key` to manipulate a MeiliSearch instance from a backend side.
+
+If the value of the main key changes, the previously generated `API Keys` do not change or expire. Thus, MeiliSearch does not re-generate the previous generated default API keys.
+
+If the main key is removed at MeiliSearch launch, the previously generated API keys no longer secure the MeiliSearch instance.
+
+If MeiliSearch is launched with the `production` value for the `MEILI_ENV` environement variable or the `--env` CLI option, a main key is mandatory to force the user to secure his instance. If the main key is omitted in that particular case, MeiliSearch launch is aborted and display the `Error: In production mode, the environment variable MEILI_MAIN_KEY is mandatory` error in stdout.
```
curquiza, line 66:
```suggestion
If MeiliSearch is launched with the `production` value for the `MEILI_ENV` environment variable or the `--env` CLI option, the main key is mandatory to force the user to secure his instance. If a main key is omitted in that particular case, MeiliSearch launch is aborted and displays the `Error: In production mode, the environment variable MEILI_MAIN_KEY is mandatory` error in stdout.
```

text/0085-api-keys.md
```diff
@@ -0,0 +1,453 @@
...
+
+Only the main key allows managing the API keys.
+
+#### Main Key
+
+The main key exists to secure a MeiliSearch instance. As soon as a main key is set via the  `MEILI_MAIN_KEY` environment variable or the `--main-key` CLI option , the endpoint `/keys` is accessible only for the main key holder. It can be seen as a super admin key; It must be shared only with people who have to manage the security of a MeiliSearch instance.
+
+This main key is not an API Key, thus is not stored and fetchable from an API endpoint. It must be seen as a runtime lock that activates the security of MeiliSearch as soon as an instance is launched with it.
+
+At the first launch of MeiliSearch with a main key, MeiliSearch automatically generate two default API keys (See the `GET - /keys` example) to cover many of the most basic needs. It generates a `Default Search API Key` dedicated to the search that can be used on the client-side and a `Default Admin API Key` to manipulate a MeiliSearch instance from a backend side.
```
curquiza, line 60:
```suggestion
At the first launch of MeiliSearch with a main key, MeiliSearch automatically generates two default API keys (See the `GET - /keys` example) to cover many of the most basic needs. It generates a `Default Search API Key` dedicated to the search that can be used on the client-side and a `Default Admin API Key` to manipulate a MeiliSearch instance from a backend side.
```

text/0085-api-keys.md
```diff
@@ -0,0 +1,453 @@
...
+|-------------|---------|--------------------------------------------------|
+| key         | string  | The generated key. **Generated by MeiliSearch**. |
+| actions     | array   | A list of actions permitted for the key. `["*"]` for all actions. See Actions list definition part. |
+| indexes     | array   | A list of indexes permitted for the key. `["*"]` for all indexes. |
+| expiresAt   | string  | Represent the expiration date and time as `ISO-8601` format. `null` equals to no expiration time. |
+| description | string  | A description for the key. `null` if empty. |
+| createdAt   | string  | Represent the date and time as `ISO-8601` format when the API key has been created. **Generated by MeiliSearch** |
+| updatedAt   | string  | Represent the date and time as `ISO-8601` format when the API key has been updated. **Generated by MeiliSearch** |
+
+#### `POST`/`PUT` - `/keys` - API Key object payload definition
```
curquiza, line 84: Put or patch then?

text/0085-api-keys.md
bidoubiwa: Can you change them in any way?

gmourier: Yes, they are editable!

open-api.yaml
gmourier: Yes, I forgot to update the spec for that but we should use `patch` to be in accordance with the use of HTTP verbs. As we're going to pass on to most of the HTTP routes before 1.0, we'll have to update them when we can. We could also do this at once to correct the use of verbs already.

text/0085-api-keys.md
gmourier: `patch`, it was a miss on my side 🥲. The route described below uses the verb `patch`.

curquiza (approved):

text/0085-api-keys.md
bidoubiwa: How can you edit them?

text/0085-api-keys.md
```diff
@@ -0,0 +1,463 @@
...
+
+### V. `API Key` Explanations
+
+#### Summary Key Points
+
+- Occurrences of `master key` are changed by `main key`.
+- `MEILI_MASTER_KEY` environment variable is renamed `MEILI_MAIN_KEY`.
+- `--master-key` CLI option is renamed `--main-key`.
+- `/keys` management is restricted to the main key.
+- When a main key is set at MeiliSearch first-launch, we generate two pre-configured `API Key` resources. A `Default Search API Key` restricted to the search action and a `Default Admin API Key` to handle all operations (except managing API Keys) on MeiliSearch.
```
guimachiavelli, line 40: What is the `expiresAt` value of the pre-configured keys? Are they any different from user-generated keys? Could a user change the `default search api key` to, say, include `document.add`?

text/0085-api-keys.md
```diff
@@ -0,0 +1,463 @@
...
+
+#### Response
+
+- 🔴 Accessing this route with an `API Key` that don't have sufficient permissions to access it returns an [invalid_api_key](0061-error-format-and-definitions.md#invalid_api_key) error.
+
+## 2. Technical Aspects
+
+### Synchronous write of `API Key` resources
+
+Synchronous writing would make it possible to dissociate key management from the update store, which handles asynchronous writing.
```
guimachiavelli, line 452: I'm a bit confused: will the API Key endpoint be sync or async? The phrasing makes me interpret this as a suggestion, but I'm assuming I'm wrong?

guimachiavelli:
Possibly stupid question: why is this PR labelled as Q4:2022? Shouldn't it be Q4:2021?

text/0061-error-format-and-definitions.md
```diff
@@ -89,6 +89,123 @@ HTTP code `400 Bad Request`
 }
 ```
 
+- The `:errorMessage` is inferred when the message is generated.
+
+---
+
+## missing_parameter
```
guimachiavelli, line 96: Is this error tied to this particular spec?

gmourier: Yes, this is an error to distinguish a parameter that would be invalid and a mandatory parameter that is not filled in for a resource.

So we would have a more general `missing_parameter` error when a mandatory parameter is not filled in for a resource in the payload. 

We keep the `invalid_:resourceName_:parameterName` convention that we already had with `invalid_index_uid` for example when the field found is not valid.

text/0085-api-keys.md
gmourier: Indeed, it is not clear. Thanks, @guimachiavelli. I rephrased that part to make it clear that the operations on /keys are synchronous.

text/0085-api-keys.md
```diff
@@ -0,0 +1,463 @@
...
+
+- Occurrences of `master key` are changed by `main key`.
+- `MEILI_MASTER_KEY` environment variable is renamed `MEILI_MAIN_KEY`.
+- `--master-key` CLI option is renamed `--main-key`.
+- `/keys` management is restricted to the main key.
+- When a main key is set at MeiliSearch first-launch, we generate two pre-configured `API Key` resources. A `Default Search API Key` restricted to the search action and a `Default Admin API Key` to handle all operations (except managing API Keys) on MeiliSearch.
+- New endpoints are added to manage the `API Key` resource.
+- `API keys` can have restrictions on which methods can be accessed via an `actions` list; they can also `expiresAt` a specific date time and be restricted to a specific set of `indexes`.
+
+#### Managing `API Key`
```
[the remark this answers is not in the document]
gmourier: I moved the part about changing the `master` key to `main` key above.

The summary key points part is there to describe all the changes of the specification to have a wide spectrum at a glance.

text/0085-api-keys.md
```diff
@@ -0,0 +1,463 @@
...
+
+`Anna` manages the MeiliSearch instance; she uses the main key she defined at startup to generate an `API Key` resource for `Mark` to use in his client code to communicate with the MeiliSearch instance.
+
+`Anna` can define access rights to certain indexes to define an expiration date and also authorized `actions` for an `API Key` (See API Key Actions List Definition Part).
+
+Only the main key allows managing the API keys.
+
+#### Main Key
+
+The main key exists to secure a MeiliSearch instance. As soon as a main key is set via the  `MEILI_MAIN_KEY` environment variable or the `--main-key` CLI option , the endpoint `/keys` is accessible only for the main key holder. It can be seen as a super admin key; It must be shared only with people who have to manage the security of a MeiliSearch instance.
```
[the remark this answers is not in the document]
gmourier: The `keys` endpoints will be inaccessible since the user will not be able to fill in a valid `main` key if he has not generated one at the MeiliSearch launch. I'll fill this in! Thank you for pointing this.

text/0085-api-keys.md
```diff
@@ -0,0 +1,463 @@
...
+| updatedAt   | string  | Represent the date and time as `ISO-8601` format when the API key has been updated. **Generated by MeiliSearch** |
+
+#### `POST`/ `PATCH` - `/keys` - API Key object payload definition
+
+| field       | type    | required |description                     |
+|-------------|---------|----------|--------------------------------|
+| indexes     | array   | Required | `[*]` for all indexes. **Default**: `No Default` |
+| description | string  | Optional | A description for the API key. **Default**: `null` |
+| actions     | array   | Required | A list of actions permitted for the API key. `["*"]` for all actions. **See Actions list definition part**. `*` character can be used as a wildcard. e.g. `documents.*` to authorize access on all documents endpoints. **Default**: `No default` |
+| expiresAt   | string  | Required | The expiration date and time as `ISO-8601` format. `null` equals to no expiration time. Sending only the date part e.g `2021-12-01` leads to having an `expiresAt` value set to `2021-12-01T00:00:00`. **Default**: `No Default` |
```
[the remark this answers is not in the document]
gmourier: > Why make it mandatory and allow users to set it to `null` anyway?

Mostly for security concerns. By making it mandatory, it force the user to think of it and explicitly set that the API Key being generated will not have an expiration date by passing `null` instead of omitting the field.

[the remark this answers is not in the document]
gmourier: > Also, are there any best/recommended practices regarding `expiresAt`? E.g.: avoid creating keys that never expire, avoid creating many keys with short expiry dates, etc

Very good question @guimachiavelli. 

In general, it is a good practice to always put expiration dates for security reasons. The fact of not putting any is in the eyes of the user who must understand the importance of his decision.

For a company, it also allows to have governance on the management of the keys and to make sure that the rotation of the keys is done at regular intervals. Regularly scheduled changes to an API secret ensure that their app is less vulnerable to being compromised.

text/0085-api-keys.md
gmourier: I will dedicate a part to its `Default API keys` generated by MeiliSearch.

These keys do not have an expiration date by default but can be modified at the convenience of the user. What is interesting is that Algolia for example in its system does not allow to modify this type of key.

They are mostly there to accelerate the implementation of the user's architecture like the public and private keys we had before.

In general, in a production environment, you must avoid having an API key with the search action that shares other actions because the search API key will most often be provided in the front-end code exposed to end-users and a malicious end-user could spoof it to perform unwanted operations on a MeiliSearch instance.

open-api.yaml
```diff
@@ -517,6 +517,71 @@ components:
...
+      type: object
+      properties:
+        key:
+          type: string
+          description: The generated key. Generated by MeiliSearch.
+          example: d0552b41536279a0ad88bd595327b96f01176a60c2243e906c52ac02375f9bc4
+          readOnly: true
+        actions:
+          type: array
+          description: 'A list of indexes permitted for the key. ["*"] for all indexes.'
```
dichotommy, line 531: I think this description and the description of `indexes` are switched.

guimachiavelli:
A couple of follow-up questions from the @meilisearch/docs-team:

1. What is the upgrade process for users coming from <v0.25? Will `master` be automatically converted into `main` and the two automatically generated API keys be mapped neatly into the current `public` and `private` keys? 

2. Users can set `expiresAt` to `null` if they want a key to last for an indefinite amount of time. @dichotommy mentioned that we do not allow people to send `null` as parameter values in other API endpoints. Will we be making an exception for this endpoint? What happens if a user sets `expiresAt` to an empty string?

text/0061-error-format-and-definitions.md
guimachiavelli: So, to confirm: this error is introduced in this spec, but we expect to use it for other endpoints in the future, correct?

gmourier: Yes @guimachiavelli!

open-api.yaml
gmourier: Thanks, @dichotommy! Good catch!

dichotommy:
> What is the upgrade process for users coming from <v0.25? Will master be automatically converted into main and the two automatically generated API keys be mapped neatly into the current public and private keys?

If I understand the current behavior correctly, each time the MeiliSearch instance is started the master key must be defined (if it is being used) and new public and private keys are generated in response. Since updating requires restarting the instance anyway, there shouldn't be any need to "convert" keys into the new format. Please correct me if I'm wrong.

However, the question about how the two new automatically generated keys correspond to the current public and private keys is a very pertinent one. Can users expect to use them interchangeably, or will access change?

gmourier:
> > What is the upgrade process for users coming from <v0.25? Will master be automatically converted into main and the two automatically generated API keys be mapped neatly into the current public and private keys?
> 
> If I understand the current behavior correctly, each time the MeiliSearch instance is started the master key must be defined (if it is being used) and new public and private keys are generated in response. Since updating requires restarting the instance anyway, there shouldn't be any need to "convert" keys into the new format. Please correct me if I'm wrong.
> 
> However, the question about how the two new automatically generated keys correspond to the current public and private keys is a very pertinent one. Can users expect to use them interchangeably, or will access change?

I have filled in part 1.5.3 about the update process and what it means for the user.

It's transparent to the user except for the fact that they will have to change the name of the variable used at launch whether it's `--master-key` or `MEILI_MASTER_KEY`. We can keep the same way we generate the public and private for the `Default Search API KEY` and the `Default Admin API Key` so the user doesn't have to change the frontend code.

If the user upgrades to v0.25 and keeps the same value for the `main` key as the previous `master` key. MeiliSearch will generate the two default search and admin API keys with the value of the previous `public` and `private` key transparently.

These two default API Keys can be modified or deleted.

If MeiliSearch has previously generated those API keys for a particular `data.ms`, it will note regenerated them in every case (e.g. changing the value of the `main` key).

Now thinking about it, I think the master key is still semantically appropriate and we should avoid changing this name. It would avoid an extra step for the user and it would also simplify the set of changes to be made on all our repos and avoid the user having to do gymnastics on this subject if he has an instance at a version prior to v0.25.0 and he reads the documentation.

Semantically, the term master key is a real term used in locksmithing e.g. "**a key that can be used to open any of several different locks**". The term master here refers to the fact that the master key can be used to open any of several different locks, whereas the term main seems more vague.

I would like to have the opinion of the @meilisearch/docs-team on this before potentially making the decision to keep `master` or `main`.

If you just have a spontaneous feeling to give without any particular explanation, you can put 👍  to keep the `master` key or 👎  to keep the `main` term.

Thanks a lot for all your feedbacks, it helps a lot to improve the specification.

gmourier:
> 2. Users can set `expiresAt` to `null` if they want a key to last for an indefinite amount of time. @dichotommy mentioned that we do not allow people to send `null` as parameter values in other API endpoints. Will we be making an exception for this endpoint? What happens if a user sets `expiresAt` to e.g. an empty string?

To answer you @guimachiavelli, in fact, we wanted to force the user to think about security rather than to go fast. By making the field mandatory, he won't be able to forget this question of defining if this key must have an expiration date or not. The only way to define a null expiration date is to set the value `null` on this field.

If the API accepts a missing `expiresAt` field at creation, the user could easily make a mistake and issue keys that don't have an expiration date. This is really to emphasize the security of the key creation process.

We have some routes that accept `null` as a payload. Every `PUT` route under the `/settings` accepts that. This could change because it is a completely different case. 🤓

guimachiavelli:
Regarding key naming: I think `master` is immediate and, as you note, it has a real-life counterpart that makes it sound logical and easy to remember. My main question is: is there any difference between the current `master` key and the proposed `main` key? If there isn't and `master` and `main` are exactly the same thing, then I vote for keeping `master`; if they are somehow different, well, they should have different names to indicate that.

---

Regarding `expiresAt` and `null`: so, just to be 100% sure, if a user were to send `''` or `false` as a value for `expiresAt`, they'd receive an error response, correct?

dichotommy:
I greatly prefer master key over main key, since the first one has a real-world meaning (key that opens all other locks in the house) whereas the other one is very conceptual. skeleton key has a similar meaning but i don't think it would communicate as well for non-native speakers.

super admin key is fantastic 🌈  but possibly too confusing w/ the automatically generated default admin api key. what about just "super key"? 

> To sum it up the public key is transformed to a real API key resource having the exact same rights as the current public key. This is also the case for private transformed to a real API Key.
> ...
> If MeiliSearch has previously generated those API keys for a particular data.ms, it will not regenerate them in every case (e.g. changing the value of the main key).

So if I understand correctly, these two keys will not be generated for completely new users who boot up MeiliSearch for the first time in v0.25.

MarinPostma:
also master key is a cryptography concept, whose meaning is "the key used to generate other key"

https://www.ibm.com/docs/en/zos/2.3.0?topic=keys-master-key-concept
https://link.springer.com/referenceworkentry/10.1007%2F978-1-4419-5906-5_86

I mean, there is a word that is exactly the definition of what we are trying to do, why not use it?

gmourier:
Thank you all for your feedback which is a great help. 

After much thought, we will push to be as consistent as possible with what we consider to be the best choices to move forward today towards a stable version. Compared to the previous explanations I gave to the @meilisearch/docs-team team about the update process, this one will change and will include breaking changes that will be explained.

1. We keep the `master` key term for now. The technical term is straightforward. We may change this in the future if the need arises. Today for ease of understanding, we keep it because we believe that the other terms are less easy to make their meaning equivocal. We need to move forward.

2. Someone upgrading to v0.25 from an earlier version will have to update the client code to use the newly generated keys that replace `public` and `private` even if they use the same value for their `master` key. In previous versions, the private and public keys were tightly coupled to the master key value and were generated in cascade. The `default API Keys` will now be generated independently of each other like any API Key resource. This also avoids having technical debt in security-related code and a potential security issue.

3. We'll take advantage of this new feature to use the `Authorization` HTTP header dedicated to this rather than our custom `X-MEILI-API-KEY` header.

In short, we'll push as much breaking as possible now rather than later while we're at it, so we don't spread it out over several versions when it comes to authorization and API key management.

gmourier:
> Regarding key naming: I think `master` is immediate and, as you note, it has a real-life counterpart that makes it sound logical and easy to remember. My main question is: is there any difference between the current `master` key and the proposed `main` key? If there isn't and `master` and `main` are exactly the same thing, then I vote for keeping `master`; if they are somehow different, well, they should have different names to indicate that.
> 
> Regarding `expiresAt` and `null`: so, just to be 100% sure, if a user were to send `''` or `false` as a value for `expiresAt`, they'd receive an error response, correct?

They are exactly the same!

Concerning `expiresAt` field validation, you understand it well @guimachiavelli

gmourier:
> So if I understand correctly, these two keys will not be generated for completely new users who boot up MeiliSearch for the first time in v0.25.

The `Default Search API Key` and the `Default Admin API Key` will be generated as soon as the user launch MeiliSearch with a `master-key` as of today with the `private` and `public`. 

But contrary to what is done today, they will not be regenerated if the master key changes.

text/0085-api-keys.md
```diff
@@ -0,0 +1,530 @@
...
+
+> 🚨 The master-key should never be exposed to the public or bad-intentioned persons for security measures as it may compromise a MeiliSearch instance.
+
+> Note that the master key does not appear on the `/keys` endpoints.
+
+#### 1.5.3 Default API Keys
+
+When the user accessing the machine launches MeiliSearch with a `master` key the first time, MeiliSearch will generate two API keys described below, as it did before with the `public` and `private` key.
+
+If the user changes the value of the `master` key later, these two default keys are not modified. However, these two API keys can be changed using the `/keys' endpoints.
```
dichotommy, line 70:
```suggestion
If the user changes the value of the `master` key later, these two default keys are not modified. However, these two API keys can be changed using the `/keys` endpoints.
```

gmourier:
A change that will surely interest you @meilisearch/docs-team.

@ManyTheFish has been working on an elegant way to have increased security.

If the value of the master key changes, all previously generated `API Keys` change, allowing all previously generated keys to be invalidated and regenerated with a different value for the `key` field. This is particularly useful in the case where the master key has been leaked and the user needs to regenerate the entire set of keys at once to re-secure the instance.

The advantage is that it saves the instance administrator from having to delete and recreate the keys one by one when the issue is critical at that moment. It just needs to update the master-key value.

Another advantage is that dumps and snapshots if they leak cannot be retro-engineered. We store an 8 character prefix that the master-key completes at run-time. A person with bad intentions could not do anything with these prefixes. See the Technical details part for more information 🤓.

text/0085-api-keys.md
```diff
@@ -0,0 +1,547 @@
...
+    "updatedAt": "2021-11-12T10:00:00Z"
+}
+```
+
+##### Requirements
+
+- `actions` is mandatory and should be an array of valid `actions`.
+- `indexes` is mandatory and should be an array of string.
+- `expiresAt` is mandatory and must be a valid `ISO 8601` datetime in the future or `null`.
+- If set, `description` should be a string or `null`.
```
bidoubiwa, line 224: Is it possible to have an additional field that contains an unique name for the key? In order to be able to fetch the apiKey without having to watch with human eyes at which index the key is: 

### with name

```js 
const keys = client.getKeys()
const privateKey = keys.find(name => name == "default_private")
```

### Without name

```js
const keys = client.getKeys()
console.log(keys) // look at the description and either count at which index the right key is, or copy paste it from the terminal
```

text/0085-api-keys.md
```diff
@@ -0,0 +1,547 @@
...
+"Authorization: Bearer :masterKey"
+```
+
+##### Response
+
+`200 Success`
+
+```json
+{
+    "results": [
```
bidoubiwa, line 376: I'm not against the idea at all, just for curiosity: Did you add it in a `results` key in order to be able to add new fields at the root in the future?

gmourier: This is a convention that we will have from now on when the API lists resources. It actually allows us to add fields in the future quite easily. This is also what we did on the `task` list.

text/0085-api-keys.md
gmourier: Yes @bidoubiwa! We could also have a filter system on the indexes and action fields, also a parameter to search in the description field.

This may be in a future iteration.

text/0061-error-format-and-definitions.md
```diff
@@ -555,6 +672,27 @@ HTTP Code: `404 Not Found`
...
+### Context
+
+This error happens when a requested api key can't be found.
+
+#### Error Definition
+
+HTTP Code: `404 Not Found`
+
+```json
+    "message": "API key :apiKey not found.",
```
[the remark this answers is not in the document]
gmourier: The value is extracted from the `/keys/:key` parameter when the user with the master key requests a particular API key.

text/0085-api-keys.md
```diff
@@ -0,0 +1,547 @@
...
+- `API keys` can have restrictions on which methods can be accessed via an `actions` list; they can also `expiresAt` a specific date time and be restricted to a specific set of `indexes`.
+- There is no possibility to regenerate the value of the `key` field for an `API key` in this first iteration.
+- New errors are added and the `missing_authorization_header` message is updated.
+- The current state of the `API Keys` resource is propagated to snapshots and dumps.
+
+#### 1.5.2 Master Key
+
+The master key exists to secure a MeiliSearch instance. As soon as a master key is set via the  `MEILI_MASTER_KEY` environment variable or the `--master-key` CLI option , the endpoint `/keys` is accessible only for the master key holder. It can be seen as a super admin key; It must be shared only with people who have to manage the security of a MeiliSearch instance.
+
+This master key is not an API Key, thus is not stored and fetchable from the `/keys` API endpoint. It must be seen as a runtime lock that activates the security of MeiliSearch as soon as an instance is launched with it.
```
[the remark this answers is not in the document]
gmourier: I should precise "API Key resource". It is not an API key in the sense that it is not to be shared, is not stored, and is not displayed as a resource in the list of API Keys.

text/0085-api-keys.md
bidoubiwa: Very good!

text/0085-api-keys.md
bidoubiwa: Thanks :)

text/0061-error-format-and-definitions.md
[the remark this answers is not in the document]
bidoubiwa: Oh I see!

gmourier:
@meilisearch/docs-team After some tests of use. We finally keep the display, access, and update of expired keys on `/keys` endpoints until we implement a filter or an automatic archiving system to make users' life easier. This will allow them to extend the validity of an API key after it has expired rather than recreating one from scratch.

text/0085-api-keys.md
```diff
@@ -0,0 +1,548 @@
...
+
+### 1.5 `API Key` Explanations
+
+#### 1.5.1 Summary Key Points
+
+- `X-MEILI-API-KEY` header is replaced by the `Authorization` header. `API keys` must be specified with the bearer authorization method. See examples.
+- `/keys` management is restricted to the master key.
+- When a master key is set at MeiliSearch first-launch, we generate two pre-configured default `API Key` resources. A `Default Search API Key` restricted to the search action and a `Default Admin API Key` to handle all operations (except managing `/keys` resource) on MeiliSearch.
+- If the master-key changes, all `API Keys` are re-generated.
+- These default API Keys can be modified/deleted with the `/keys` endpoint but are not re-created if MeiliSearch has already generated them at some point.
```
curquiza, line 40: The second sentence is not clear regarding the previous point.
I understand that MeiliSearch will re-generate (so re-create) the key if the master key changes.

text/0085-api-keys.md
```diff
@@ -0,0 +1,548 @@
...
+- New endpoints are added to manage the `API Key` resource.
+- `API keys` can have restrictions on which methods can be accessed via an `actions` list; they can also `expiresAt` a specific date time and be restricted to a specific set of `indexes`.
+- There is no possibility to regenerate the value of the `key` field for a specific `API key` in this first iteration.
```
curquiza, line 43: Not sure to get what it means. What do you mean by "no possibility to regenerate the value"? You mean you have to create a new key, not to ask literally for a regeneration?

text/0085-api-keys.md
gmourier: Re-creating is adding another 2 new Default API keys to the `/keys` list while re-generating is changing the `key`  field value.

text/0085-api-keys.md
gmourier: Absolutely @curquiza. This is a future possibility!

PR #89, opened 2021-11-03
Tenant Token

gmourier:
# Summary

A `Tenant token` is generated by the user code to be used by an end-user when making search queries.

It allows users to have multi-tenant indexes and thus restricts access to documents depending on the end-user making the search request.

A Tenant Token is a JWT containing the information necessary for Meilisearch to verify it and extract permission/rules to apply it to the end user's search.

## Key Points

- `Tenant tokens` are JWTs generated on the user side by using Meilisearch SDKs or their custom code. `Tenant tokens` are not stored nor retrievable on the Meilisearch side.
- `Tenant tokens` contain rules that ensure that a `Tenant token` holder (e.g. an end-user) only has access to documents matching rules chosen at the `tenant token` creation.
- `Tenant tokens` are signed from a Meilisearch `API key` resource on the user's code.
- `Tenant tokens` must not be signed by the master key.
- `Tenant tokens` cannot be more permissive than the signing `API key`.
- `Tenant tokens` must be signed by an `API Key` having the `search` action defined.
- `Tenant tokens` can have different rules for each index accessible by the signing API key. These rules are described in the `searchRules` JSON object.
- The only rule available in the `searchRules` object is the search parameter `filter`.
- `Tenant tokens` are sent to Meilisearch via the `Authorization` header like any `API Keys` or the master key.
- When Meilisearch receives a search query emitted with a `Tenant token`, the `tenant token` is decoded, then the `searchRules` are applied for the search request before the search parameters.

# Motivation

`Tenant tokens` are introduced to solve multi-tenant indexes use-case.

> Multi-Tenant Indexes Definition: It is an index that stores documents that may belong to different tenants. In our case, a tenant within an index can be a user or a company, etc. In general, the data of one tenant should not be accessible by other tenants.

Users today need to set up workarounds to have multi-tenant indexes. They have to use server code to implement the access restriction logic before requesting Meilisearch. It isn't easy to maintain, to implement, and the performance is not optimal because the frontend code does not communicate directly with Meilisearch.

---

[Rendered](https://github.com/meilisearch/specifications/blob/scoped-api-keys/text/0089-tenant-tokens.md)


curquiza (approved):
Nice spec! 💪 

This should be read by the whole integration team since there is a crucial implementation on their side regarding the user experience 👍

gmourier:
> Nice spec! 💪
> 
> This should be read by the whole integration team since there is a crucial implementation on their side regarding the user experience 👍

Thanks @curquiza! Ping @meilisearch/integration-team ⚡️

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+- Specification PR: [#89](https://github.com/meilisearch/specifications/pull/89)
+- Discovery Issue: [#51](https://github.com/meilisearch/product/issues/51)
+
+# Scoped API Keys
+
+## 1. Functional Specification
+
+### I. Summary
+
+The SDKs can generate `Scoped API Keys` inheriting from a MeiliSearch's `API key` generated to force filters during a search for an end-user. A `Scoped API Key` is generated on the user server-side code to be used by an end-user. It allows users to have multi-tenant indexes and thus restricts access to documents depending on the end-user making the search request.
```
bidoubiwa, line 12: >  generated to force filters during a search for an end-user

This is confusing. What do you mean?

> A `Scoped API Key` is generated on the user server-side code

Do you mean the request to generate is made at the server side? Or is MeiliSearch instance the server side?

> depending on the end-user making the search request

I wouln't use `making the search request` as you may create a scoped API key for users managing indexes and not specifically for search requests.

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+### II. Motivation
+
+`Scoped API Keys` are introduced to solve multi-tenant use-case. By managing this, we reduce one of the last major deal-breakers that makes users not choose MeiliSearch as a solution despite all our advantages.
```
bidoubiwa, line 16:
```suggestion
`Scoped API Keys` are introduced to solve multi-tenant use-cases. By providing the management of API Keys, we remove one of the last major deal-breakers that makes users not choose MeiliSearch as a solution despite all our advantages.
```

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+Users regularly request Multi-Tenant indexes. Users today need to set up workarounds to meet this need. Some of them implement reverse-proxy or managed authentication systems like Hasura or Kong to filter what can and cannot be read. Others decide to use server code as a facade to implement the access restriction logic before requesting MeiliSearch. It is difficult to maintain, less efficient, and requires necessary skills that not everyone has.
+
+### III. Glossary
+
+| Term               | Definition |
+|--------------------|------------|
+| Main Key         | This is the main key that allows the creation of other API keys. The main key is defined by the user when launching MeiliSearch. |
```
bidoubiwa, line 24: Will master key be renamed to main key?

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+| API Key            | API keys are stored and managed from the endpoint `/keys` by the main key holder. These are the keys used by the technical teams to interact with MeiliSearch at the level of the client code. |
```
bidoubiwa, line 25: >  to interact with MeiliSearch at the level of the client code.

What do you mean with this?

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+| Scoped API Key     | These keys are not stored and managed by a MeiliSearch instance. They are generated for each end-user by the backend code from a MeiliSearch API Key. They are used by the end-users to only search the documents belonging to them. |
+| Multi-Tenancy      | By multi-tenancy, we mean that an end-user only accesses data belonging to him within an index shared with other end-users. |
```
bidoubiwa, line 27:
```suggestion
| Multi-Tenancy      |  An instance that only accesses a limited set of data within an index shared with other end-users. |
```

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+- `Scoped API keys` are generated from a MeiliSearch parent `API key` on the client's server code to resolve multi-tenancy use-case by restricting access to data within an index according to the criteria chosen by the team managing a MeiliSearch instance.
+- These `Scoped API keys` can't be less restrictive than a parent `API key` and can only be used for the search action with a predefined forced filter field.
+- These `Scoped API keys` are not stored and thus retrievable on MeiliSearch. This is why we highly advise setting an expiration time on that type of API key for security reasons.
+
+#### Solving Multi-Tenancy with `Scoped API Keys`
+
+![](https://i.imgur.com/J4jVe1n.png)
+
+Let's say that `Mark` is a developer for a SaaS platform. He would like to ensure that an end-user can only access his documents at search time. **His database contains many users and he hopes to have many more in the future.**
```
bidoubiwa, line 48:
```suggestion
Let's say that `Mark` is a developer for a SaaS platform. He would like to ensure that an end-user can only access their documents at search time. **His database contains many users and he hopes to have many more in the future.**
```

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+When a user registers, the backend-side client code generates a `Scoped API Key` specifically for that end-user so he can only access his documents.
```
bidoubiwa, line 50:
```suggestion
When a user registers, the backend-side client code generates a `Scoped API Key` specifically for that end-user so they can only access their documents.
```

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+A `filter` parameter is set to restrict the search for documents having a `user_id` attribute. This `filter` parameter is contained in the `Scoped API key` payload and cannot be modified during the search by the end-user making the request. `filter` can be made of any valid filters. e.g. `user_id = 10 and category = Romantic`
```
bidoubiwa, line 52:
```suggestion
The `filter` parameter is set to restrict the search for documents having an `user_id` attribute. This `filter` parameter is contained in the `Scoped API key` payload and cannot be modified during the search by the end-user making the request. `filter` can be made of any valid filters. e.g. `user_id = 10 and category = Romantic`
```

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+
+This `Scoped API Key` is generated from a parent `API key` used to cipher the `Scoped API Key`. On the MeiliSearch side, this permits checking permissions and authorizing the `UserX` search request using this `Scoped API Key`.
```
bidoubiwa, line 54:
```suggestion
This `Scoped API Key` is generated from a parent `API key` used to cipher the `Scoped API Key`. On the MeiliSearch side, the payload of the `scoped API key` defines what data the user is allowed to retrieve, in order words, MeiliSearch grants authorization based on the permission defined in the Scoped API key payload.
```

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+#### Generating a `Scoped API Key`
+
+```javascript
+const scopedApiKeyRestrictions = {
+     "indexesPolicy": {
+         "products": {
+             "filter": "user_id = 1"
+         },
+         "reviews": {
+             "filter": "user_id = 1 AND published = true"
```
bidoubiwa, line 67: This example is a bit confusing as `products` do not have an associated `user_id` normally. Since we are not a relational database I'm not sure what use cases it represents

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+    //create the secured part (parentApiKey + queryParameters)
+    const securedKey = crypto
+      .createHmac('sha256', parentApiKey)
+      .update(queryParameters)
+      .digest('hex');
+
+    //return the generated `Scoped Api Key`
+    return Buffer.from(prefixKey + securedKey + queryParameters).toString('base64');
+  };
+};
```
bidoubiwa, line 93: Is this expected from the user?

text/0089-scoped-api-keys.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+##### scopedApiKeyRestrictions
+
+The format allows defining specific enforced search filters for accessible indexes (these indexes must be defined in the parent `Api Key` used to generate the `Scoped API key` and have the search action).
+
+If the user does not want to define specific filters for each index accessible to the end-user, he can use the `*` index wildcard rule.
```
bidoubiwa, line 100: Some indexes may not be accessible?

bidoubiwa (changes requested):
Can a user be restricted to access a field?

text/0089-scoped-api-keys.md
gmourier: > generated to force filters during a search for an end-user

Q: This is confusing. What do you mean?
A: It means that a `Scoped API Key` is forged with a specific `filter` parameter that cannot be deleted. They are inserted in the `Scoped API Key` payload. When the client-code emits a search request, the search engine will first validate the authorization of the `Scoped API Key`, extract the Scoped API Key `filter` value and add the search request `filter` parameter behind an `AND` operator.

> A Scoped API Key is generated on the user server-side code

Q: Do you mean the request to generate is made at the server side? Or is MeiliSearch instance the server side?
A: It is generated by calling the SDK method on the server-side user code. MeiliSearch receives nothing and stores nothing when generating a Scoped API Keys when a user generates one for an end-user.

> depending on the end-user making the search request

Q: I wouldn't use making the search request as you may create a scoped API key for users managing indexes and not specifically for search requests.
A: For now, a `Scoped API Key` can only be used for a search request. Using a Scoped API Key to reach an endpoint other than /search will raise an `invalid_api_key` error. We can improve this in the future! 🤓

text/0089-scoped-api-keys.md
gmourier: Absolutely! This is specified in https://github.com/meilisearch/specifications/pull/85.

text/0089-scoped-api-keys.md
gmourier: Client code here means anything that can communicate and interact with MeiliSearch. A curl script, an SDK etc

text/0089-scoped-api-keys.md
gmourier: This is the pseudo-code for an SDK, `meilisearch-js` in this case.

When the user will want to create a `Scoped API Key` for an end-user, his server code will have to call the `generateScopedApiKey` method and give it to the frontend code to authorize the end user's search requests.

text/0089-scoped-api-keys.md
gmourier: The accessible indexes are those defined by the parent `API Key` used (which must have the `search` action permission) to generate the `Scoped API Key`. This format here allows setting a filter on particular fields if there is no common field between documents coming from different accessible indexes or to force some particular `filter` rule for specific indexes.

text/0089-scoped-api-keys.md
gmourier: Imagine that you have 2 indexes: `products` and `reviews` and you only want end-users to retrieve documents containing `user_id = :userId`. You also want to ensure that end-users will only access published reviews when searching documents from the `reviews` index.

bidoubiwa (approved):
A part for the naming part that we already discussed :') the rest is very good to me 🔥

bidoubiwa:
# Naming 

About the naming, I want to make a suggestion to better differ the `api keys` and the `scoped API keys`. 

Today, the problem I have with this is that when we say `api keys` I don't know if we talk about an API key that is able to create `scoped API keys` or if we are talking about a `scoped API keys`. I don't know if we talk about an apiKey that we can access through `/keys` or if it is not stored. 

So I don t think they should be in the name `api key` set. 
For example, given a set of numbers: 

valid_group (1)


Both floats and integers are part of the `number` group. 

But in the case of the API keys we have the following: 


non_valid_api_keys (3)


⚠️ When I say `stored`, in the next img, it means specifically stored inside MeiliSearch and accessible through the `/keys` route. 

This shows why they should not be in the same set. 

While, if we went for another naming we then can create a parent group that differentiate them clearly: 

valid_apikeys (3)

(permission is just a naming that is completely not the naming I specifically want)
 

## Then why tokens or scoped tokens?

It is a common word that describe a key with limited set of accesses. Typically we might think of `jwt`. 
> The tokens are signed either using a private secret or a public/private key.

They are rarely stored in a database, unless in the use-case of `enable-list` and `disable-list`. Which is also possible with the `scoped API keys` of MeiliSearch. 

The word is commonly used to describe the very common scheme: 

```
Authorization: Bearer <token>
``` 

## Conclusion

By naming them differently we can now very easelly what we are mentioning when using the word `token` or `apiKey`. I think it could help users but also in general inside MeiliSearch for smoother comprehension.

irevoire:
Good idea; personally, I know nothing about this domain. I've never deployed anything like that but what I can say for sure is that since the first wg on the subject, I've been struggling with the naming.
Multi-tenant and scoped API keys are not the most explicit terms if you're not already familiar with the subject :unamused:.
I don't know what terms are logical or not, but I totally agree with the fact that having two things named key makes the spec (and the doc later) even harder to remember and understand.
I don't know if it's too late, though :see_no_evil:

alallema:
> # Naming
> About the naming, I want to make a suggestion to better differ the `api keys` and the `scoped API keys`.
> 
> ...
> 
> ## Conclusion
> By naming them differently we can now very easelly what we are mentioning when using the word `token` or `apiKey`. I think it could help users but also in general inside MeiliSearch for smoother comprehension.

I agree that the term `api_key` is not very clear maybe because it has no qualifier. I think it's a good idea to change one of the two or `scoped api key` or `api key`.

gmourier:
I think we can rename the term Scoped API Keys and refocus it on what it does in the first place and what users expect from it aka Multi-Tenancy. (Index having documents for multiple customers (**tenants**)). However I remain quite attached to Scoped API Keys for SEO purposes since it seems to be the terms expected by people looking to solve this problem. The term multi-tenant might be more efficient.

In my opinion, the term API Keys can/should remain as it is because it is equivocal.

Some proposals to replace the term Scoped API Keys:

- `Multi-tenant key`
- `User-defined key`
- `Tenant key`

MarinPostma:
I agree with @bidoubiwa, I'm getting confused with the scoped API key terminology. I like the term `token`, and I think everybody here will know what I am referin to when I talk about token, so something like `auth token`, or just `token` would help me a lot in understanding what we are talking about.

@gmourier, I understand that for SEO reasons you like the term scoped API key, but would that be enough to mention the term when we talk about what we call tokens? Because honnestly we been talking about this for months, and I am still confused :laughing:

gmourier:
I don't think `token` is the best choice because any API key can be considered as an authorization token in essence. I think it is necessary to be more explicit and give a business term that allows understanding what this token is used for.

bidoubiwa:
Yes, going with `token` is a good comprise for me. 
Even without understanding the complexity of multi-tenancy and its use cases, I can at least know that it is a kind of token and not a stored API key. 

All your suggestions with tokens are good for me. I'll give my order of preferences but all are viable choices! 

- Multi-tenant token 4️⃣
- User-defined token  2️⃣
- Scoped token  1️⃣
- Tenant token  3️⃣


> If the main problem is that Scoped does not speak to you, we should keep API Key and find another term for Scoped API Key
👍

dichotommy:
Hello everybody, this is a very interesting discussion happening over here! I want to contribute my perspective, but I have very little expertise in this area so please take all of my comments with a grain of salt.

I agree with Charlotte that these should not be called keys if they do not function by the same rules as our API keys (storage, retrieval, use etc.). We could also apply this same principle to the master key, but that's a subject for another time 😉 

Of all the options presented so far, I prefer "tenant token". Since tenant in everyday English means a person who rents an apartment or resource, it immediately creates a contrast with the "owner". It implies that this is a token created by an owner for a tenant, and has limited privileges. So I like this term a lot on a conceptual level.

> However, I remain quite attached to Scoped API Keys for SEO purposes since it seems to be the terms expected by people looking to solve this problem. The term multi-tenant might be more efficient to gain natural traffic...

It makes sense that we should use the terms "scoped" and "multi-tenant" on the docs and landing page to gain traffic even if that isn't our official title for the feature. If we put these terms aside, I wonder if we could come up with a name that's even more specific to MeiliSearch and what this token can do. For example "search token" pops into my mind, meaning a token that can only be used for search requests.

MarinPostma:
I like the simple and plian `tenant token` @dichotommy proposes

gmourier:
Feel free to propose or vote for other names 😇. For the moment `tenant token` seems to get favorable opinions. Next Monday, a final name will be chosen for the first iteration of this feature.

qdequele:
I'm not really against `tenant token` but it seems to me that the word tenant can't be used alone in computer science. There are single-tenant and multi-tenant, but I don't think I've ever met the word tenant by itself. In our case, these tokens will be used to make multi-tenant, so I would be more in favor of `multi-tenant token`.

gmourier:
> I'm not really against `tenant token` but it seems to me that the word tenant can't be used alone in computer science. There are single-tenant and multi-tenant, but I don't think I've ever met the word tenant by itself. In our case, these tokens will be used to make multi-tenant, so I would be more in favor of `multi-tenant token`.

With the @dichotommy explanation, it makes sense etymologically. A `tenant token` allows having `multi-tenant indexes`. Now I surely agree that if we use the term Tenant Token to describe that very specific end-user keys, I think we should explain the feature in the guides or link the explanations to the term `Multi-Tenant`/`Multi-Tenancy`.

gmourier:
The specification has been updated to use the term `Tenant Token`. Thanks to all for your suggestions and feedback. 🧱

text/0089-tenant-token.md
```diff
@@ -23,7 +23,7 @@ Users regularly request Multi-Tenant indexes. Users today need to set up workaro
 |--------------------|------------|
 | Master Key         | This is the master key that allows the creation of other API keys. The master key is defined by the user when launching MeiliSearch. |
 | API Key            | API keys are stored and managed from the endpoint `/keys` by the master key holder. These are the keys used by the technical teams to interact with MeiliSearch at the level of the frontend/backend code. |
-| Scoped API Key     | These keys are not stored and managed by a MeiliSearch instance. They are generated for each end-user by the backend code from a MeiliSearch API Key. They are used by the end-users to only search the documents belonging to them. |
+| Tenant Token     | These keys are not stored and managed by a MeiliSearch instance. They are generated for each end-user by the backend code from a MeiliSearch API Key. They are used by the end-users to only search the documents belonging to them. |
```
bidoubiwa, line 26:
```suggestion
| Tenant Token     | These tokens are not stored and managed by a MeiliSearch instance. They are generated for each end-user by the backend code from a MeiliSearch API Key. They are used by the end-users to only search the documents belonging to them. |
```

text/0089-tenant-token.md
```diff
@@ -33,32 +33,32 @@ Users regularly request Multi-Tenant indexes. Users today need to set up workaro
...
 
-### V. `Scoped API Key` Explanations
+### V. `Tenant Token` Explanations
 
 #### Summary Key Points
 
-- `Scoped API keys` are generated from a MeiliSearch parent `API key` on the user's backend code to resolve multi-tenancy use-case by restricting access to data within an index according to the criteria chosen by the team managing a MeiliSearch instance.
-- These `Scoped API keys` can't be less restrictive than a parent `API key` and can only be used for the search action with a predefined forced filter field.
-- These `Scoped API keys` are not stored and thus retrievable on MeiliSearch. This is why we highly advise setting an expiration time on that type of API key for security reasons.
+- `Tenant tokens` are generated from a MeiliSearch parent `API key` on the user's backend code to resolve multi-tenancy use-case by restricting access to data within an index according to the criteria chosen by the team managing a MeiliSearch instance.
```
bidoubiwa, line 40:
```suggestion
- `Tenant tokens` are generated from a MeiliSearch parent `API key` on the user's backend code. They are meant to resolve multi-tenancy by restricting access to data within an index according to the criteria chosen by the team managing the MeiliSearch instance.
```

Cut the sentences in chunks for readability

text/0089-tenant-token.md
```diff
@@ -33,32 +33,32 @@ Users regularly request Multi-Tenant indexes. Users today need to set up workaro
...
+- These `Tenant tokens` can't be less restrictive than a parent `API key` and can only be used for the search action with a predefined forced filter field.
```
bidoubiwa, line 41:
```suggestion
- `Tenant tokens` cannot be less restrictive than their respective parent `API key` and can only be used for searching with a predefined forced filter field.
```


consistency with previous bullet point

text/0089-tenant-token.md
```diff
@@ -33,32 +33,32 @@ Users regularly request Multi-Tenant indexes. Users today need to set up workaro
...
+- These `Tenant tokens` are not stored and thus retrievable on MeiliSearch. This is why we highly advise setting an expiration time on that type of API key for security reasons.
```
bidoubiwa, line 42:
```suggestion
- `Tenant tokens` are not stored and thus not retrievable on MeiliSearch. This is why we highly advise setting an expiration time on them for security reasons.
```

text/0089-tenant-token.md
```diff
@@ -33,32 +33,32 @@ Users regularly request Multi-Tenant indexes. Users today need to set up workaro
...
 
-#### Solving Multi-Tenancy with `Scoped API Keys`
+#### Solving Multi-Tenancy with `Tenant tokens`
 
 ![](https://i.imgur.com/J4jVe1n.png)
 
 Let's say that `Mark` is a developer for a SaaS platform. He would like to ensure that an end-user can only access their documents at search time. **His database contains many users and he hopes to have many more in the future.**
 
-When a user registers, the backend code generates a `Scoped API Key` specifically for that end-user so he can only access his documents.
+When a user registers, the backend code generates a `Tenant token` specifically for that end-user so he can only access his documents.
```
bidoubiwa, line 50:
```suggestion
When a user registers, the backend code generates a `Tenant token` specifically for that end-user so they can only access their documents.
```

text/0089-tenant-token.md
```diff
@@ -0,0 +1,125 @@
...
+
+Users regularly request Multi-Tenant indexes. Users today need to set up workarounds to meet this need. Some of them implement reverse-proxy or managed authentication systems like Hasura or Kong to filter what can and cannot be read. Others decide to use server code as a facade to implement the access restriction logic before requesting MeiliSearch. It is difficult to maintain, less efficient, and requires necessary skills that not everyone has.
+
+### III. Glossary
+
+| Term               | Definition |
+|--------------------|------------|
+| Master Key         | This is the master key that allows the creation of other API keys. The master key is defined by the user when launching MeiliSearch. |
+| API Key            | API keys are stored and managed from the endpoint `/keys` by the master key holder. These are the keys used by the technical teams to interact with MeiliSearch at the level of the frontend/backend code. |
+| Tenant Token     | These tokens are not stored and managed by a MeiliSearch instance. They are generated for each end-user by the backend code from a MeiliSearch API Key. They are used by the end-users to only search the documents belonging to them. |
```
alallema, line 26:
```suggestion
| Tenant Token     | These tokens are not stored and managed by a MeiliSearch instance. They are generated for each end-user by the backend code from a MeiliSearch API Key. 
They are used by end-users to search only for documents to which the tenant token gives them access. |
```
Did the documents really belong to the end-users or do they just have access to them? Not sure it's better but it still confusing for me.

text/0089-tenant-token.md
```diff
@@ -0,0 +1,125 @@
...
+    return Buffer.from(prefix + securedKey + queryParameters).toString('base64');
+  };
+};
+```
+
+##### tenantTokenRestrictions
+
+The format allows defining specific enforced search filters for accessible indexes (these indexes must be defined in the parent `Api Key` used to generate the `Tenant Token` and have the search action).
+
+If the user does not want to define specific filters for each index accessible to the end-user, he can use the `*` index wildcard rule.
```
alallema, line 97: Can we add explicitly that this gives access to all indexes and to all documents only in search? If it's the case.

alallema (commented):
Nice spec @gmourier! 🔥
It clarifies a lot of things for me.

text/0089-tenant-token.md
gmourier: Yes, the documents belong to them or in any case should be accessible only to them. By adding the tenant tokens it allows us to overcome two pain-full workarounds to improve the quality of life of a MeiliSearch user. Here is what they have to do to be sure that a document is really accessible to one tenant.

1 - Make an index by tenant (user, company, ...), which can be difficult to manage.
2 - To have a single index by adding a proxy that will inject the forced filter from its backend which forces to have network round trips that will become useless with the tenant token once generated since it will be able to directly exchange with MeiliSearch.

alallema: Thanks for the clarification 😊

text/0089-tenant-token.md
gmourier: It does not give access to every index on the MeiliSearch instance but defines this specific filter for every index accessible for the parent API key used to generate the tenant token.

I'll clarify that!

alallema: Yeah, sorry I thought that `*` gives access to everything.

gmourier: No problem, the specification must be more precise.

alallema (dismissed):
LGTM! 🎉

gmourier:
Hello everyone!

A few design questions, one of which is a major one (the first one) I'd like your opinion on.

1 - Thinking about it, I have the impression that it would be better to use a type of token known by all and easy to integrate for future interoperability purposes. I'm thinking of generating tenant tokens in the form of `JWT`. 

I have a hard time seeing any flaws in using JWT instead of a custom format chosen by us in our SDKs by asking myself "What do we gain by not using JWT to represent a tenant token?"

SDKs do not integrate external libraries, if we choose to use JWT we would certainly need to add a library allowing us to generate them. What is your position on the question @meilisearch/integration-team?

After a discussion with @ManyTheFish, it would be possible to decode a JWT without too many technical complexities on the MeiliSearch side as long as we send the right format.

2 - Regarding the `expiresIn` expressed in seconds, it seems to me to be the better choice than an `expiresAt` with a date time. Since a tenant token is not modifiable over time and generally follows the duration of a user session, `expiresIn` seems to me to be the simplest format for the intrinsic use of a Tenant Token. If we choose `expiresIn`, the SDK method will have to automatically transform it into a precise date or a timestamp representing the end of validity of the token so that MeiliSearch can verify its validity at the end.

`expiresIn` seems to be easier and faster to use in the context of the tenant token but I would like to have your opinion on the question before making a choice.

alallema:
> 

1- I don't like the idea of being dependent on external libraries but anyway we will have to create a token and I think it will be much easier to use a known format with a ready-made library rather than write it, and that should be more acceptable if it is for security reasons.
How it's going to work for users who don't use SDK?

2- For the `expiresIn` I don't have a clear opinion but I always find it complicated to read for a human compared to date. Also, `seconds` seem to be very short and generate big numbers quickly, minutes or hours would not be a good choice? 
And I maybe miss something but why the SDK should transform this field into a date?

MarinPostma:
@gmourier, wait... How exactly is `expireIn` supposed to work? I have the feeling that this is a value that the emitter of the token sets when he creates it, but expire in is a duration, for I am not sure how we are supposed to know when it actually expires if we don't have a referential timestamp. Am I missing something?

gmourier:
Thanks for your answer 😌

@alallema, the tenant token is not expected to be human-readable in most cases. A machine will generate it from code and the end-user will use it transparently from a web app frontend. I thought that `expiresIn` (expressed in seconds) is easier for the developer to code the tenant token generation instead of thinking about a "strict" date-time.

> "How will this work for users who don't use the SDK?" 

That's a good question and in my opinion raises another advantage for using a JWT instead of a custom token format. It seems easier to me to explain a token generation in JWT format for the user (if we decide to do it for those who don't use our SDKs) since it is a widely used format but I could be wrong on that. In short, we would only have to explain the information that needs to be sent to the JWT token generation library rather than explaining in detail how we generate them if we use a custom format.

@MarinPostma The SDK will "virtually" transform the `expiresIn` to an `expiresAt` using e.g. Now() + `expiresIn` value into the JWT / token payload.

bidoubiwa:
Hey!

I agree with the use of JWT. It is a very common standard with great documentation. 

## Question 1: JWT and external libraries

About the libraries, we might have to add in the SDK's. I made some research and I think we do not need any library at all. To explain that, I need to come back on what JWT's contains and how they are created.

Feel free to ignore and just look at the code examples in the different categories. I made them using `javascript` for both `node` and `browser` env.

### JWT's

JWTs are composed of three sections: a `header`, a `payload` (containing the data), and a `signature`. The header and payload are JSON objects, which are serialized to UTF-8 bytes, then encoded using base64url encoding.

The JWT's header, payload, and signature are concatenated with periods (.). As a result, a JWT typically takes the following form:

```
{Base64url encoded header}.{Base64url encoded payload}.{Base64url encoded signature}
```

Example: 
```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
```


### The header 

Conventionally, it contains the type of the token, which is JWT, and the signing algorithm being used, such as HMAC SHA256 or RSA.

Example: 
```js
{
  "alg": "SHA256",
  "typ": "JWT"
}
```

Then the header is encoded in base64 without requiring any external library.
```js
// node 
Buffer.from(header).toString('base64')

// browser
btoa(JSON.stringify(header));
```
( code example is missing serialization in UTF-8 bytes but it does not require any library, it is a simplified example)

### Payload

> The JWT payload contains information about the JWT, such as the target of the token, the issuer, the time the token was issued, and/or the lifetime of the token. Like the JWT header, the JWT payload is a JSON object and is used in the calculation of the signature.

The payload is encoded in base64 as well (see header for code example)

For example, our tokens will contain `expiresIn` and `indexesPolicy`

#### Google example

A payload example, used by the Google API's

The require 3 fields `aud`, `iat`and `exp`

```
{
  "aud": "my-project",
  "iat": 1509654401,
  "exp": 1612893233
}
```

- `iat`:  ("Issued At"): The timestamp when the token was created, specified as seconds...
- `exp`:  ("Expiration"): The timestamp when the token stops being valid, specified as seconds...
- `aud`: ("Audience"): This must be a single string containing the cloud project ID where the device is registered


### Signature


Although JWTs can be encrypted to also provide secrecy between parties, this example uses `signed tokens`. Signed tokens can verify the integrity of the claims contained within it, while encrypted tokens hide those claims from other parties. When tokens are signed using public/private key pairs, **the signature also certifies that only the party holding the private key is the one that signed it** 🔥.

To create the signature part you have to take the encoded header, the encoded payload, a secret, the algorithm specified in the header, and sign that.

For example if you want to use the HMAC SHA256 algorithm, the signature will be created in the following way:

```js
// pseudo code
const signature = HMACSHA256(
  base64UrlEncode(header) + "." +
  base64UrlEncode(payload),
  secret)
```

In the case of our use case, the `secret` is the API key that created the token.

```js
// In node `crypto` is a standard library
require('crypto')
  .createHash('sha256')
  .update(header + payoad + apiKey)
  .digest('hex');

// Browser
crypto.subtle.digest('sha256', header + payoad + apiKey)
```


### Combining

Once we have the header the payload and signature, we combine them the following way:
```
{Base64url encoded header}.{Base64url encoded payload}.{Base64url encoded signature}
```

What is interesting is that the `Header` and the `payload` are not encrypted, we just need to decode them from base64.
But the signature confirms that the JWT was made by the party holding the private key.

for example paste this JWT in the playground here: https://jwt.io/
```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOiIxIiwibmFtZSI6IkJJRE9VQklXQSA-OkQifQ.Lzf7SkWASuA1nIamkpR2D-8OMnq1O6KZGuMLHkH7klw
```

### Validate integrity

At the receiving end, you need to validate the token ⚠️.

Since the signature is hashed with the payload, the header and the API key, if the content changes, the JWT is not valid anymore.
To check this, you need access to the `api key` of the issuer of the token. This also raises the usefulness of the `name` in the APi keys.

For example:

header: 
```json
{
...
}
```

payload: 
```json
{
   "indexesPolicy": "*",
   "expiresIn": 10000,
   "issuer": "custom" 
}
```

to check if the JWT token is valid this is how you would do it :
1. Decode the payload  from base64 and deserialize it
2. Fetch the `issuer`
3. Fetch the API key from the issue in our database
4. recreate the token to check if it is valid with the stored API key

```js

const apiKey = apiKeyFrom(issuer) // custom 

// in case we have no issuer, it will be required to iterate on all API keys to find on that matches
const checkToken = base64(jwt.header) + base64(jwt.payload) + apiKey

if (checkToken !== recievedToken) { 
  throw "Token is not signed correctly"
}
```

gmourier:
Thanks a lot, @bidoubiwa for this excellent detailed answer, it will help us in a crazy way I think.

gmourier:
# WG: Tenant Token Format

## Date: 2022-01-10T15:00:00 GMT+1

### Duration: 45mins

## Attendees

- @bidoubiwa
- @alallema
- @ManyTheFish

## Goals

- Recall some context and explain why we are doing it.
- Choose if we use a JWT format or a custom format for the tenant token.
- Get knowledge on the required payload scheme for MeiliSearch.
- Answer related product / technical questions if any.

---

### Context

Some users would like to have native Multi-Tenant indexes within MeiliSearch. Today they are forced to lose performance and ease of use by going back and forth through a proxy that makes sure to add filters for each user request.

#### Expectations

Following the release of the API Keys management, we aim to release this feature for v0.26.0 to maximize the user impact following the v0.25.0. Moreover, the context remains fresh in the teams on this particular technical subject.

---

### Use a JWT format or a custom format

#### What are the Strengths/Weaknesses of using a JWT format?

| Strengths | 
| -------- | 
| Interoperability (MeiliSearch knows how to parse a JWT format)    | 
| Don't re-invent the wheel with another custom format  | 
| Easier to explain because it is known by a majority | 

| Weaknesses |
|--------------|
| MeiliSearch is dependent on a format |

#### What are the Strengths/Weaknesses of using a custom format?

| Strengths|
| -------- | 
| The size of the token can be reduced because the header part is not needed |
| Not dependent on a format | 

| Weaknesses |
|--------------|
| More difficult to document and explain because it is a custom format |
| Not interoperable |

#### If we choose a JWT format, what are the Strengths/Weaknesses of using an external library on the SDK side?

| Strengths | 
| -------- |
| Less code to maintain |
| Can be more user-friendly depending on some JWT library (e.g. Transform function `expiresIn: 2d` to an`exp` timestamp) |

| Weaknesses |
|--------------|
| Adds an external dependency to the SDKs  |
| JWT libraries might not be equivalent and offer the same set of helper functions |

#### Final Decision

We chose to use a JWT format for its massive use and known by developers. This allows us not to reinvent the wheel with another custom format.

We decided not to implement the external library in our SDKs for this first iteration. A function generating a JWT in the desired format as a helper within the javascript SDK at first as it is the most used SDKs will be provided. Others will follow later.

---

### What is the expected payload format on the MeiliSearch side?

#### Header: Algorithm and token type

At first, MeiliSearch accepts tokens signed with the combination of `HMAC + SHA256`.

e.g.

```json
{
  "alg": "HS256",
  "typ": "JWT"
}
```

#### Payload: Data

MeiliSearch needs information within the token to check its validity and use it to perform requests. 

This information can be separated into two parts, on one hand, the information allows to check the validity of the Tenant Token, on the other hand, the business logic information allows to have a token with pre-defined parameters for the end user's search.

##### Validity related

| Fields   | Description | Comments |
| -------- | -------- | -------- |
| `iss` (Issuer claim)    | Must contain the first 8 characters of the parent `MeiliSearch API key` used to generate the JWT (Tenant Token)  | Is this the most appropriate field? We could decide to generate a custom one since nothing prohibits adding one and claims are optional according to the [JWT RFC-7519](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1). | |
| `exp` (Expiration Time claim) | A JSON numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until the specified UTC date/time, ignoring leap seconds. | This value is optional. However, the SDKs function that generates a JWT specifically expects `null` for the parameter generating `exp` to ensure that the token is explicitly intended without an expiration date. | 

##### Business logic related

| Fields   | Description | Comments |
| -------- | -------- | -------- |
| `indexesPolicy` | This field contains descriptions of the rules/limits applied for search queries performed with the JWT on all or only some of the indexes of the parent API key used to generate the JWT. | The final accepted format for this field will be defined within the specification. The name needs to be reviewed before finalization. |

##### Payload example

Given a MeiliSearch API Key used to generate the JWT from the user backend code. Here is a valid payload data for a JWT.

e.g `MeiliSearch API key: rkDxFUHd02193e120218f72cc51a9db62729fdb4003e271f960d1631b54f3426fa8b2595`

```json
{
    "iss": "rkDxFUHd",
    "exp": 1641835850,
    "indexesPolicy": {
        "*": {
            "filter": "user_id = 1"
        }
    }
}
```
> In this example, `indexesPolicy` allows to specify that no matter which index is searched (among all those accessible by the parent API key that generated the tenant token), this filter will be applied on all search requests.
---

#### Overall JWT generation sample code

```javascript

header = {
  "alg": "HS256",
  "typ": "JWT"
}

base64Header = base64Encode(header)

payload = {
    "iss": "rkDxFUHd",
    "exp": 1641835850,
    "indexesPolicy": {
        "*": {
            "filter": "user_id = 1"
        }
    }
}

base64Payload = base64Encode(payload)

signature = HS256(base64Header + '.' + base64Payload, 'rkDxFUHd02193e120218f72cc51a9db62729fdb4003e271f960d1631b54f3426fa8b2595')

TenantToken = base64Header + '.' + base64Payload + '.' + signature
```
### Additional Q/A 

None, we have exceeded the time limit.

## TLDR;

- MeiliSearch expects Tenant-Token represented as JWT.
- MeiliSearch expects JWTs signed by the HS256 method for this first iteration.
- The `iss` claim will be used to contain the first 8 characters of the MeiliSearch API key used to generate this tenant token.
- The `exp` claim will be used to hold the expiration date of a Tenant Token (JWT) if there is one.
- The SDKs will not use an external library to generate them.
- The Javascript SDK will be the first to provide a helper function to generate JWT to communicate with MeiliSearch.

gmourier:
Hi @meilisearch/docs-team 👋

During our ideation phase, we proposed the use of the term `indexesPolicy` to store the global/specific per index rules to be applied when a search request is issued from a tenant token within it.

The information contained in this `indexesPolicy` field can also be seen as limits/restrictions. For example, in the future, we could use it to allow users to specify a maximum value for the `offset` / `limit` fields at search time.

What do you think of this name? Do you have a better suggestion? Thanks! 😇

text/0089-tenant-token.md
```diff
@@ -0,0 +1,125 @@
...
+
+### III. Glossary
+
+| Term               | Definition |
+|--------------------|------------|
+| Master Key         | This is the master key that allows the creation of other API keys. The master key is defined by the user when launching MeiliSearch. |
+| API Key            | API keys are stored and managed from the endpoint `/keys` by the master key holder. These are the keys used by the technical teams to interact with MeiliSearch at the level of the frontend/backend code. |
+| Tenant Token     | These tokens are not stored and managed by a MeiliSearch instance. They are generated for each end-user by the backend code from a MeiliSearch API Key. They are used by the end-users to only search the documents belonging to them. |
+| Multi-Tenancy      | By multi-tenancy, we mean that an end-user only accesses data belonging to him within an index shared with other end-users. |
```
guimachiavelli, line 27: I'm missing a definition of what we mean by tenant. Is every end-user a tenant?

text/0089-tenant-token.md
```diff
@@ -0,0 +1,125 @@
...
+
+#### Solving Multi-Tenancy with `Tenant tokens`
+
+![](https://i.imgur.com/J4jVe1n.png)
+
+Let's say that `Mark` is a developer for a SaaS platform. He would like to ensure that end-user can only access their documents at search time. **His database contains many users and he hopes to have many more in the future.**
+
+When a user registers, the backend code generates a `Tenant token` specifically for that end-user so they can only access their documents.
+
+The `filter` parameter is set to restrict the search for documents having an `user_id` attribute. This `filter` parameter is contained in the `Tenant token` payload and cannot be modified during the search by the end-user making the request. `filter` can be made of any valid filters. e.g. `user_id = 10 and category = Romantic`
```
guimachiavelli, line 52: I understand the scenario for this spec is to restrict searches so an end-user can only search on documents they have access to, but are there other use-cases for tenant tokens?

text/0089-tenant-token.md
```diff
@@ -0,0 +1,125 @@
...
+
+This `Tenant token` is generated from a parent `API key` used to cipher the `Tenant token`. On the MeiliSearch side, the payload of the `tenant token` defines what data the user is allowed to retrieve, in order words, MeiliSearch grants authorization based on the permission defined in the Tenant token payload.
+
+---
+
+#### Generating a `Tenant token`
```
guimachiavelli, line 58: So, the ideal workflow is not immediately clear to me. I'm assuming the code below is front-end due to it being javascript, but of course correct me if I'm wrong.

After a developer has generated a token from a set of restrictions and an API key, what do they do with? I'm assuming they must send it to MeiliSearch. Is there a new search parameter to do that? Or does the token effectively replace an API key?

Either way, on the back-end, inside an instance: is there a new setting so an index only accepts queries with a tenant token? If not, wouldn't it be fairly easy for someone to bypass restrictions?

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,139 @@
...
+
+### 1.3 `Tenant Token` Explanations
+
+#### 1.3.1 Example: Solving Multi-Tenancy with `Tenant tokens`
+
+![](https://user-images.githubusercontent.com/3692335/149373631-de6f3c5f-a514-4c8d-b018-ee09ccaeaf4d.png)
+
+`Mark` is a developer for a SaaS platform. He would like to ensure that every end-user can only access their documents at search time.
+
+When an end-user registers, the Mark's backend code generates a `Tenant token` for that end-user so they can only access their documents.
```
dichotommy, line 40:
```suggestion
When an end-user registers, Mark's backend code generates a `Tenant token` for that end-user so they can only access their documents.
```

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,139 @@
...
+
+This tenant-token is signed with a MeiliSearch API Key so that MeiliSearch can ensure that the tenant-token has been generated from a known entity.
+
+MeiliSearch check if the Tenant Token is authorized to make the search request.
+
+Then MeiliSearch extract the Tenant Token's rules to apply for the search request.
```
dichotommy, line 46:
```suggestion
MeiliSearch checks if the Tenant Token is authorized to make the search request.

Then MeiliSearch extracts the Tenant Token's rules to apply for the search request.
```

dichotommy:
About the parameter/key name `indexesPolicy`, this is really a tricky one 🤔

As I understand it, `indexesPolicy` alters search results by applying additional parameters to the search request depending on the index that's being searched. If it's empty, that means no additional parameters are added regardless of index.

Based on that definition, I see arguments for and against including the term `index`. On the one hand, it's not affecting indexes directly; on the other, the first level of the object expects index `uid`s as keys. After a bit of brainstorming, here are some two-word favorites:

- `searchModifiers`
- `searchTransform`
- `searchRules`

I think you could add `index` to the front or `byIndex` to the back of any of these. None of them strike me as perfect, though 😕

guimachiavelli:
From @dichotommy's suggestions, I think `searchModifiers` is the best option but am not sure whether I prefer it over `indexesPolicy`.

`searchModifiers` is very explicit about what's going to happen, at least according to the current spec. That's great, but could it be limiting in the future? Is it all possible we might want to allow users to do something more with Tenant Tokens besides modifying searches?

`indexesPolicy` is less explicit and is harder to understand without consulting documentation. Like, I know that a `searchModifier` will modify a search, but I don't know what a policy will do to an index.

There's also a question of what we want to emphasise here. If `indexesPolicy` is only going to be used in the context of Tenant Tokens, perhaps we could make that connection more explicit? For instance:

- `tenantPolicies`
- `tenantRestrictions`
- `tenantSearchModifiers`

Not sure any of these make more sense than what we already have.

text/0089-tenant-token.md
gmourier: I've added a definition in the specifcation. Let me know if it's clearer!

text/0089-tenant-token.md
gmourier: Yes, another possible use-case in the future could be to restrict tokens in terms of rate-limiting for example. That said, the tenants tokens for the moment are generated only for search, it's something we could extend to other actions in the future but in my opinion in the current state of discovery, there is little chance.

text/0089-tenant-token.md
gmourier: Here is the workflow that I hope will help you to understand better:

## User Objective

The user makes a medical SaaS and has an index that contains all medical reports of patients (end-user). He wants to make sure that the end-users will not see the medical reports of other patients.

## Tenant Token to the rescue

1. When the user goes to register or log in, he will have an identifier given by the user's system.
2. This system will generate a token signed by a MeiliSearch API Key (to know at search time that this token has been issued by a "valid" entity (Having knowledge of a MeiliSearch API Key) for this specific end-user. This token will hold within its payload, a description of rules (`indexesPolicy` claim) to be used at search time to ensure that this user can only see his data within a shared index.
3. This token will be given to the frontend code in a transparent way for the end-user.

## Benefits of the solution

- Completely transparent for MeiliSearch. MeiliSearch does not have to store business logic for each end-user and even less to store tokens for each end-user. There could be millions of them.
- The user generates the tenant tokens at will. He does not have to make any specific network exchange with MeiliSearch for that.

gmourier:
@guimachiavelli and @dichotommy Thanks for your help!

Let's keep in mind the use-case requested by our users until today which is to propose a practical way to have multi-tenant indexes at search.

I think the `searchModifiers` or `searchRules` proposals are the best from my point of view. I have a preference for `searchRules` because in the future we could add limits on some of the search parameters per index, for example, `max_offet` or `max_limit`. The term `rule` seems more appropriate than `modifier` in this context from my POV. What do you think about it?

At the JWT root level, we could also have other restrictions related parameters in the future like a rate-limiting or even something like `restrictSources` to ensure that a tenant is only valid if the request is issued from an IP or range of IPs. This is why I am more in favor of something like `searchRules` over `tenantRestrictions`/`tenantPolicies` since this field is specifically related to enforced search parameters/restrictions.

gmourier:
The use of the JWT claim `iss` has been replaced by a custom claim `apiKeyPrefix` in order not to collide with Auth Providers like Okta/Auth0 which use this field to define the identity of the JWT generator. In their case, this field is used to specify the provider url.

ManyTheFish:
Hey @gmourier to replace the `indexesPolicy`name I would suggest:
- `searchScope` (we precise in which index we can search and with a filter or not)
- `searchRight`

I like `searchRules` too. I think we shouldn't choose something like `searchRestriction` because the current implementation is more an `optin` than an `optout`.

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,279 @@
...
+
+> In this example, `indexesPolicy` allows to specify, that no matter which index is searched (among all those accessible by the signing API key that generated the tenant token), this filter will be applied on all search requests.
+
+##### 2.1.2.4 `apiKeyPrefix` field
+
+`apiKeyPrefix` permits to verify that the signing API key of the Token is known and valid within MeiliSearch. It must contain the first 8 characters of the MeiliSearch API key that generates and signs the Tenant Token.
+
+##### 2.1.2.5 `exp` field
+
+`exp` permits to specify the expiration date of the Tenant Token if needed. The format is a JSON numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until the specified UTC date/time, ignoring leap seconds.
```
bidoubiwa, line 114:
```suggestion
`exp` is used to specify the expiration date of the Tenant Token if needed. The format is a JSON numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until the specified UTC date/time, ignoring leap seconds.
```

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+}
+```
+
+> In this example, `indexesPolicy` allows to specify, that no matter which index is searched (among all those accessible by the signing API key that generated the tenant token), this filter will be applied on all search requests.
+
+##### 2.1.2.4 `apiKeyPrefix` field
+
+`apiKeyPrefix` permits to verify that the signing API key of the Token is known and valid within MeiliSearch. It must contain the first 8 characters of the MeiliSearch API key that generates and signs the Tenant Token.
```
bidoubiwa, line 110:
```suggestion
`apiKeyPrefix` contains the 8 first characters of the API key that generated the tenant token. `apiKeyPrefix` is used to find the whole `apiKey` inside the stored API keys in MeiliSearch. Which in turn, validates the signature of the token.
```

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,279 @@
...
+
+##### 2.1.2.5 `exp` field
+
+`exp` permits to specify the expiration date of the Tenant Token if needed. The format is a JSON numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until the specified UTC date/time, ignoring leap seconds.
+
+##### 2.1.2.6 `indexesPolicy` JSON object
+
+`indexesPolicy` is a description of the possible rules for each index.
```
bidoubiwa, line 118:
```suggestion
`indexesPolicy` contains the rules of search on different indexes.
```
`Description` implies more an explanation than a format.

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,279 @@
...
+
+Here are some valid examples in an attempt to cover all possible use cases.
+
+---
+
+> In this case, all searchable indexes from the signing API Key will be searchable by the tenant token without specific rules.
```
bidoubiwa, line 124:
```suggestion
> In this case, all indexes on which the signing API Key has permissions are searchable by the tenant token without any restrictions.
```

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,279 @@
...
+
+```json
+{
+    "indexesPolicy": ["*"] //This notation does not allow the addition of specific rules. The search will just be accessible on all accessibles indexes from the signing API Key for the Tenant Token without specific rules.
+}
+```
+
+---
+
+> In this case, all searchable indexes from the signing API Key will be searchable by the tenant token and MeiliSearch will apply the filter definition before applying the search parameters added by the end user.
```
bidoubiwa, line 155:
```suggestion
> In this case, all indexes on which the signing API Key has permissions are searchable by the tenant token but the additional `filters` limit the search results of the token user. The filters are always applied and overwrite similar filters the end user might add to the search parameters. 
```

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,279 @@
...
+        "*": {
+            "filter": "user_id = 1"
+        }
+    }
+}
+```
+
+---
+
+> In this case, if the medical_records index is searchable from the signing API Key, the tenant token can only search in the medical_records index without applying specific rules before applying the search parameters added by the end user.
```
bidoubiwa, line 169:
```suggestion
> In this case, if the `medical_records` index is searchable from the signing API Key, the tenant token can only search in the `medical_records` index. No further rules impact their search results on `medical_records`.
```

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,279 @@
...
+
+```json
+{
+    "indexesPolicy": ["medical_records"]
+}
+```
+
+---
+
+> In this case, if the medical_records index is searchable from the signing API Key, the tenant token can only search in the medical_records index and specific rules will be applied at search time before applying the search parameters added by the end user.
```
bidoubiwa, line 198:
```suggestion
> In this case, if the medical_records index is searchable from the signing API Key, the tenant token can only search in the medical_records index, plus specific rules limits their search results. These rules are always applied and cannot be overwritten by the users search parameters.
```

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,279 @@
...
+        "medical_records": {
+            "filter": "user_id = 1"
+        }
+    }
+}
+```
+
+---
+
+> In this case, if the medical_records and medical_appointments indexes are searchable from the signing API Key, the tenant token can only search in those indexes and will apply specific rules given the searched index before applying the search parameters added by the end user.
```
bidoubiwa, line 212:
```suggestion
> In this case, if the medical_records and medical_appointments indexes are searchable from the signing API Key, the tenant token can only search in those indexes and specific rules limits their search results. These rules are always applied and cannot be overwritten by the users search parameters.
```

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,279 @@
...
+        "medical_appointments": {
+            "filter": "user_id = 1 AND accepted = true"
+        }
+    }
+}
+```
+
+---
+
+> In this case, all searchable indexes from the signing API Key will be searchable and the rules under `*` will be applied at search time. The medical_appointments index policy will replace the `*` rules specifically when this index is searched by those specified for that index.
```
bidoubiwa, line 229:
```suggestion
> In this case, all searchable indexes from the signing API Key are searchable and have their search apply specific rules defined in the wildcard field: `*`. The `medical_appointments` index rules, defined in the field of the same name, overwrites the rules defined in the wildcard field`*`.
```

gmourier:
@meilisearch/docs-team, the specification is near to being in a final version except for corner-cases that could be seen during the tests of the integration team but our first tests with @alallema are rather satisfactory.

Decided in comittee @bidoubiwa, @alallema and @ManyTheFish today:

- `indexesPolicy` is renamed `searchRules`, thanks to all your feedback ! 
- MeiliSearch accepts JWT signed with the following algorithms: `HS256`, `HS384`, `HS512`, `RS256`, `RS384`, `RS512`, `PS256`, `PS384`, `PS512`, `ES256`, `ES384`.

ManyTheFish (changes requested):

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,298 @@
...
+
+#### 1.1.1 Summary Key Points
+
+- `Tenant tokens` are JWTs generated on the user side. Thus not stored or retrievable on MeiliSearch side.
+- `Tenant tokens` contain rules that ensure that a `Tenant token` holder (e.g an end-user) only has access documents matching rules chosen at the `tenant token` creation.
+- `Tenant tokens` are signed from a MeiliSearch `API key` on the user's code.
+- `Tenant tokens` cannot be less restrictive than the signing `API key` and can only be used for searching. A Tenant Token cannot search within more indexes than the API Key that signed that Tenant Token.
+- `Tenant tokens` can have different rules for each index accessible by the signing API key. These filters rule per index are described in an `searchRules` json object.
+- The only rule at the moment is the search parameter `filter`. Other rules may be added in the future.
+- When a request is made with the Tenant Token, MeiliSearch checks if this Tenant Token is authorized to make the request and in this case injects the rules at search time.
```
brunoocasali, line 24: Can we make this more explicit @gmourier? Because reading that now, makes me think in two possibilities:

- the SDK will decode the JWT and then get the `searchRules` to send them into the search request
- the Meilisearch server will receive the tenant token and it should be responsible to decode + apply these search rules on the fly.

gmourier: I've added some precisions. Let me know if it's enough. 😇

text/0089-tenant-tokens.md
```diff
@@ -0,0 +1,299 @@
...
+}
+```
+
+---
+
+> In this case, if the `medical_records` and `medical_appointments` indexes are searchable from the signing API Key, the tenant token can only search in those indexes. MeiliSearch will apply the filter rule before applying the search parameters added by the end user.
+
+```json
+{
+    "searchRules": {
```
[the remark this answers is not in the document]
MarinPostma: Also, it could be good idea to validate a token, by performing a placeholder search with it.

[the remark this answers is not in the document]
gmourier: Some users have asked for the possibility to hide the `tenant` field that can be used in a `searchRules, which will probably be implemented in the future. With this type of error, an end-user could have the information of the field that allowed the multi-tenancy when it could have been explicitly hidden via a `hideAttributes` attribute.

The tenant tokens will be generated programmatically in the vast majority of cases. It might be useful to include the TenantToken source in the error message to make it clear that the error comes from `searchRules` format and thus facilitate error resolution, but I wonder how this would work for an end-user who is reporting the bug on a field that needs to be `hidden`.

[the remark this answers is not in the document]
brunoocasali: Hey @gmourier I understood your suggestion, and I think it is not the best since we will be creating a connection between two different API's and like you said, blocking the future possibilities of this, but I understood that maybe not possible to add the source to the error message either.

I just wanted to raise the concern with you. What do you think to add some sentence in the `Future possibilities` regarding this topic? 
Talking to @curquiza she told me that is best to add a new [discussion in the product](https://github.com/meilisearch/product/discussions/398) in order to keep track of this.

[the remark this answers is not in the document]
gmourier: Thanks Bruno, I will add it to the Future Possibilities!

gmourier:
Light reorganization of the specification and fix of some typos.

I removed the pseudo-JS code because it doesn't bring much to the specification. SDKs and documentation are much better for this since they are presenting working code and I want to avoid managing 2 base codes.

I need 2 approval reviews to merge this specification. If one person from the @meilisearch/integration-team and one person from the @meilisearch/docs-team can take a look at it, it will be really appreciated 🙏

brunoocasali (approved):
Good job @gmourier and team! 🎉 🌮

curquiza (approved):
💪

PR #90, opened 2021-11-15
Add `details` payload by task type and index deletion behavior on task retrieval

gmourier:
## Task Details definition

We have the possibility to add more information than initially planned. We propose to add essentially a diff detail for tasks that are supposed to receive documents and finally index some or all of them. `documentsAddition` / `documentsPartial` / `documentsDeletion`.

We also specify which `primaryKey` is sent for index creation and update if it is filled in.

For `settingsUpdate` task type, the `details` field does not change.

## Deleting an index

Deleting an index means that tasks for that index can no longer be found via `indexes/:indexName/tasks`. However, they are still accessible via `/tasks` for this version.

### DIFF
--- a/text/0060-refashion-updates-apis.md
+++ b/text/0060-refashion-updates-apis.md
@@ -30,7 +30,7 @@ The specification makes any writing operation on an index asynchronous to serve 
     - Attributes of an error appearing in a `failed` `task` are now contained in a dedicated `error` object.
     - `type` is no longer an object. It now becomes a string containing the values of its `name` field previously defined in the `type` object.
     - The possible values for the `type` field are reworked to be more clear and consistent with our naming rules.
-    - A `details` object is added to contain specific information related to a `task` payload that was previously displayed in the `type` nested object. Previous `number` key is renamed `numberOfDocuments`.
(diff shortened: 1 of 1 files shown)


text/0060-refashion-updates-apis.md
```diff
@@ -112,7 +112,70 @@ The motivation is to stabilize the current `update` resource to a version that c
...
+
+| name     | description |
+| -------- | --------    |
+| primaryKey | Value for the `primaryKey` field into the PUT index payload. |
+
+##### indexDeletion
+
+| name     | description |
+| -------- | --------    |
+| deletedDocuments    | Number of deleted documents. Should be all documents contained in the deleted index. |
```
MarinPostma, line 157: Is this one really necessary? We don't interact with the index when deleting it, we simply remove it from fs, so this information require that we open and read from the index before we delete it. It's possible to do it, just making sure that this is necessary.

gmourier: It seems interesting to me to give this information to have consistency on the information returned when documents are deleted. Do you see any potential problems or technical constraints to open and read the index before cleaning it of these documents?

MarinPostma: not really

MarinPostma (approved):

PR #91, opened 2021-11-15
Fix documentation of search `sort` query param

paulRbr:
It seems there's a missing empty line to differentiate two call-outs in the documentation of the `sort` query parameter of the `GET` search route.

Strangely enough, the `POST` search route doesn't have this `info` call-out maybe it's an oversight?

# Pull Request

## What does this PR do?

Fixes a documentation visual defect in the OpenAPI contract [here](https://bump.sh/doc/meilisearch#get-indexes-parameter-search-sort).

### DIFF
--- a/open-api.yaml
+++ b/open-api.yaml
@@ -189,6 +189,9 @@ components:
 
         > warn
         > Attribute(s) used in `sort` should be declared as sortable attributes. See [Sorting](https://docs.meilisearch.com/reference/features/sorting.html).
+
+        > info
+        > _geoPoint({lat}, {long}) built-in sort rule can be used to sort documents around a geo point.
       example:
         - 'price:desc'
       title: sort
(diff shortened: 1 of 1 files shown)


gmourier:
Thanks a lot @paulRbr, it's super appreciated!

paulRbr:
no problem @gmourier :slightly_smiling_face:.



> Indeed it is an oversight for the POST route. There should also be this call-out!

I've pushed a modification to include the `info` call-out in the `POST` route too.

gmourier:
Thanks @paulRbr! I'm gonna merge it to develop!

gmourier:
Interestingly the CI bump diff didn't succeed on this pull-request due to "Error: You are not allowed to deploy to this documentation."

Can it be because it comes from a fork?

paulRbr:
> Can it be because it comes from a fork?

Yes that's exactly why. For now the `diff` feature of Bump is only
available when used with a Bump account (in this case your secret is
only available on the `meilisearch/specifications` repository and not
on forks).

We're working on making the `diff` feature available for anybody (not
connected users) which will allow those kind of use-cases (CI on
forks) to work. 🙂


Thanks for your review and merging the fix!

PR #94, opened 2021-11-29
Add technical detail about the first Launched event and the segment identify call

gmourier:
The segment identify call is only sent after the first hour if the instance is still running. At the first launch, MeiliSearch sends a `Launched` event on an instance ID equal to `total_launch` in order to avoid tracking instances usage for nothing when they could be shut down and never restarted.


irevoire (approved):
Approved and implemented :+1:

PR #96, opened 2021-12-01
Auto-Batching

gmourier:
# Summary

Meilisearch can automatically group consecutive asynchronous `documentAddition` or `documentPartial` tasks for the same index into a single batch via an automatic batching mechanism.

This auto-batching behavior is enabled by the user through various command flag options.

# Motivation

We have been collecting user pain points regularly over the last year pointing out the slow indexing.

We have explained several times to users to make batches containing a maximum of documents to be updated/added in order to compress the indexing time of certain data structures.

In order to make MeiliSearch easier to use, we explored the idea of automatically creating these batches within MeiliSearch before indexing users’ documents.

## Key points

- Auto-batching mechanisms are disabled by default.
- A `batchUid` field is added on a fully qualified `task` API object.
- If auto-batching is disabled, every batch is made of `1` task.
- To have access to the commands allowing to configure auto-batching, the flag `--enable-auto-batching` must be passed to MeiliSearch.
- `--max-batch-size`: Set the maximum number of tasks that can be processed in a batch. If not specified, this is unlimited.
- `--max-documents-per-batch`: Set a limit to the maximum number of documents that can be indexed together in a single batch. Since the batch must contain at least one update, this value can be exceeded. If not specified, this is unlimited.
- `--debounce-duration-sec`: Wait at least `SECS` seconds between the time the scheduler is notified of a new `task` and the processing of the related batch. Defaults to 0secs (process immediately).
- For now, only the `documentAddition` and `documentPartial` tasks type consecutive and collocated within the same index can be auto-batched.


text/0096-auto-batching.md
```diff
@@ -0,0 +1,67 @@
...
+
+Only consecutive `documentsAddition` and `documentsPartial` `tasks` for the same index can have a shared `batchUid` since the scheduler will be able to group them together. This means that all `tasks` concerning other operations will also be part of a `batchUid` having only one task. This could be extended in the future by enhancing the scheduler capabilities.
+
+This is to avoid having multiple incompressible computation time by grouping them together to increase performance when applied to a maximum number of documents.
+
+#### 1.3.2 Grouping tasks to a single batch
+
+The scheduling program that groups tasks within a single batch is triggered when an asynchronous `task` currently processed reaches a terminal state as `succeeded` or `failed`.
+
+In other words, when the next `task` should be picked from the FIFO task queue. The scheduler fetch and group all consecutive `documentsAddition` for a similar index in a batch until it encounters another task type or a similar task type but for a different index.
```
qdequele, line 37: Is it really more complex to continue to unpack events if you encounter a task from a different Index?

```suggestion
In other words, when the next `task` should be picked from the FIFO task queue. The scheduler fetch and group all consecutive `documentsAddition` for a similar index in a batch until it encounters another task type.
```

E.g.
Here is a task list (t: taskId, i: index, a: action (update or delete), b: batchUid)
```
last_task_id = 50;
last_batch_uid = 4;

t:51 i:movies 	a: update 	b: 5
t:52 i:movies 	a: update 	b: 5
t:53 i:shows 	a: delete 	b: ?
t:54 i:shows 	a: update 	b: ?
t:55 i:movies 	a: update 	b: 5
t:56 i:movies 	a: delete 	b: ?
t:57 i:shows 	a: update 	b: ?
t:58 i:shows 	a: update 	b: ?
t:59 i:movies 	a: update 	b: 5
t:60 i:movies 	a: delete 	b: ?

-- Next tick
last_task_id = 60;
last_batch_uid = 5;

t:53 i:shows 	a: delete 	b: 6
t:54 i:shows 	a: update 	b: ?
t:56 i:movies 	a: delete 	b: ?
t:57 i:shows 	a: update 	b: ?
t:58 i:shows 	a: update 	b: ?
t:60 i:movies 	a: delete 	b: ?
t:61 i:movies 	a: delete	b: ?

-- Next tick
last_task_id = 61;
last_batch_uid = 6;

t:54 i:shows 	a: update 	b: 7
t:56 i:movies 	a: delete 	b: ?
t:57 i:shows 	a: update 	b: 7
t:58 i:shows 	a: update 	b: 7
t:60 i:movies 	a: delete 	b: ?
t:61 i:movies 	a: delete	b: ?

-- Next tick
last_task_id = 61;
last_batch_uid = 7;

t:56 i:movies 	a: delete 	b: 8
t:60 i:movies 	a: delete 	b: 8
t:61 i:movies 	a: delete	b: 8

done.
```

qdequele: I don't even know if this is a good idea.

gmourier: Thanks, @qdequele  The goal is mainly to improve the process during the first massive document indexing that is not batched the right way by the user.

This is something we might want to do for the scheduler but it implies having a more intelligent and complex system. 

I'm thinking about compacting updates before adding documents for example. If the scheduler is not more intelligent, it seems to me that we end up with documents that are less up-to-date than they should be.

@MarinPostma your opinion will surely be much more enlightened than mine on this subject!

qdequele: Understood. Thanks @gmourier :)

MarinPostma: @qdequele We need to guarantee that updates are processed in the order the user provided them for an index, or at least that the the order we process them is equivalent. In you example, you process `[51, 52, 55, 59]` in the same batch, but `59 `happens after `56` which is a delete operation, and there is no reason to believe that processing 59 before 56 would yield the same result as the opposite order. e.g: `56` could delete docid 1, and `59` insert docid 1, it is easy to see that the order wuld matter in that case

qdequele: Indeed I made a mistake in my example that I corrected. :/

qdequele: I think we can solve this conversation. I agree that we can go to the simplest for the first iteration since the future evolutions will change the behavior through optimizations but not the API.

MarinPostma:
I am wondering if it would not be a good idea for auto-batching to be an opt-in until we have a more permissive error system. I don't think that the current "fail all" is acceptable in most usecases.

text/0096-auto-batching.md
```diff
@@ -0,0 +1,67 @@
...
+![Auto-batching Process](https://user-images.githubusercontent.com/3692335/145787054-4cb07b5e-c80e-498a-8843-d0cc46329e9b.png)
+
+##### 1.3.2.2 `batchUid` generation
+
+The identifier chosen for the `batchUid`field corresponds to the `uid` value of the first task grouped within a batch. The batch identifiers are therefore unique and consecutive.
+
+#### 1.3.3 Impacts on `task` API object format
+
+- The different tasks grouped in a batch will be processed within the same transaction. That is to say that if a task fails within a batch, all the tasks fail or succeed.
+- A `batchUid` field is added on fully-qualified `task` API objects.
```
curquiza, line 54: Maybe precise this is only when calling `/task` route and not the response of an asynchronous task (like `POST /indexes`). This is guessable when reading the whole spec, but it can help the docs-team during.

curquiza (commented):
👍

text/0096-auto-batching.md
MarinPostma: for me this is what he means when he says `fully-qualified` tasks object, as opposed to the `summarized` task objects that are returned on update creation

gmourier:
> I am wondering if it would not be a good idea for auto-batching to be an opt-in until we have a more permissive error system. I don't think that the current "fail all" is acceptable in most usecases.

@MarinPostma It is a good idea to leave this option to the user. Could we also go further and let him override a maximum number of documents for a batch?

MarinPostma:
I have implemented that already: max number of tasks per batch and max number of documents per batch.

text/0096-auto-batching.md
```diff
@@ -0,0 +1,67 @@
...
+
+This is to avoid having multiple incompressible computation time by grouping them together to increase performance when applied to a maximum number of documents.
+
+#### 1.3.2 Grouping tasks to a single batch
+
+The scheduling program that groups tasks within a single batch is triggered when an asynchronous `task` currently processed reaches a terminal state as `succeeded` or `failed`.
+
+In other words, when the next `task` should be picked from the FIFO task queue. The scheduler fetches and groups all consecutive `documentAddition` for a similar index in a batch until it encounters another task type or a similar task type but for a different index.
+
+> Note that we are considering implementing a configurable limit of maximum documents to process for a batch and a flag to activate the auto-batching mechanism.
```
guimachiavelli, line 39: So basically a way of telling an instance "batch `documentAddition` up to X amount of documents" so we avoid memory consumption problems?

text/0096-auto-batching.md
```diff
@@ -0,0 +1,67 @@
...
+
+The more similar consecutive tasks the user sends in a row, the more likely the batching mechanism will be able to group these tasks in a batch. It can be seen as an automatic back-pressure mechanism.
+
+##### 1.3.2.1 Schema
+
+![Auto-batching Process](https://user-images.githubusercontent.com/3692335/145787054-4cb07b5e-c80e-498a-8843-d0cc46329e9b.png)
+
+##### 1.3.2.2 `batchUid` generation
+
+The identifier chosen for the `batchUid` field corresponds to the `uid` value of the first task grouped within a batch. The batch identifiers are therefore unique and consecutive.
```
guimachiavelli, line 49: Meaning that if there are 5 consecutive `documentAddition` tasks and the first of these has a `uid` of `142`, the `batchUid` for these 5 tasks will be `142`, correct?

Is this behaviour the same in tasks other than `documentAddition` and `documentPartial`? As in, a task of type `settingsUpdate` with an `uid` of `211` will have a `batchUid` of `211`?

gmourier: That's it @guimachiavelli

text/0096-auto-batching.md
```diff
@@ -0,0 +1,86 @@
+- Title: Auto-Batching
+
+# Auto-Batching
+
+## 1. Summary
+
+Meilisearch can automatically group consecutive asynchronous `documentsAddition` or `documentsPartial` tasks for the same index via an automatic batching mechanism.
```
curquiza, line 7:
```suggestion
Meilisearch can automatically group consecutive asynchronous `documentAddition` or `documentPartial` tasks for the same index via an automatic batching mechanism.
```

curquiza (approved):
💪

gmourier:
@MarinPostma Can we rename `--enable-autobatching` to --enable-auto-batching` ? 

@meilisearch/docs-team 📡

gmourier:
@meilisearch/docs-team 📡  `--enable-autobatching` is renamed --enable-auto-batching` 🥷! Thanks @MarinPostma

curquiza (approved):

text/0096-auto-batching.md
```diff
@@ -0,0 +1,86 @@
...
+
+All `tasks` are linked to a `batchUid` field. The `batchUid` field identifies that several identical, consecutive tasks have been grouped in the same batch.
+
+Only consecutive `documentAddition` and `documentPartial` tasks for the same index can have the same `batchUid`. All `tasks` concerning other operations will also be part of a `batchUid` having only one task.
+
+#### 3.1.1. Grouping tasks to a single batch
+
+The scheduling program that groups tasks within a single batch is triggered when an asynchronous `task` currently processed reaches a terminal state as `succeeded` or `failed`.
+
+In other words, when a scheduled `task` is picked from the FIFO task queue, the scheduler fetches and groups all consecutive `documentAddition` for a similar index in a batch until it encounters another task type or a similar task type but for a different index.
```
MarinPostma, line 29: Actually, it doesn't stop on the first update of another index. It takes as many updates as possible for an index, possibly skipping over updates for other indexes to take more.

text/0096-auto-batching.md
```diff
@@ -0,0 +1,86 @@
...
+
+The more similar consecutive tasks the user sends in a row, the more likely the batching mechanism can group these tasks.
+
+##### 3.1.1.1. Schema
+
+![Auto-batching Process](https://user-images.githubusercontent.com/3692335/145787054-4cb07b5e-c80e-498a-8843-d0cc46329e9b.png)
+
+##### 3.1.1.2. `batchUid` generation
+
+The identifier chosen for the `task` `batchUid` field corresponds to the `uid` value of the first task grouped within a batch. The batch identifiers are therefore unique and consecutive.
```
MarinPostma, line 39: The fact that it is the id of the first update in the batch should not be relied upon, and should not be part of the specification IMO. It is irrelevant for the user, and we should keep the opportunity to change that in the future without breaking it.

text/0096-auto-batching.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### 3.1.2. Impacts on `task` API resource
+
+- The different tasks grouped in a batch are processed within the same transaction. If a task fails within a batch, all the tasks fail or succeed.
```
MarinPostma, line 43: Just to clarify that it is not one or the other, if one task fails, the batch fails, I know this is what you meant :)
```suggestion
- The different tasks grouped in a batch are processed within the same transaction. If a task fails within a batch, the whole batch fails.
```

text/0096-auto-batching.md
```diff
@@ -0,0 +1,86 @@
...
+- A `batchUid` field is only added on fully-qualified `task` API objects. It corresponds to the `task` `uid` value of the first task grouped within a batch. `batchUid` values are therefore unique and consecutive.
```
MarinPostma, line 44: cf my previous comment about it being the first id

text/0096-auto-batching.md
```diff
@@ -0,0 +1,86 @@
...
+
+`--max-documents-per-batch <NUM>` allows setting a limit to the maximum number `NUM` of documents that can be indexed together within a single batch.
+
+Since the batch must contain at least one update, this value can be exceeded.
+
+If not specified, this is unlimited.
+
+### 3.2.4. `--debounce-duration-sec`
+
+`--debounce-duration-sec <SECS>` allow to wait at least `SECS` seconds before processing a scheduled batch.
```
MarinPostma, line 74:
```suggestion
`--debounce-duration-sec <SECS>` allows waiting for `SECS` after receiving an update before attempting to make a batch and processing it.
```

MarinPostma (changes requested):
Sorry, I am a bit late for the party.

I just gave some more thought to the debounce duration, and it appears that it would also block jobs from being performed, I think we should remove it for now.

text/0096-auto-batching.md
gmourier: Was this added late? I am pretty sure that the original behavior was as described in the specification 🤯

I'll describe it.

The review is late though. It's probably going to be a last-minute job for the documentation team and dev rel team that is already well advanced for the v026.0 release. 😟

text/0096-auto-batching.md
gmourier: I don't see too much problem describing it, the feature is experimental and we warn that we can break everything on this one! 

It was also to explain the behavior of `batchUid` for internal teams. It touches the API of the tasks and it is, therefore visible in the user's eyes.

text/0096-auto-batching.md
gmourier: The proposed phrasing seems difficult to understand with "before" and "after". What do you want to highlight @MarinPostma? By reading your sentence, it seems to me that the debounce is restarted with each new update received.

gmourier:
Signal @meilisearch/docs-team 📡  Not sure if it impacts the doc directly but you mention 'immediately consecutive tasks`, that's why I am pointing it out to you. 

We made some clarifications after a late technical review. Sorry for this.

The scheduler is able to dodge tasks from another index when preparing a batch.

When the scheduler is notified with a task queue like that:

- task `10` (`**index: 1**` - `documentAddition`)
- task `9` (`**index: 2**` - `documentAddition`)
- task `8` (`**index: 1**` - `documentAddition`) 

It will create a batch with tasks `10` and `8`.

The creation of the batch ends if, for the same index, it meets a task of a different type.

text/0096-auto-batching.md
```diff
@@ -0,0 +1,88 @@
...
+![Auto-batching Process](https://user-images.githubusercontent.com/3692335/145787054-4cb07b5e-c80e-498a-8843-d0cc46329e9b.png)
+
+##### 3.1.1.2. `batchUid` generation
+
+The identifier chosen for the `task` `batchUid` field corresponds to the `uid` value of the first task grouped within a batch. The batch identifiers are unique and ascending.
+
+#### 3.1.2. Impacts on `task` API resource
+
+- The different tasks grouped in a batch are processed within the same transaction. If a task fails within a batch, the whole batch fails.
+- A `batchUid` field is only added on fully-qualified `task` API objects. It corresponds to the `task` `uid` value of the first task grouped within a batch. `batchUid` values are unique and in ascending order.
```
MarinPostma, line 44:
```suggestion
- A `batchUid` field is only added on fully-qualified `task` API objects. `batchUid` values are unique and strictly increasing.
```

text/0096-auto-batching.md
```diff
@@ -0,0 +1,88 @@
...
+
+The more similar consecutive tasks the user sends in a row, the more likely the batching mechanism can group these tasks.
+
+##### 3.1.1.1. Schema
+
+![Auto-batching Process](https://user-images.githubusercontent.com/3692335/145787054-4cb07b5e-c80e-498a-8843-d0cc46329e9b.png)
+
+##### 3.1.1.2. `batchUid` generation
+
+The identifier chosen for the `task` `batchUid` field corresponds to the `uid` value of the first task grouped within a batch. The batch identifiers are unique and ascending.
```
MarinPostma, line 39:
```suggestion
The batch identifiers are unique and strictly increasing.
```

text/0096-auto-batching.md
```diff
@@ -0,0 +1,88 @@
...
+
+All `tasks` are linked to a `batchUid` field. The `batchUid` field identifies that several identical, consecutive tasks have been grouped in the same batch.
+
+Only consecutive `documentAddition` and `documentPartial` tasks for the same index can have the same `batchUid`. All `tasks` concerning other operations will also be part of a `batchUid` having only one task.
+
+#### 3.1.1. Grouping tasks to a single batch
+
+The scheduling program that groups tasks within a single batch is triggered when an asynchronous `task` currently processed reaches a terminal state as `succeeded` or `failed`.
+
+In other words, when a scheduled `documentAddition` task is picked from the task queue, the scheduler fetches and groups all `documentAddition` tasks for a similar index in a batch until it encounters another task type for that index.
```
MarinPostma, line 29:
```suggestion
In other words, when a scheduled `documentAddition` task for a given index is picked from the task queue, the scheduler fetches and groups all `documentAddition` tasks for that same index in a batch until it encounters another task type for that index.
```

text/0096-auto-batching.md
```diff
@@ -0,0 +1,88 @@
...
+
+We have regularly collected user pain points pointing out the slow indexing over the last year. We explained several times to users to make batches containing a maximum of documents to be updated/added to compress the indexing time of specific data structures.
+
+To make Meilisearch easier to use, we explored the idea of automatically creating these batches within Meilisearch before indexing users’ documents.
+
+## 3. Functional  Specification
+
+### 3.1. Explanations
+
+All `tasks` are linked to a `batchUid` field. The `batchUid` field identifies that several identical, consecutive tasks have been grouped in the same batch.
```
MarinPostma, line 21:
```suggestion
All `tasks` are part of a batch identified by a `batchUid`. A task batch preserves the logical order of the tasks for a given index.
```

MarinPostma (changes requested):

MarinPostma (dismissed):

text/0096-auto-batching.md
```diff
@@ -0,0 +1,88 @@
...
+Meilisearch can automatically group consecutive asynchronous `documentAddition` or `documentPartial` tasks for the same index via an automatic batching mechanism.
+
+The user can enable this auto-batching behavior through various command flag options.
+
+## 2. Motivation
+
+We have regularly collected user pain points pointing out the slow indexing over the last year. We explained several times to users to make batches containing a maximum of documents to be updated/added to compress the indexing time of specific data structures.
+
+To make Meilisearch easier to use, we explored the idea of automatically creating these batches within Meilisearch before indexing users’ documents.
+
```
irevoire, line 16: There is another motivation actually, the feature won't only help the unexperimented users.

Currently, users can't really make batches bigger than 100mo without changing their http-conf, with the auto-batching you can actually index waaay more documents than before.

text/0096-auto-batching.md
```diff
@@ -0,0 +1,88 @@
...
+If not specified, this is unlimited.
+
+### 3.2.3. `--max-documents-per-batch`
+
+`--max-documents-per-batch <NUM>` allows setting a limit to the maximum number `NUM` of documents that can be indexed together within a single batch.
+
+Since the batch must contain at least one update, this value can be exceeded.
+
+If not specified, this is unlimited.
+
```
irevoire, line 71:
```suggestion
`--max-documents-per-batch <NUM>` allows setting a limit to the maximum number `NUM` of documents that can be indexed together within a single batch.

Since the batch can't split one update in half, this value is rounded up to the number of documents in the last document addition.

If not specified, this is unlimited.
```

From what I understood, we're actually going to exceed this value every time; it's not really related to the minimum number of doc addition we need to put in a batch, so I tried to rephrase this.
Not sure if my sentence is clearer though :unamused:

text/0096-auto-batching.md
```diff
@@ -0,0 +1,88 @@
...
+## 4. Technical Aspects
+N/A
+
+## 5. Future Possibilities
+
+- Extends it for all consecutive payload types.
+- Add a filter capability by `batchUid` on the `/tasks` endpoints.
+- Do not fail the entire transaction if a document is not valid. Report the documents that could not be indexed to the user.
+- Enable auto-batching by default.
```
irevoire, line 88:
```suggestion
- Extends it for all consecutive payload types.
- Add a filter capability by `batchUid` on the `/tasks` endpoints.
- Do not fail the entire transaction if a document is not valid. Report the documents that could not be indexed to the user.
- Enable auto-batching by default.
- Optimize some tasks sequence, for example if there is a document addition followed by an index deletion, we could skip the document addition
```

irevoire (changes requested):

text/0096-auto-batching.md
gmourier: It's much clearer! Thanks @irevoire!

irevoire (approved):
Good to me :+1:

PR #98, opened 2021-12-08
Accepts 0 documents payload on `POST`/`PUT`- `/documents`

gmourier:
# Description

Following feedback on https://github.com/meilisearch/product/discussions/319. 

MeiliSearch does not return an error to guide the user anymore when a payload with 0 documents is sent for the routes `POST`- `PUT` `/documents`.

This is to facilitate the creation of an empty index and the use of message queues sending empty payloads to be indexed.

### DIFF
--- a/text/0061-error-format-and-definitions.md
+++ b/text/0061-error-format-and-definitions.md
@@ -875,19 +875,6 @@ HTTP Code: `400 Bad Request`
 - The `:payloadType` is inferred when the message is generated. e.g. `json`, `ndjson`, `csv`
 - The `:syntaxErrorHelper` is inferred when the message is generated.
 
-#### Variant: Sending a valid but empty `[]`/`{}` payload on `POST`/`PUT` - `indexes/:indexUid/documents`.
-
-```json
-{
-    "message": "The `:payloadType` payload must contain at least one document.",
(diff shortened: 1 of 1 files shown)


curquiza:
Is it for v0.26.0?

gmourier:
I was going to schedule it for v0.26  to not add extra work but the load looks very small. What's your opinion on this?

curquiza:
Agreed, I just need to open an issue, I will do it today :)

MarinPostma:
Can you elaborate more on what should be done with an empty update?

gmourier:
> Can you elaborate more on what should be done with an empty update?

This case is already specified in the `ndjson` and `csv` specification e.g. (https://github.com/meilisearch/specifications/blob/main/text/0029-indexing-ndjson.md#errors). 

Sending an empty payload will lead to a 400 Bad Request - **missing_payload** error code.

We can also remove this error for this particular endpoint if we decide to make it error permissive.

bidoubiwa (approved):
Is it worth noting that it stills creates an index? Anyway not sure, just wanting your opinion on it. Apart from that it looks good to me :)

bidoubiwa (dismissed):

PR #100, opened 2021-12-14
Patch Async/Sync part of errors definition

gmourier:
As the index management is now asynchronous. We take the opportunity to update the information of an error if it is asynchronous, synchronous, or both.

@MarinPostma, on the part of the so-called internal errors I'm really not sure. Don't hesitate to make a pass and adjust the synchronicity label if it is wrong. 🙏


curquiza (approved):
Thanks for this details 🙏

PR #101, opened 2021-12-14
Patch `task` `type` enum naming

gmourier:
The naming of task types is updated to be corrected. Seen with @dichotommy 🤓

@meilisearch/docs-team 👋

### DIFF
--- a/text/0060-refashion-updates-apis.md
+++ b/text/0060-refashion-updates-apis.md
@@ -60,7 +60,7 @@ The motivation is to stabilize the current `update` resource to a version that c
 | uid      | integer | Unique sequential identifier           |
 | indexUid | string | Unique index identifier |
 | status  | string  | Status of the task. Possible values are `enqueued`, `processing`, `succeeded`, `failed`                                |
-| type    | string  | Type of the task. Possible values are `indexCreation`, `indexUpdate`, `indexDeletion`, `documentsAddition`, `documentsPartial`, `documentsDeletion`, `settingsUpdate`, `clearAll` |
+| type    | string  | Type of the task. Possible values are `indexCreation`, `indexUpdate`, `indexDeletion`, `documentAddition`, `documentPartial`, `documentDeletion`, `settingsUpdate`, `clearAll` |
 | details | object |  Details information for a task payload. See Task Details part. |
 | error | object | Error object containing error details and context when a task has a `failed` status. See https://github.com/meilisearch/specifications/pull/61|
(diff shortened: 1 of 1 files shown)


text/0060-refashion-updates-apis.md
```diff
@@ -102,9 +102,9 @@ The motivation is to stabilize the current `update` resource to a version that c
 |  -         | indexCreation |
 |  -         | indexUpdate   |
 |  -         | indexDeletion |
-| DocumentsAddition | documentsAddition  |
-| DocumentsPartial | documentsPartial  |
-| DocumentsDeletion  | documentsDeletion |
+| DocumentsAddition | documentAddition  |
+| DocumentsPartial | documentPartial  |
```
curquiza, line 106: Sorry, I'm "relou", but is `document partial` means something? What do you think @dichotommy?

dichotommy: Doesn't mean a lot to me. I'm guessing it's... a document update? As in, `PUT /indexes/:index_uid/documents`

gmourier: It's a deeper problem than that, we didn't find a better term during our thinking. A document addition can also be an update. Partial here is only to indicate that the update is done from the `PUT` endpoint. @MarinPostma correct me if I'm wrong.

dichotommy: I think I understand what the idea was with the name (partial update of document), but since users are required to provide the entire document anyway (it's `PUT` not `PATCH`), and this endpoint can also be used to add entirely new documents, it seems like not the best choice.

dichotommy: If I've correctly understood which endpoint this task type represents, why not `documentUpdate`?

gmourier: I think we should use the verb `PATCH` but in this case, I'm not sure because we don't process only one resource but several `documents` resources. It can make sense if we say that we treat an index resource and that we PATCH (partially) a part of the documents within the index resource.

dichotommy (approved):
This change seems good to me, although it would be _even better_ if `documentPartial` could be changed to something more clear 🚀

text/0060-refashion-updates-apis.md
gmourier: That's something we'll deal with later. We could have a single method that allows you to add documents, update and delete documents from a single payload. (This is just an idea for now)

PR #104, opened 2022-01-04
Release v0.25.0

gmourier:
# Major Changes

- [Task  Monitoring API](https://github.com/meilisearch/specifications/blob/develop/text/0060-tasks-api.md)
- [API Keys Management](https://github.com/meilisearch/specifications/blob/develop/text/0085-api-keys.md)

# Requirements
- [ ] Update Open API Specification


text/0085-api-keys.md
```diff
@@ -0,0 +1,550 @@
...
+
+##### 1.5.3.1 Default Search API Key
+
+The `Default Search API key` gives access to the same rights as the old `public` key.
+
+Here is how the `Default Search API Key` is represented after its generation.
+
+```json
+{
+    "description": "Default Search API Key (Use it to search from the frontend)",
```
brunoocasali, line 85: There is any possibility to prevent the user from using the wrong key in the frontend, and making their infrastructure unsecured?

Maybe an alert, log, or something else in the mini-dashboard. 🌳

text/0085-api-keys.md
```diff
@@ -0,0 +1,550 @@
...
+| key         | string  | The generated key. **Generated by MeiliSearch**. |
+| actions     | array   | A list of actions permitted for the key. `["*"]` for all actions. See Actions list definition part. |
+| indexes     | array   | A list of indexes permitted for the key. `["*"]` for all indexes. |
+| expiresAt   | string  | Represent the expiration date and time as `ISO-8601` format. `null` equals to no expiration time. |
+| createdAt   | string  | Represent the date and time as `ISO-8601` format when the API key has been created. **Generated by MeiliSearch** |
+| updatedAt   | string  | Represent the date and time as `ISO-8601` format when the API key has been updated. **Default**: Value of `createdAt`. **Generated by MeiliSearch** |
+
+#### 1.5.5 `POST`/ `PATCH` - `/keys` - API Key object payload definition
+
+| field       | type    | required |description                     |
```
brunoocasali, line 145: We could add a new column here to list "Default" what do you think? 🧐

brunoocasali (commented):
CC: @guimachiavelli @curquiza 

I'm adding my suggestions, asking about some pertinent things to me at this time, feel free to reach me out to discuss if you want to 😉 

btw: I don't know if this is the right place to send this.

text/0060-tasks-api.md
```diff
@@ -0,0 +1,578 @@
...
+    - `type` is no longer an object. It now becomes a string containing the values of its `name` field previously defined in the `type` object.
+    - The possible values for the `type` field are reworked to be more clear and consistent with our naming rules.
+    - A `details` object is added to contain specific information related to a `task` payload that was previously displayed in the `type` nested object.
+    - An `indexUid` field is added to give information about the related index on which the task is performed.
+    - `duration` format has been updated to express an `ISO 8601` duration.
+    - `processed` status changes to `succeeded`.
+    - `startedProcessingAt` is updated to `startedAt`.
+    - `processedAt` is updated to `finishedAt`.
+- `202 Accepted` requests previously returning an `updateId` are now returning a summarized `task` object.
+- `MEILI_MAX_UDB_SIZE` env var is updated `MEILI_MAX_TASK_DB_SIZE`.
```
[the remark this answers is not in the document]
gmourier: Hi @brunoocasali!  

In general, we do this by specification in dedicated pull-requests while they are in reviews but this is a bit special, it's the pull-request that gathers the specifications that will go into main when v0.25 is released... So it's never too late to fix some typos! Thanks 😃

Indeed it is a replacement because we have renamed the resource `update` by `task`. In this case, `U` meant update but `UDB` as such was not explicit enough.

text/0085-api-keys.md
```diff
@@ -0,0 +1,550 @@
...
+
+## 1. Functional Specification
+
+### 1.1 Summary
+
+Granular management of API keys is added to MeiliSearch. It is possible to restrict the access of an API key to certain actions on specific indexes.
+
+### 1.2 Motivation
+
+To make MeiliSearch more reliable for teams, we extend the management and the possibilities of restrictions for the management of a MeiliSearch instance by introducing a concrete API reséource (`API Key`). Security is a critical need, often tricky to negotiate as the stakes are high for a company.
```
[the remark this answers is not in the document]
gmourier: Thanks! I'm fixing that

text/0085-api-keys.md
gmourier: Absolutely, and it can be done every time we need to describe some API resources object or payload.

text/0085-api-keys.md
gmourier: We can't prevent the use of a "bad" key, because we can't know where it is called. We have added the description to help as much as possible a MeiliSearch user to use the right key for the right purposes. Adding logs could be very verbose but it can be considered according to the feedback and the use that will be made of it.

guimachiavelli: For what it's worth, we do our best in the docs to make it clear to users they need to be careful with keys—especially when it comes to the master key.

Which reminds me of something @gmourier and I spoke about a while ago. The master key can access all API endpoints, even when by design we want people to only use it when managing API keys. We could go a step further beyond warnings and simply block the master key specifically from accessing any routes except `/keys`. This wouldn't have an effect on API keys, though.

brunoocasali: > We could go a step further beyond warnings and simply block the master key specifically from accessing any routes except /keys. This wouldn't have an effect on API keys, though.

Well, I think that definitely will prevent this kind of mistake @guimachiavelli, (I'm always a pessimist on these subjects 😅)

gmourier: @guimachiavelli That's a good idea, subject to debate because the general philosophy that seems to prevail is not to restrict or force the hand of the users by blocking him. We had/have many debates on this subject with @bidoubiwa during our design reviews/feedbacks discussion, that said, from a purely personal point of view when it comes to security, I'm rather for restricting in order to better accompany the user rather than leaving him the possibility to put himself in danger.

guimachiavelli: @gmourier: apologies if it sounded as if I were saying that we should totally do it 😅. Like you and @brunoocasali, I'm also for restricting harmful behaviour, but there's certainly a case to be made for educating rather than forcing users to "do the right thing."

PR #105, opened 2022-01-12
Dumps

gmourier:
## Objectives

This specification is added to define the current behavior of dumps and add some small evolutions to make the use of CLI flags consistent with the snapshot feature by adding `--ignore-missing-dump` and `--ignore-dump-if-db-exists`.

It is also an opportunity to provide the @meilisearch/saas-team with a CLI that allows them to do automatic upgrades more easily.

A current pain point that prevents them from moving forward on this feature with the current state of the dumps is that if a Kubernetes pod drops and restarts if it was created/launched with `import-dump` in the command, will overwrite the database by re-executing the command when automatically re-launching the MeiliSearch instance.  

Adding `ignore-dump-if-db-exists`, allows them to solve this problem.

### DIFF
--- a/text/0000-dumps.md
+++ b/text/0000-dumps.md
@@ -26,7 +26,7 @@ The dumps exists to upgrade from a previous version to a more recent version. It
 - The `dumps` directory can be customized using the `--dumps-dir` configuration option. If the dump directory does not already exist when the dump creation process is called, MeiliSearch will create it.
 - If MeiliSearch is restarted after a dump creation, the dump's status will not appear on the `GET - /dumps/:uid/status` endpoint.
 - A `.dump` file can be imported using the `--import-dump` command-line flag.
-- The MeiliSearch server will started when the dump has been fully imported.
+- The MeiliSearch server starts when the dump is fully imported and indexed.
 - By default, importing a dump when a database already exists (a non-empty data.ms folder in the same directory as the MeiliSearch binary) will stop the process and throw an error.
 - When using the command-line flag `--ignore-dump-if-db-exists=true`, MeiliSearch will use the existing database to start an instance instead of throwing an error. The dump will be ignored.
 - By default, trying to import a dump that do not exists, will stop the process and throw an error.
--- a/text/0000-dumps.md
+++ b/text/0000-dumps.md
@@ -1,0 +1,120 @@
+- Title: Dumps
+- Start Date: 2022-01-12
+- UpdatedAt: 2022-01-12
+
+# Dumps
+
+## 1. Functional Specification
+
+### 1.1 Summary
+
+A dump is a compressed file containing an export of a MeiliSearch instance. It contains all indexes, documents, settings, tasks, and api keys but in a raw unprocessed form. A dump isn't an exact copy of a database—it is closer to a blueprint that allows to create an identical dataset.
+
+### 1.2 Motivation
+
+The dumps exists to upgrade from a previous version to a more recent version. It can also be a useful tool for loading a production state on a staging server to make changes and test them before propagating them to production.
+
+### 1.3 Explanations
+
+#### 1.3.1 Summary Key Points
+
+- A dump creation can be scheduled from the MeiliSearch API using the `POST - /dumps` endpoint.
+- A dump creation status can be tracked using the `GET - /dumps/{uid}/status` endpoint.
+- MeiliSearch can only create one dump at a time.
+- While a dump is in creation, the task queue is paused and no write operations can occur on the database.
+- By default, dumps are created in a folder named `dumps`, and can be found in the same directory as the MeiliSearch binary.
+- The `dumps` directory can be customized using the `--dumps-dir` configuration option. If the dump directory does not already exist when the dump creation process is called, MeiliSearch will create it.
+- If MeiliSearch is restarted after a dump creation, the dump's status will not appear on the `GET - /dumps/:uid/status` endpoint.
+- A `.dump` file can be imported using the `--import-dump` command-line flag.
+- The MeiliSearch server will started when the dump has been fully imported.
+- By default, importing a dump when a database already exists (a non-empty data.ms folder in the same directory as the MeiliSearch binary) will stop the process and throw an error.
+- When using the command-line flag `--ignore-dump-if-db-exists=true`, MeiliSearch will use the existing database to start an instance instead of throwing an error. The dump will be ignored.
+- By default, trying to import a dump that do not exists, will stop the process and throw an error.
+- When using the command-line flag `--ignore-missing-dump`, MeiliSearch will continue its process and not throw an error.
+
+---
+
+#### 1.3.2 API Definition
+
+##### **As a user, I want to create a dump**
+
+##### Request Definition
+
+`POST` - `/dumps`
+
+##### Headers
+
+```
+"Authorization: Bearer :apiKey"
+"Content-Type: application/json"
+```
+
+##### Body Payload
+N/A
+
+##### Response
+
+`202 Accepted`
+
+```json
+{
+    "uid": "20220112-151751438",
+    "status": "in_progress",
+    "startedAt": "2022-01-12T15:17:51.438881Z"
+}
+```
+
+##### Errors
+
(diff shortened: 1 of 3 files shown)


text/0000-dumps.md
```diff
@@ -0,0 +1,120 @@
...
+
+- A dump creation can be scheduled from the MeiliSearch API using the `POST - /dumps` endpoint.
+- A dump creation status can be tracked using the `GET - /dumps/{uid}/status` endpoint.
+- MeiliSearch can only create one dump at a time.
+- While a dump is in creation, the task queue is paused and no write operations can occur on the database.
+- By default, dumps are created in a folder named `dumps`, and can be found in the same directory as the MeiliSearch binary.
+- The `dumps` directory can be customized using the `--dumps-dir` configuration option. If the dump directory does not already exist when the dump creation process is called, MeiliSearch will create it.
+- If MeiliSearch is restarted after a dump creation, the dump's status will not appear on the `GET - /dumps/:uid/status` endpoint.
+- A `.dump` file can be imported using the `--import-dump` command-line flag.
+- The MeiliSearch server will started when the dump has been fully imported.
```
irevoire, line 29:
```suggestion
- The MeiliSearch server starts when the dump is fully imported and indexed.
```

text/0000-dumps.md
```diff
@@ -0,0 +1,120 @@
...
+The dumps exists to upgrade from a previous version to a more recent version. It can also be a useful tool for loading a production state on a staging server to make changes and test them before propagating them to production.
+
+### 1.3 Explanations
+
+#### 1.3.1 Summary Key Points
+
+- A dump creation can be scheduled from the MeiliSearch API using the `POST - /dumps` endpoint.
+- A dump creation status can be tracked using the `GET - /dumps/{uid}/status` endpoint.
+- MeiliSearch can only create one dump at a time.
+- While a dump is in creation, the task queue is paused and no write operations can occur on the database.
```
irevoire, line 24: If I remember correctly what we did, while a dump is being created the task queue can't process any other updates, but it can still register new updates.
Do you confirm @MarinPostma?

text/0000-dumps.md
```diff
@@ -0,0 +1,120 @@
...
+    "startedAt": "2022-01-12T15:17:51.438881Z"
+}
+```
+
+##### Errors
+
+- 🔴 Accessing this route without the `Authorization` header returns a [missing_authorization_header](0061-error-format-and-definitions.md#missing_authorization_header) error.
+- 🔴 Accessing this route with a key that does not have permissions (i.e. other than the master-key) returns an [invalid_api_key](0061-error-format-and-definitions.md#invalid_api_key) error.
+- 🔴 Attempting to create a dump while a dump is already being created return an [dump_already_in_progress](0061-error-format-and-definitions.md#dump_already_in_progress) error.
+
```
irevoire, line 72:
```suggestion
- 🔴 Attempting to create a dump while a dump is already being created return an [dump_already_in_progress](0061-error-format-and-definitions.md#dump_already_in_progress) error.
- 🔴 Attempting to create a dump when MeiliSearch doesn't have the right to write in the dump directory.
- 🔴 Attempting to create a dump when there is no space left on the device.
```

I don't know if we list these errors usually.
Also, in the code, these two are considered as `Internal Error`, so maybe it's not helpful to list them.

gmourier: It's a good idea to list them 🤓


> - 🔴 Attempting to create a dump when MeiliSearch doesn't have the right to write in the dump directory.

Regarding this one, we have a `dump_creation_failed` that can be asynchronous given the specification. Could it be this one? Where is the error of a dump displayed asynchronously as it does not appear in the tasks? I can only think of the stdout 🤔 

>  - 🔴 Attempting to create a dump when there is no space left on the device.

Regarding this one, we have a `no_space_left_on_device` error. I may expect it to be async and be on the throw on the stdout as well as the previous one. Am I wrong?

gmourier:
Signal to @irevoire, @curquiza, @meilisearch/docs-team. 

Compared to the current version, version `0.26.0` will add these behaviors:

1. When importing a dump using `--import-dump` if a database already exists, an error message is thrown on the CLI. This message is `Error: database already exists at ":pathToDataMs/data.ms", try to delete it or rename it`.
2. Similar to the behavior of the snapshot feature. The flags `--ignore-dump-if-db-exists` and `--ignore-missing-dump` are added.

text/0105-dumps.md
```diff
@@ -0,0 +1,145 @@
...
+
+## 1. Functional Specification
+
+### 1.1 Summary
+
+A dump is a compressed file containing an export of a MeiliSearch instance. It contains all indexes, documents, settings, tasks, and api keys but in a raw unprocessed form. A dump isn't an exact copy of a database—it is closer to a blueprint that allows creating of an identical dataset.
+
+### 1.2 Motivation
+
+The dumps exist to upgrade from a previous version to a more recent version. It can also be a helpful tool for loading a production state on a staging server to make changes and test them before propagating them to production.
```
curquiza, line 15:
```suggestion
The dumps exist to upgrade a MeiliSearch instance from a previous version to a more recent version. It can also be a helpful tool for loading a production state on a staging server to make changes and test them before propagating them to production.
```

text/0105-dumps.md
```diff
@@ -0,0 +1,145 @@
...
+- Start Date: 2022-01-12
+- UpdatedAt: 2022-01-12
+
+# Dumps
+
+## 1. Functional Specification
+
+### 1.1 Summary
+
+A dump is a compressed file containing an export of a MeiliSearch instance. It contains all indexes, documents, settings, tasks, and api keys but in a raw unprocessed form. A dump isn't an exact copy of a database—it is closer to a blueprint that allows creating of an identical dataset.
```
curquiza, line 11:
```suggestion
A dump is a compressed file containing an export of a MeiliSearch instance. It contains all indexes, documents, settings, tasks, and API keys but in a raw unprocessed form. A dump isn't an exact copy of a database—it is closer to a blueprint that allows creating of an identical dataset.
```

text/0105-dumps.md
```diff
@@ -0,0 +1,145 @@
...
+- By default, dumps are created in a folder named `dumps`, and can be found in the same directory as the MeiliSearch binary.
+- The `dumps` directory can be customized using the `--dumps-dir` configuration option. If the dump directory does not already exist when the dump creation process is called, MeiliSearch will create it.
+- If MeiliSearch is restarted after a dump creation, the dump's status will not appear on the `GET - /dumps/:uid/status` endpoint.
+- A `.dump` file can be imported using the `--import-dump` command-line flag.
+- The MeiliSearch server starts when the dump is fully imported and indexed.
+- By default, importing a dump when a database already exists (a non-empty data.ms folder in the same directory as the MeiliSearch binary) will stop the process and throw an error.
+- When using the command-line flag `--ignore-dump-if-db-exists=true`, MeiliSearch will use the existing database to start an instance instead of throwing an error. The dump will be ignored.
+- By default, trying to import a dump that do not exists, will stop the process and throw an error.
```
curquiza, line 32:
```suggestion
- By default, trying to import a dump that does not exist, will stop the process and throw an error.
```

curquiza (commented):
👆 Just noticed you already included the limits we wrote on the issue, my bad

bidoubiwa:
Does this change the behavior of the dumps compared to the v0.25 ?

text/0000-dumps.md
MarinPostma: Yes it can register new updates while processing dump

gmourier: Thanks guys!

@meilisearch/docs-team "While a dump is processing, the task queue is paused and no write operations can occur on the database." We may want to delete this statement in the [documentation](https://docs.meilisearch.com/reference/api/dump.html#create-a-dump) 📡

gmourier: Hi @maryamsulemani97 

> When a dump is being imported, the task queue is paused and no write operations can occur on the database.

Yes, to be more clear, the task queue is not really **paused**, the API is not available at all. No **read or write** operations can be performed, except the import of the dump into the database at that time.

> When a dump is being created, the task queue can receive other future operations to perform but will not process any further tasks during dump creation, but will still accept new tasks in the queue.

Yes!

PR #106, opened 2022-01-12
Add telemetry node to get auto inferred identifiers during `Documents Added` and `Documents Updated` events

gmourier:
# Objectives

Determine if this automatic inference chooses `id` most of the time. 

Given this information, we may propose a new way to determine identifiers and thus remove the automatic inference that takes the first field containing `id` during the first indexing. We already observed that the usage of `primaryKey` is mostly used to specify `id` or `_id`. 

It may turn out that this automatic inference causes more problems than it facilitates access to MeiliSearch, we want to acknowledge that fact.

### DIFF
--- a/text/0034-telemetry-policies.md
+++ b/text/0034-telemetry-policies.md
@@ -1,6 +1,6 @@
 - Title: Anonymous Analytics Policy
 - Started At: 2021-04-16
-- Updated At: 2021-10-13
+- Updated At: 2022-01-12
 
 # Anonymous Analytics Policy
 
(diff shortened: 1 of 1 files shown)


irevoire (approved):
:+1:

gmourier:
Signal for the @meilisearch/docs-team 📚

gmourier:
Closed for now.

It takes too much time and changes in the code to get this information. We'll do something else and get more qualitative feedback.

Signal to @curquiza, @meilisearch/docs-team 📡

PR #109, opened 2022-01-27
Update the integration analytics following the rebranding

curquiza:
Checked with @gmourier: @brunoocasali, we need to adapt with the rebranding :)


gmourier (approved):

brunoocasali (approved):
Oh yeah! I remembered that when I was going to bed yesterday! :p

PR #110, opened 2022-02-03
Handle empty cell as `null` value for CSV format

gmourier:
# Summary

Consider empty cell in a CSV payload as `null` value at indexing.

## `null` value

- If a field is of type `string`, then an empty cell is considered as a `null` value (e.g. `,,`), anything other is turned into a string value (e.g. `, ,` is a single whitespace string)
- If a field is of type `number`, when the trimmed field is empty, it's considered as a `null` value (e.g. `,,` `, ,`); otherwise Meilisearch try to parse the number.

## Example

Given the CSV payload
```
id:number,label,price:number,colors
1,t-shirt,,red
```
the search result should be displayed as
```json
{
  "hits": [
    {
      "id": 1,
      "label": "t-shirt",
      "price": null,
      "colors": "red"
    }
  ],
  ...
}
```

### DIFF
--- a/text/0028-indexing-csv.md
+++ b/text/0028-indexing-csv.md
@@ -49,6 +49,11 @@ While there's [RFC 4180](https://tools.ietf.org/html/rfc4180) as a try to add a 
 - CSV text should be encoded in UTF8.
 - The format can't handle array cell values. We are providing `nd-json` format to deal with theses types of attribute in a easier way.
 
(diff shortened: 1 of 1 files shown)


Kerollmops (changes requested):
Apart from this small change of wording, looks good to me!

Kerollmops (approved):

PR #114, opened 2022-02-16
Keys API - Update the error message for `expiresAt`

irevoire:
This follow this discussion (monologue?): https://github.com/meilisearch/meilisearch/issues/2172


irevoire:
Also I noticed none of the errors related to the API keys use the backquote when talking about the field the caused an error, I think we should patch that too, I can do it in this PR if you’re ok with the change.

For example:
```
indexes field value `:value` is invalid. It should be an array of string representing index names.
```

Would become:
```
`indexes` field value `:value` is invalid. It should be an array of string representing index names.
```

gmourier:
I was misguided by the format and thought we were supporting ISO 8601, but we are actually dealing with RFC 3339 (that format is sharing notation we use with the ISO 8601 format).

We still express a task `duration` with the ISO-8601 period part.

gmourier (approved):

curquiza (approved):

PR #115, opened 2022-02-21
Tasks API - Seek/Keyset based pagination

gmourier:
🤖 [API Diff](https://github.com/meilisearch/specifications/pull/115#issuecomment-1129027036)

## Summary

`task` resource lists now support seek pagination, allowing users to browse multiple sets of `task` items.

### Summary Key Points

- Add `limit`, `from`, `next` response attributes to task list response.
- It only concerns the `/tasks` endpoint. `/indexes/:uid/tasks` endpoint is removed by https://github.com/meilisearch/specifications/pull/116.

## Motivation

Following the specification aiming to stabilize the `task` API resource, we want to give Meilisearch capabilities to adapt to many use-cases, thus involving unpredictable orders of magnitude regarding `task` growth over time.

This feature should be usable by both solo developers and technical teams.

The objectives set during the discovery's exploration phase for the pagination of the task lists are:

- Performance should remain nearly constant, regardless of task volume.
- The results must be browsable in a consistent way between each call to the API, no matter how much the number of tasks is increased or reduced while the navigation is happening.
- Do not reinvent the wheel and therefore facilitate its implementation by selecting a proven paging interface.
- A subsidiary, but critical objective, is to feed our API guideline to respond quickly to such needs in the future within the company. By proposing a proven solution, we can build faster with confidence.

### DIFF
--- a/text/0060-tasks-api.md
+++ b/text/0060-tasks-api.md
@@ -568,6 +568,138 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
 }
 ```
 
+#### 9. Paginate `task` resource lists
+
+The API endpoints `/tasks` and `indexes/{indexUid}/tasks` are browsable using a cursor based pagination.
+
+##### 9.1 Why a cursor based pagination?
+
+As seen in the [Rest API Format Convention](https://github.com/meilisearch/product/issues/44#issuecomment-895888679), cursor-based pagination is more appropriate when the data can grow or shrink quickly in terms of magnitude.
+
+###### 9.1.1 Pros
+
+The performance is better than the not-so-good but old pagination with `offset`/`limit`.
+
+Cursor pagination keeps the results consistent between each page as the data evolves. It avoids the [Page Drift effect](https://use-the-index-luke.com/sql/partial-results/fetch-next-page), especially when the data is sorted from the most recent to the oldest.
+
+Moreover, the performance is superior to traditional pagination since the computational complexity remains constant to reach the identifier marking the beginning of the new slice to be returned from a hash table.
+
+###### 9.1.2 Cons
+
+The main drawback of this type of pagination is that it does not navigate within a finite number of pages. It is also limited to a precise sorting criterion on unique identifiers ordered sequentially.
+
+##### 9.2 Response attributes
+
+| field | type | description                          |
+|-------|------|--------------------------------------|
+| limit | integer  | Default `30`. |
+| after | integer - nullable  | Represents the query parameter to send to fetch the next slice of the results. The first item for the next slice starts at `after+1`. When the returned value is null, it means that all the data have been browsed in the given order. |
+
+##### 9.3 GET query parameters
+
+| field | type | required | description  |
+|-------|------|----------|--------------|
+| limit | integer  | No       | Default `30`. Limit on the number of tasks to be returned, between `1` and `100`. |
+| after | integer  | No       | Limit results to tasks with uids greater/lower than the specified uid. |
+
+##### 9.4 Usage examples
+
+This part demonstrates cursor paging on `/tasks`, but it should be equivalent for `indexes/:uid/tasks`. Although the uids can be "holey" on the `/indexes/:uid/tasks` endpoint if several indexes are managed within the instance. The items uid remains sorted sequentially and can be used to navigate a list of `tasks` objects.
+
+---
+
+**Initial default slice of `tasks`**
+
+`GET` - `/tasks`
+
+```json
+{
+    "results": [
+        {
+            "uid": 1350,
+            "indexUid": "movies",
+            "type": "documentAddition",
+            ...,
+        },
+        ...,
+        {
+            "uid": 1330,
+            "indexUid": "movies_reviews",
+            "type": "documentAddition",
+            ...,
+        }
+    ],
+    "limit": 20,
+    "after": 1330
+}
+```
+
+**Request the next slice of `tasks` items with a limit of `50` tasks**
+
+`GET` - `/tasks?after=1330&limit=50`
+
+```json
+{
+    "results": [
+        {
+            "uid": 1329,
+            "indexUid": "movies",
+            "type": "documentAddition",
+            ...,
+        },
+        ...,
+        {
+            "uid": 1279,
+            "indexUid": "movies",
+            "type": "settingsUpdate",
+            ...,
+        }
+    ],
+    "limit": 50,
+    "after": 1279
+}
+```
+
+**End of cursor pagination**
+
+`GET` - `/tasks?after=20`
+
+```json
+{
+    "results": [
+        {
+            "uid": 19,
+            "indexUid": "movies",
+            "type": "documentsAddition",
+            ...,
+        },
+        ...,
+        {
+            "uid": 0,
+            "indexUid": "movies",
+            "type": "documentsAddition",
+            ...,
+        }
+    ],
+    "limit": 10,
+    "after": null
+}
+```
+
+- 💡 `after` response parameter is null because there are no more `tasks` to fetch. It means that the response represents the last slice of results for the given resource list.
+
+##### 9.5 Behaviors for `limit` and `after` query parameters
+
+###### 9.5.1 `limit`
+
+- If `limit` is not set, the default value is chosen.
+- If `limit` is set and it is not between the minimum and maximum values, the default value is chosen.
+- If `limit` is not an integer, the default value is chosen.
+
+###### 9.5.2 `after`
+
+- If `after` is set with an out of bounds task `uid`, the response returns an empty `results` array and `after` is set to `null`.
+
 ## 2. Technical details
 
 ### I. Measuring
--- a/text/0060-tasks-api.md
+++ b/text/0060-tasks-api.md
@@ -593,14 +593,15 @@ The main drawback of this type of pagination is that it does not navigate within
 | field | type | description                          |
 |-------|------|--------------------------------------|
 | limit | integer  | Default `20`. |
-| after | integer - nullable  | Represents the query parameter to send to fetch the next slice of the results. The first item for the next slice starts at `after+1`. When the returned value is null, it means that all the data have been browsed in the given order. |
+| from | integer | The first task uid returned |
+| next | integer - nullable  | Represents the query parameter to send to fetch the next slice of the results. The first item for the next slice starts at this exact number. When the returned value is null, it means that all the data have been browsed in the given order. |
 
 ##### 9.3 GET query parameters
 
 | field | type | required | description  |
 |-------|------|----------|--------------|
 | limit | integer  | No       | Default `20`. Limit on the number of tasks to be returned. |
-| after | integer  | No       | Limit results to tasks with uids greater/lower than the specified uid. |
+| from | integer  | No       | Limit results to tasks with uids equal to and lower than this uid. |
 
 ##### 9.4 Usage examples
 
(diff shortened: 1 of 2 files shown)


bidoubiwa:
I exported all my tasks and it weights 3.1M

`3.1M Mar 16 12:35 tasks.json` 

So 💯💯💯

text/0060-tasks-api.md
```diff
@@ -568,6 +568,138 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+```
+
+- 💡 `after` response parameter is null because there are no more `tasks` to fetch. It means that the response represents the last slice of results for the given resource list.
+
+##### 9.5 Behaviors for `limit` and `after` query parameters
+
+###### 9.5.1 `limit`
+
+- If `limit` is not set, the default value is chosen.
+- If `limit` is set and it is not between the minimum and maximum values, the default value is chosen.
```
brunoocasali, line 696: What are the min and max values? They are hardcoded in the engine?

brunoocasali (commented):
Clean spec @gmourier 🧱  🎉

text/0060-tasks-api.md
gmourier: min is always 0, while maximum is the total number of task. I will rephrase it! Thanks @brunoocasali

gmourier: I think the `maximum` could be adjusted given some tests!

gmourier:
Signal @meilisearch/docs-team and @meilisearch/devrel-team; perhaps we should use the terms seek or keyset pagination. 

Using the term cursor does not seem to be appropriate. Indeed, pagination by cursor makes me think that a state is established on the server-side to manipulate a cursor's instance, which is not the case here. There is no particular state being created.

gmourier:
@meilisearch/integration-team and @meilisearch/core-team 

I have a design question, which I'm not sure about. In the specification, I've specified a maximum acceptable value of `100` for the `limit` field to make sure that the performance stays good.

What do you think about not setting a maximum acceptable value for the `limit` field and let users report performance problems to adjust this value in the future better if needed?

irevoire:
Personally, I would prefer not to have any hard limit for that route. If someone has a performance problem while asking for 1000000 things, he’ll know it’s his fault.

Also, that’s what I usually see on other products.

curquiza:
Since the `/tasks` endpoint is a management endpoint and is not exposed to the wild internet (contrary to the `/search` endpoint), I agree with Tamo, it's ok not putting any limit

gmourier:
- Specification uses the word "keyset" instead of "cursor"
- Removes the boundaries for the `limit` query-parameter

text/0060-tasks-api.md
```diff
@@ -568,6 +568,137 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+
+##### 9.1 Why a keyset based pagination?
+
+Keyset-based pagination is more appropriate when the data can grow or shrink quickly in terms of magnitude.
+
+###### 9.1.1 Pros
+
+The performance is better than the not-so-good but old pagination with `offset`/`limit`.
+
+Cursor pagination keeps the results consistent between each page as the data evolves. It avoids the [Page Drift effect](https://use-the-index-luke.com/sql/partial-results/fetch-next-page), especially when the data is sorted from the most recent to the oldest.
```
bidoubiwa, line 583: Is cursor still the term you want to use here?

text/0060-tasks-api.md
```diff
@@ -568,6 +568,137 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+###### 9.1.2 Cons
+
+The main drawback of this type of pagination is that it does not navigate within a finite number of pages. It is also limited to a precise sorting criterion on unique identifiers ordered sequentially.
+
+##### 9.2 Response attributes
+
+| field | type | description                          |
+|-------|------|--------------------------------------|
+| limit | integer  | Default `20`. |
+| after | integer - nullable  | Represents the query parameter to send to fetch the next slice of the results. The first item for the next slice starts at `after+1`. When the returned value is null, it means that all the data have been browsed in the given order. |
```
bidoubiwa, line 596:
```suggestion
| after | integer - nullable  | Used to fetch the next slice of the results. The first item for the next slice starts at `after+1`. When the returned value is null, it means that all the data have been browsed in the given order. |
```

text/0060-tasks-api.md
```diff
@@ -568,6 +568,137 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+
+##### 9.3 GET query parameters
+
+| field | type | required | description  |
+|-------|------|----------|--------------|
+| limit | integer  | No       | Default `20`. Limit on the number of tasks to be returned. |
+| after | integer  | No       | Limit results to tasks with uids greater/lower than the specified uid. |
```
Kerollmops, line 603:
```suggestion
| limit | integer  | No       | Default `20`. Limit on the number of tasks to be returned. |
| from | integer  | No       | Limit results to tasks with uids equal to and lower than this uid. |
```

text/0060-tasks-api.md
```diff
@@ -568,6 +568,137 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+        }
+    ],
+    "limit": 50,
+    "after": 1279
+}
+```
+
+**End of cursor pagination**
+
+`GET` - `/tasks?after=20`
```
Kerollmops, line 665:
```suggestion
`GET` - `/tasks?from=20`
```

text/0060-tasks-api.md
```diff
@@ -568,6 +568,137 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+            "type": "documentsAddition",
+            ...,
+        }
+    ],
+    "limit": 10,
+    "after": null
+}
+```
+
+- 💡 `after` response parameter is null because there are no more `tasks` to fetch. It means that the response represents the last slice of results for the given resource list.
```
Kerollmops, line 689:
```suggestion
- 💡 `next` response parameter is null because there are no more `tasks` to fetch. It means that the response represents the last slice of results for the given resource list.
```

text/0060-tasks-api.md
```diff
@@ -568,6 +568,137 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+##### 9.5 Behaviors for `limit` and `after` query parameters
+
+###### 9.5.1 `limit`
+
+- If `limit` is not set, the default value is chosen.
+- If `limit` is not an integer, the default value is chosen.
+
+###### 9.5.2 `after`
+
+- If `after` is set with an out of bounds task `uid`, the response returns an empty `results` array and `after` is set to `null`.
```
Kerollmops, line 700:
```suggestion
##### 9.5 Behaviors for `limit` and `from` query parameters

###### 9.5.1 `limit`

- If `limit` is not set, the default value is chosen.

###### 9.5.2 `from`

- If `from` is set with an out of bounds task `uid`, the response returns the tasks that are the nearest to the specified uid, the `next` field is set to the next page. It will be equivalent to call the `/tasks` route without any parameter.
```

text/0060-tasks-api.md
gmourier: No! Good spot!

text/0060-tasks-api.md
```diff
@@ -568,6 +568,145 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+- If `limit` is not set, the default value is chosen.
+
+###### 9.5.2 `from`
+
+- If `from` is set with an out of bounds task `uid`, the response returns the tasks that are the nearest to the specified uid, the `next` field is set to the next page. It will be equivalent to call the `/tasks` route without any parameter.
+
+###### 9.5.3 Errors
+
+- 🔴 Sending a value with a different type than `Integer` for `limit` returns a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
+- 🔴 Sending a value with a different type than `Integer` for `from` returns a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
```
bidoubiwa, line 708: Does the bad_request specify which field is wrong?

bidoubiwa: Thanks for the information :)

text/0060-tasks-api.md
```diff
@@ -568,6 +568,145 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+##### 9.3 GET query parameters
+
+| field | type | required | description  |
+|-------|------|----------|--------------|
+| limit | integer  | No       | Default `20`. Limit on the number of tasks to be returned. |
+| from | integer  | No       | Limit results to tasks with uids equal to and lower than this uid. |
+
+##### 9.4 Usage examples
+
+This part demonstrates keyset paging in action on `/tasks`, but it should be equivalent for `indexes/:uid/tasks`. Although the uids can be "holey" on the `/indexes/:uid/tasks` endpoint if several indexes are managed within the instance. The items uid remains sorted sequentially and can be used to navigate a list of `tasks` objects.
```
bidoubiwa, line 608: What do you mean with holey?

text/0060-tasks-api.md
```diff
@@ -568,6 +568,145 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+
+The main drawback of this type of pagination is that it does not navigate within a finite number of pages. It is also limited to a precise sorting criterion on unique identifiers ordered sequentially.
+
+##### 9.2 Response attributes
+
+| field | type | description                          |
+|-------|------|--------------------------------------|
+| limit | integer  | Default `20`. |
+| from | integer | The first task uid returned |
+| next | integer - nullable  | Represents the value to send in `from` to fetch the next slice of the results. The first item for the next slice starts at this exact number. When the returned value is null, it means that all the data have been browsed in the given order. |
```
bidoubiwa, line 597:
```suggestion
| next | integer - nullable  | Represents the value to send in `from` to fetch the next slice of the results. The first item for the next slice starts at this exact number. When the returned value is null, it means that all the data have been browsed in the given order. `next = form +/- limit` depending on the order of the tasks |
```
For me it is a bit more clear but feel free to ignore

text/0060-tasks-api.md
```diff
@@ -568,6 +568,145 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+
+The performance is better than the not-so-good but old pagination with `offset`/`limit`.
+
+Seek/Keyset pagination keeps the results consistent between each page as the data evolves. It avoids the [Page Drift effect](https://use-the-index-luke.com/sql/partial-results/fetch-next-page), especially when the data is sorted from the most recent to the oldest.
+
+Moreover, the performance is superior to traditional pagination since the computational complexity remains constant to reach the identifier marking the beginning of the new slice to be returned from a hash table.
+
+###### 9.1.2 Cons
+
+The main drawback of this type of pagination is that it does not navigate within a finite number of pages. It is also limited to a precise sorting criterion on unique identifiers ordered sequentially.
```
bidoubiwa, line 589: What do you mean with `does not navigate within a finite number of pages` ?

bidoubiwa (commented):
Very nice 🔥

text/0060-tasks-api.md
```diff
@@ -568,6 +568,145 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
 }
 ```
 
+#### 9. Paginate `task` resource lists
+
+The API endpoints `/tasks` and `indexes/{indexUid}/tasks` are browsable using a keyset based pagination.
```
dichotommy, line 573:
```suggestion
The API endpoint `/tasks` is browsable using a keyset-based pagination.
```

text/0060-tasks-api.md
gmourier: It is not possible to access a specific page like with `offset`/`limit`, the navigation is done as you go with `from`

PR #116, opened 2022-02-21
Tasks API - Filter tasks list by `type`/`status`/`indexUid`

gmourier:
🤖  [API Diff](https://github.com/meilisearch/specifications/pull/116#issuecomment-1105255632)

# Summary

Add filtering capabilities to the `tasks` endpoints to facilitate the management of an instance and its indexes, or a specific index.

This first iteration adds filters on the `status`, `type` and `indexUid` attributes of the `task` API resource.

## Changes

- Add filtering capabilities on `type`, `status` and `indexUid` for `GET` `task` lists endpoints.
- It is possible to specify several values for a filter using the `,` character. e.g. `?status=enqueued,processing`
- Between two different filters, an AND operation is applied. e.g. `?status=enqueued&type=indexCreation` is equivalent to `status=enqueued AND type = indexCreation`
- `GET /indexes/:indexUid/tasks` is removed
    - It can be replaced by `GET /tasks?indexUid=:indexUid`
- `GET /indexes/:indexUid/tasks/:taskUid` is removed
    - It's already replaced by `GET /tasks/:taskUid`
- Task types `documentPartial` and `documentAddition` are renamed `documentAdditionOrUpdate`
- Task type `clearAll` is renamed `documentDeletion`

### Telemetry

- `Index Tasks Seen` event is removed.
- Add `filtered_by_index_uid` boolean property for `Tasks Seen` event.
- Add `filtered_by_type` boolean property for `Tasks Seen` event.
- Add `filtered_by_status`boolean property for `Tasks Seen` event.

# Motivation

Following the specification aiming to stabilize the `task` API resource, we want to give users the capability to refine the lists of task according to several criteria to find precise information more quickly.

### DIFF
--- a/text/0060-tasks-api.md
+++ b/text/0060-tasks-api.md
@@ -700,6 +700,5 @@ n/a
 
 - Use Hateoas capability to give direct access to a `task` resource.
 - Add dedicated task type names modifying a sub-setting. e.g. `SearchableAttributesUpdate`.
-- Reconsider existence of `/indexes/:indexUid/tasks/:taskUid` route since it is similar to `/tasks/:taskUid`.
 - Add an archived state for old `tasks`.
 - Indicate the `API Key` identity that added a `task`.
--- a/text/0060-tasks-api.md
+++ b/text/0060-tasks-api.md
@@ -1,9 +1,4 @@
-- Title: Tasks API
-- Start Date: 2021-08-13
-- Specification PR: [#60](https://github.com/meilisearch/specifications/pull/60)
-- Discovery Issue: [#48](https://github.com/meilisearch/product/issues/48)
-
-# Refashion Updates APIs
+# Tasks API
 
 ## 1. Functional Specification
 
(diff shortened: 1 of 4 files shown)


bidoubiwa:
Is it possible to filter on `index` ? Since using `GET /indexes/:index/task/0` would cause to throw a 404 error when creating the index we could suggest to users to always use `GET /tasks&index=movies` as it would never return a 404. 

Also, by doing so, we would avoid having to fetch all tasks and then filter them just to have the ones from one index. Which in turn if there are a lot of different indexes may result in the user having to paginate to be sure they fetched all the tasks 

Exemple:

```js
const tasks = await client.getTasks() // could contain a lot of tasks unrelated to the index
const indexTasks = tasks.filter(task => indexUid === "movies")
``` 

vs 

```js
const tasks = await client.getTasks({ indexUid: "movies" })
```

text/0061-error-format-and-definitions.md
```diff
@@ -728,6 +728,54 @@ HTTP Code: `404 Not Found`
...
+{
+    "message": "`:status` is invalid. Available task statuses are: `:taskStatuses`.",
+    "code": "invalid_task_status",
+    "type": "invalid_request",
+    "link":"https://docs.meilisearch.com/errors#invalid_task_status"
+}
+```
+
+- The `:status` is inferred when the message is generated.
+- The `:taskStatuses` is inferred when the message is generated.
```
MarinPostma, line 751: I am not sure what that means

MarinPostma:
> Is it possible to filter on `index` ? Since using `GET /indexes/:index/task/0` would cause to throw a 404 error when called directly after creating the index. If added, we could suggest to users to always use `GET /tasks&index=movies` as it would never return a 404.
> 
> Also, by doing so, we would avoid having to fetch all tasks and then filter them just to have the ones from one index. Which in turn if there are a lot of different indexes may result in the user having to paginate to be sure they fetched all the tasks
> 
> Exemple:
> 
> ```js
> const tasks = await client.getTasks() // could contain a lot of tasks unrelated to the index
> const indexTasks = tasks.filter(task => indexUid === "movies")
> ```
> 
> vs
> 
> ```js
> const tasks = await client.getTasks({ indexUid: "movies" })
> ```

technically calling `/indexes/:index/tasks` is syntactic sugar for a filter on `index`. The spec doesn't say anything about allowing filters on `/indexes/:uid/tasks` route, but I think it should be part of this.

text/0060-tasks-api.md
```diff
@@ -568,6 +568,151 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
+            "type": "succeeded",
+            "type": "documentAddition",
+            ...,
+        }
+    ],
+    ...
+}
+```
+
+- 💡 `status` and `type` can be used together. The two parameters are cumulated and a `AND` operation is performed between the two filters.
```
MarinPostma, line 669: How would we express a `OR` in the future if we wanted to?

MarinPostma: like I want tasks that are either `enqueued` or `processing`

MarinPostma (commented):
This API feels a bit awkward to me TBH, we have on one end index and uid that are identified as resources, and status and type that are filters. It would be more consistent to have everything be filters, wouldn't it?

text/0060-tasks-api.md
gmourier: I don't think there is a particular need to go that far but it would be possible by separating the values with commas for a `+`.

e.g. `status=enqueued+processing` or `status=enqueued,processing`.

text/0061-error-format-and-definitions.md
gmourier: This is just to say that the markers starting `:` in the error message are actually displayed at the time the error is returned by replacing by the real value of `:status` and `:taskStatuses`.

This is specified on all our error messages, I don't think there's any point in specifying it, maybe it's understandable when reading the error message.

bidoubiwa:
I'm all for removing `/indexes/:index_uid/tasks/:task_uid` as it leads to much to a possible 404. I'm curious about the analytics.

text/0060-tasks-api.md
```diff
@@ -568,6 +568,151 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
...
 
+#### 9. Filtering task resources
+
+The `/tasks` and `index/{indexUid}/tasks` endpoints are filterable by type and status query parameters.
+
+##### 9.1. Query parameters definition
+
+| parameter | type | required | description                         |
+|------|------|----------|----------------------------|
+| status | string | No | Possible values are all the status of a task. By default, when `status` query parameter is not set, all task statuses are returned. |
```
brunoocasali, line 579: Trying to think of possible use cases for this, I got the question:

Should we support the `status` as a list of statuses? Like, what if I wanted to list all the `succeeded`, `failed` at the same time?

We could support the status as an array of strings, or just a list comma-separated.

text/0060-tasks-api.md
```diff
@@ -568,6 +568,151 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
 }
 ```
 
+#### 9. Filtering task resources
+
+The `/tasks` and `index/{indexUid}/tasks` endpoints are filterable by type and status query parameters.
```
brunoocasali, line 573: I know this is not the purpose of this spec, but since we are thinking about adding filters to the `/tasks` resources and our task id is global, why not add a filter to filter by `uid=`/`index=` and remove the `index/{indexUid}/tasks`/`index/{indexUid}/tasks/{id}`? I suppose it could make the development on the core side easier since we will need to handle only two APIs, not four.

brunoocasali (commented):
I left some polemical comments 😬 😅

brunoocasali:
Oh, I think I came late to the party about the potential removal of the `index/uid/tasks`, @MarinPostma and @bidoubiwa already raised the question! 🐞

text/0060-tasks-api.md
MarinPostma: I 100% agree with that! The indexes task routes basically do just that. That's legacy

text/0060-tasks-api.md
gmourier: Hi @brunoocasali, this previous [comment](https://github.com/meilisearch/specifications/pull/116#discussion_r850466987) should answer your response. The spec needs to be updated, it could be could to write `?status=succeeded,failed` or `?status=succeeded+failed`

text/0060-tasks-api.md
curquiza: Same! We discussed this with the core team this morning: v0.28.0 should at least support the filter by index to remove the `/indexes/{uid}/tasks` in v0.28.0 as well. This would avoid a huge breaking in the future.

gmourier:
The specification is now updated with the latest information given in the comments.

curquiza (approved):

PR #117, opened 2022-02-22
Settings API - Typo Tolerance

gmourier:
Naming proposals are open until **2022-03-31**
The definitive specification is due at **2022-04-01**

---

# Summary

This specification describes the configuration options for the typo tolerance settings.

🤖 [API Diff](https://github.com/meilisearch/specifications/pull/117#issuecomment-1075433937)
📄 [Rendered](https://github.com/meilisearch/specifications/blob/typo-tolerance/text/0117-typo-tolerance-setting-api.md)

- [x] Freeze parameter naming choices
- [x] `typo` or `typoTolerance` setting name
- [x] Add dedicated errors for the typo tolerance API resource
- [x] Branch telemetry to the typo tolerance customization
- [x] Update OpenAPI specification
- [x] Add examples for `minWordSizeForTypos`
- [x] Describes Async/Sync errors

Dependent of #123 

## Key Points

- Add a `typoTolerance` index settings resource to manage customization of typo tolerance feature at index level.
    - Expose `typoTolerance` object on `/indexes/:index_uid/settings` endpoints.
    - Add `GET/POST/DELETE` - `/indexes/:index_uid/settings/typo-tolerance` endpoints.
- `typoTolerance` object definition is made of the following properties:
     - `enabled`: Whether the typo tolerance feature is enabled. (Default: `true`)
     - `disableOnAttributes`: Disable the typo tolerance feature on the specified document attributes. (Default: `[]`) 
     - `disableOnWords`: Disable the typo tolerance feature for a set of query terms given during a search query. (Default: `[]`)
     - `minWordSizeForTypos`:
         -  `oneTypo`: Customize the minimum size for a word to tolerate 1 typo (Default: `5`) 
         -  `twoTypos`: Customize the minimum size for a word to tolerate 2 typos (Default: `9`) 
---

# Informations To Reviewers

**We could naturally think of putting the typo tolerance customization in the `rankingRules` typo.**

This seems to be a **bad idea** for several reasons that have been evaluated previously:

- The `rankingRules` become less intuitive by doing multiple things. Currently, its purpose is to configure the ordering of the search results given ranking rules positioning.
- It would probably make reading the ranking rules more difficult if a full `typo` object is described.
- Moreover, if removing `typo` from the ranking rules settings deactivates the typo tolerance feature and the configuration of `typo` properties are located in the ranking rule, a user may lose information if the ranking rule is removed to deactivate it.
- In the future, if we want to expose the typo tolerance properties at search time for customization purposes. It will force us to add the customization of `rankingRules` at search time. Thus, `rankingRules` need to be updated to manage objects (Currently, it's an array of strings). 

As you can see, there are a lot of things that are not desirable if we decide to manage all the typo tolerance properties within the `typo` ranking rule.


bidoubiwa:
Hello! 

> Add a typoTolerance setting to manage customization of typo tolerance feature at index level.

👍 I agree on not adding it in the ranking rules

> enabled: Whether the typo tolerance feature is enabled. (Default: true)
👍

> disableOnAttributes: Disable the typo tolerance feature on the specified document attributes. (Default: [])

What do you think of `disabledAttributes`?

> disableOnWords: Disable the typo tolerance feature for a set of query terms given during a search query. (Default: [])

What do you think of `disabledWords`?

> minWordSizeFor1Typo: Customize the minimum size for a word to tolerate 1 typo. (Default: 5)
> minWordSizeFor2Typos: Customize the minimum size for a word to tolerate 2 typos (Default: 9)

Fiou this is a complex setting 😱 I would not add a number in the setting name. I don't think it is very common to do that. But I'm not sure either. 
I feel like the name is complex but I have no better options to suggest, so until then I agree with this one.


> Force the definition of the whole payload on POST /indexes/:index_uid/settings/typo_tolerance ?

👍

I think we should force the whole payload. I also think this is the expected behavior. This means that in the case the update is partial, the user could potentially think they removed some settings to reset to the default ones but in fact, it will still be the old value.

> Regarding the use of the POST verb, we could change it during v0.28 to PATCH during the API stabilization since it seems to better embrace the REST convention.

In which case, the partial update would make more sense indeed!

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,202 @@
...
+
+- Type: Boolean
+- Required: False
+- Default: `true`
+
+Whether the typo tolerance feature is enabled.
+
+##### 3.1.1.1. Impacts on the `typo` ranking rule
+
+The presence of `typo` in the ranking rules setting does not influence the activation/deactivation of the typo tolerance feature.
```
bidoubiwa, line 38: Im not sure I understood this sentence correctly :s

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,202 @@
...
+
+- Type: Array of String
+- Required: False
+- Default: `[]`
+
+`disableOnWords` disable the typo tolerance feature for a set of search query terms.
+
+#### 3.1.3.4. Example
+
+If `Javascript` is specified in `disableOnWords`, the engine won't apply the typo tolerance on the query term `Javascript` if typed at search time.
```
bidoubiwa, line 64: 🤨 of course you chose javascript

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,202 @@
...
+
+- Type: Integer
+- Required: False
+- Default: `9`
+
+`minWordSizeFor2Typos` customize the minimum size for a word to tolerate 2 typos.
+
+> Given the default value `9`, the search engine handles up to 2 typos on a query term if its length is at least equal to 9 characters.
+
+> See [2.1. Typos calculation section](#21-typos-calculation)
```
bidoubiwa, line 88: How are `3` and `4` typos impacted by that?
Lets say I set `minWordSizeFor2Typos` to `10`, will it set `10` for `3` and more typos as well?

gmourier: Meilisearch does not tolerate typos `>2` 🤓 It is not planned that Meilisearch will tolerate more typos but this may be a future possibility.

Let me know if you need more details about that!

text/0117-typo-tolerance-settings-api.md
gmourier: Basically, if the `typo` ranking rule is removed from the setting, the engine will still be able to match documents containing typo errors against the query term. It will just not rely on the number of typos to rank the documents.

It it clearer this way?

bidoubiwa: So it would still be possible to receive 
`helol` instead of `hello` even when removing `typo` from the ranking rules? But it will be in any random order?

gmourier:
@bidoubiwa I have no preference between `disableOnAttributes`/`disableOnWords` or `disabledAttributes/disabledWords`. Both work for me!

text/0117-typo-tolerance-settings-api.md
gmourier: Yes, the order will of course be defined by all other ranking rules.

brunoocasali (commented):
Hi @gmourier this is my first review, I will learn more about the whole typo matching feature in order to help more regarding this config 😉

open-api.yaml
```diff
@@ -400,6 +400,39 @@ components:
...
+          description: Enable the typo tolerance feature.
+          type: boolean
+          default: true
+          nullable: false
+        disableOnAttributes:
+          description: Disable the typo tolerance feature on the specified attributes.
+          type: array
+          items:
+            type: string
+          default: []
```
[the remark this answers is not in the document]
bidoubiwa: The issue is that the user might add a document with the field `surname` in the future as Meilisearch does not impose a document structure

text/0117-typo-tolerance-settings-api.md
bidoubiwa: ah I see! Thanks :)

text/0117-typo-tolerance-settings-api.md
bidoubiwa: I understand! Thanks :)

open-api.yaml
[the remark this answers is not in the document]
gmourier: Another issue is that the update is processed asynchronously, so if there is an error the message will be within a task. Which is better than nothing but not ideal to help the user directly.

open-api.yaml
```diff
@@ -1652,6 +1691,15 @@ paths:
                       - price
                       - author
                       - title
+                    typoTolerance:
+                      enabled: true
+                      disableOnAttributes:
+                        - author
+                        - price
+                      disableOnWords:
+                        - Tolkien
```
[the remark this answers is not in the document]
gmourier: Oh! About `Tolkien` here or `disableOnWords` ?

open-api.yaml
```diff
@@ -2049,6 +2108,74 @@ paths:
           description: Not Found
     parameters:
       - $ref: '#/components/parameters/indexUid'
+  '/indexes/{indexUid}/settings/typo-tolerance':
```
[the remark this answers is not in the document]
gmourier: We don't have a global configuration except for the instance options today but it's clear that it could be an interesting future possibility!

open-api.yaml
```diff
@@ -2049,6 +2108,74 @@ paths:
...
+          description: Ok
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/typoTolerance'
+        '401':
+          $ref: '#/components/responses/401'
+        '404':
+          description: Not Found
+    post:
```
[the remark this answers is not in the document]
gmourier: There is no PUT for the moment.

In fact, PUT requires all fields to be defined to modify the resource. While a PATCH accepts a partial update.

If we add PUT, I think that POST has no more sense because the resource already exists. We don't create anything here.

In the idea, it is either POST or PUT (All fields are required in the payload) + PATCH (Partial updates are allowed).

Today on `/settings` for example, POST allows making a partial update. 

I think the cleanest way would be to replace POST with PUT (require all fields on it) and add a PATCH route for partial udpates but that is out-of-scope for this specification. I would like to think about that for v0.28!

text/0034-telemetry-policies.md
```diff
@@ -114,7 +115,12 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
 | `sortable_attributes.has_geo`           | `true` if `_geo` is set as a sortable attribute, otherwise `false` | true | `Settings Updated`, `SortableAttributes Updated` |
 | `filterable_attributes.total`           | Number of filterable attributes | 3 | `Settings Updated`, `FilterableAttributes Updated` |
 | `filterable_attributes.has_geo`         | `true` if `_geo` is set as a filterable attribute, otherwise `false` | false | `Settings Updated`, `FilterableAttributes Updated`|
-| `searchable_attributes.total`           | Number of searchable attributes | 4 | `Settings Updated`, `SearchableAttributes Update` |
+| `searchable_attributes.total`           | Number of searchable attributes | 4 | `Settings Updated`, `SearchableAttributes Updated` |
+| `typo_tolerance.enabled`                 | Whether the typo tolerance is enabled | `true` | `Settings Updated`, `TypoTolerance Updated` |
+| `typo_tolerance.disable_on_attributes_used    | `true` if at least one value is defined | `false` | `Settings Updated`, `TypoTolerance Updated` |
+| `typo_tolerance.disable_on_words_used   | `true` if at least one value is defined | `false` | `Settings Updated`,  `TypoTolerance Updated` |
```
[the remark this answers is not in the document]
gmourier: The specification of the typo tolerance allows to find this information, no? I would like to avoid duplicating information. Maybe some links would do the trick here!

text/0034-telemetry-policies.md
```diff
@@ -114,7 +115,12 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
...
+| `typo_tolerance.min_word_size_for_1_typo`     | The defined value for `minWordSizeFor1Typo` property | `5` | `Settings Updated`, `TypoTolerance Updated` |
+| `typo_tolerance.min_Word_size_for_2_typos`    | The defined value for `minWordSizeFor2Typos` property | `9` | `Settings Updated`, `TypoTolerance Updated` |
```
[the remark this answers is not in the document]
gmourier: Nope! It's a miss. 🥲

text/0061-error-format-and-definitions.md
```diff
@@ -222,6 +222,113 @@ HTTP Code: `400 Bad Request`
...
+
+### Context
+
+This error happens when the `disableOnAttributes` field of `typoTolerance` resource is invalid.
+
+### Error Definition
+
+```json
+{
+    "message": "`disableOnAttributes` field value `:value` is invalid. It should be an array of string representing document attributes.",
```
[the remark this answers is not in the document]
gmourier: I like it!

text/0061-error-format-and-definitions.md
```diff
@@ -222,6 +222,113 @@ HTTP Code: `400 Bad Request`
...
+
+### Context
+
+This error happens when the `minWordSizeFor2Typos` field of `typoTolerance` resource is invalid.
+
+### Error Definition
+
+```json
+{
+    "message": "`minWordSizeFor2Typos` field value `:value` is invalid. It should be an integer superior to 0 and superior or equals to `minWordSizeFor1Typo`.",
```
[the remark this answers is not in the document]
gmourier: I will do it! Don't hesitate to suggest it directly next time so I can add it with one click. 🏃

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,202 @@
...
+- Required: False
+- Default: `[]`
+
+`disableOnWords` disable the typo tolerance feature on a list of search query terms.
+
+#### 3.1.3.4. Example
+
+If `Javascript` is specified in `disableOnWords`, the engine won't apply the typo tolerance on the query term `Javascript` if typed at search time.
+
+### 3.1.4. `minWordSizeFor1Typo`
```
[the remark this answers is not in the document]
gmourier: Sure!

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,202 @@
...
+
+> See [2.1. Typos calculation section](#21-typos-calculation)
+
+## 3.2. API Endpoints Definition
+
+### 3.2.1. `indexes/:index_uid/settings/typo-tolerance`
+
+Manage the typo tolerance configuration for an index.
+
+#### 3.2.1.1. `GET` - `indexes/:index_uid/settings/typo-tolerance`
```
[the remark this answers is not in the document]
gmourier: Yes, I will use this notation now because it is the one used in the documentation. In itself, it's not a big deal but it's good to be consistent 😀

open-api.yaml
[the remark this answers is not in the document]
brunoocasali: I bet the configuration field `disableOnWords` is. But my question is regarding the data inside of it 😬

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,202 @@
...
+    "type": "settingsUpdate",
+    "enqueuedAt": "2022-03-01T18:39:29.228155Z"
+}
+```
+
+> Returns a 202 response. See [Summarized `task` Object for `202 Accepted`](0060-tasks-api.md#summarized-task-object-for-202-accepted)
+
+##### 3.2.1.2.1. Errors
+
+- 🔴 If the index does not exist, the API returns an [index_not_found](0061-error-format-and-definitions.md#index_not_found) error.
```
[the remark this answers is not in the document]
gmourier: This one is tricky. It will depend on if the instance is secure and if the API key allows the operation or not. I'll look into it to explain it better but it will add a lot of information. 

I've added that kind of details in https://github.com/meilisearch/specifications/pull/123/files. I will backport those details here when #123 is merged!

text/0118-search-api.md
```diff
@@ -18,14 +18,14 @@ Meilisearch exposes 2 routes to perform searches:
 
 - 🔴 If the index does not exist, the API returns an [index_not_found](0061-error-format-and-definitions.md#index_not_found) error.
 
-If the instance is secured by a master-key, the auth layer will return the following errors:
+If the instance is secured by a master-key, the auth layer returns the following errors:
```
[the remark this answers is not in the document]
gmourier: I would love to be able to continue to make the specifications modular! As it's markdown, I already started to put more and more cross-link ref to avoid re-explaining things, it could work for this kind of error (Auth related)!

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,202 @@
...
+### 2.2. Typo ranking rule
+
+TODO: Describes the impact of the typo ranking rule regarding search results ranking.
+
+## 3. Future Possibilities
+
+- Expose `typoTolerance` resource as a search parameter to override index settings.
+- Add the possibility to disable the typo tolerance feature on all numeric fields.
+- Add different modes of result matching for the typo-tolerance feature. e.g. `default`/`min`/`strict`
+- Replace `POST` to `PATCH` verb to allow partial edit of the settings and embrace REST API convention.
```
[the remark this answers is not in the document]
gmourier: We have planned to stabilize the APIs for v0.28.0, I think it can be done at that time if there is a decision to do so.

open-api.yaml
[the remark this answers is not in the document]
gmourier: I think it cannot be case sensitive because internally the engine lowercase and deunicode all strings in documents. @ManyTheFish Correct me if I'm wrong 🤓

I will add it!

text/0118-search-api.md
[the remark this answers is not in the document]
gmourier: Oh okay, you mean the changes to the Search API spec?

text/0117-typo-tolerance-settings-api.md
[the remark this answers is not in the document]
gmourier: I've backported the details in the specification.

text/0034-telemetry-policies.md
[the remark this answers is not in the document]
gmourier: Fixed!

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,228 @@
...
+
+### 3.1.2. `disableOnAttributes`
+
+- Type: Array of String
+- Required: False
+- Default: `[]`
+
+`disableOnAttributes` disable the typo tolerance feature on the specified attributes list.
+
+Any attributes defined in `disableOnAttributes` won't have their values matched by the typo tolerance.
```
ManyTheFish, line 52: this part is really ambitious, needing changes in both search and indexing parts.
Moreover, it could create conflict with other works on the indexer:
- nested fields
- optimizations

I suggest making a discovery on this specific part.

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,228 @@
...
+
+- Type: Integer
+- Required: False
+- Default: `9`
+
+`minWordSizeFor2Typos` customize the minimum size for a word to tolerate 2 typos.
+
+> Given the default value `9`, the search engine handles up to 2 typos on a query term if its length is at least equal to 9 characters.
+
+> See [2.1. Typos calculation section](#21-typos-calculation)
```
ManyTheFish, line 88: This 3 features should only impact the query tree 👍

alallema (commented):
Well Done !!! 🔥 
>If we choose typoTolerance, should we change the typo ranking rule by typoTolerance ?

I like `typoTolerance` but if we leave just `typo` in the rankings rules won't it be confusing? Too bad because it's my favorite solution!

text/0117-typo-tolerance-settings-api.md
```diff
@@ -0,0 +1,228 @@
...
+    "disableOnAttributes": [
+        "title",
+        "description"
+    ],
+    "disableOnWords": [
+        "Tolkien"
+    ],
+    "minWordSizeFor1Typo": 4,
+    "minWordSizeFor2Typos": 8
+}
```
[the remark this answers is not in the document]
gmourier: Partially would not reset `minSizeFor1Typo` in your example. Touched fields are the one given in the payload.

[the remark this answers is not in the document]
alallema: I prefer a partial update as you propose, it seems much easier.

text/0117-typo-tolerance-settings-api.md
gmourier: @MarinPostma If it helps!

text/0117-typo-tolerance-settings-api.md
gmourier: @MarinPostma If it helps!

bidoubiwa:
What did you decide about the naming of
- disabledAttributes vs disableOnAttributes
- disabledWords vs disableOnWords

gmourier:
> What did you decide about the naming of
> 
> * disabledAttributes vs disableOnAttributes
> * disabledWords vs disableOnWords

I modified the spec to use `typo` instead of `typoTolerance` or `typo-tolerance`.

The specification also includes `disabledAttributes` and `disabledWords` in replacement of `disableOnWords` / `disableOnAttributes`

bidoubiwa:
Just out of curiosity, but I'm not sure I prefer it, since `searchableAttributes` is plural because it can be about multiple attributes, `typos` is also a possibility. What do you think?

MarinPostma:
just thinking semantic here, but if we name it `typo`, then it makes more sense to rename `enabled` to `allowed` because we want to enable typo tolerance or allow typos. In that sense, I agree with Charlotte that if we go with `typo` we could probably use `typos` instead.

From what I understand of what you said, @gmourier, you want to unify the ranking rule with the setting

alallema (dismissed):
LGTM! Good job 🔥

text/0061-error-format-and-definitions.md
```diff
@@ -222,6 +222,119 @@ HTTP Code: `400 Bad Request`
...
+### Error Definition
+
+```json
+{
+    "message": "`minWordSizeFor2Typos` field value `:value` is invalid. It should be an integer superior to 0 and superior or equals to the value defined in the field `minWordSizeFor1Typo`.",
+    "code": "invalid_typo_min_word_size_for_2_typos",
+    "type": "invalid_request",
+    "link": "https://docs.meilisearch.com/errors#invalid_typo_min_word_size_for_2_typos"
+}
+```
```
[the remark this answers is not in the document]
MarinPostma: in fact they are so linked that maybe they should be a single setting?

```
minWordLenForTypo: {
 "1-typo": 5,
"2-typos: 9,
}
```
or something similar?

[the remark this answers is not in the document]
bidoubiwa: As it does create more nesting of settings, it may be more readable and clear in what it does.

text/0061-error-format-and-definitions.md
```diff
@@ -222,6 +222,119 @@ HTTP Code: `400 Bad Request`
...
+
+### Context
+
+This error happens when the `minWordSizeFor2Typos` field of `typo` resource is invalid.
+
+### Error Definition
+
+```json
+{
+    "message": "`minWordSizeFor2Typos` field value `:value` is invalid. It should be an integer superior to 0 and superior or equals to the value defined in the field `minWordSizeFor1Typo`.",
```
bidoubiwa, line 327:
```suggestion
    "message": "`minWordSizeFor2Typos` field value `:value` is invalid. It should be an integer superior to 0 and superior or equal to the value defined in the field `minWordSizeFor1Typo`.",
```

text/0061-error-format-and-definitions.md
```diff
@@ -222,6 +222,119 @@ HTTP Code: `400 Bad Request`
...
+
+### Context
+
+This error happens when the `minWordSizeFor1Typo` field of `typo` resource is invalid.
+
+### Error Definition
+
+```json
+{
+    "message": "`minWordSizeFor1Typo` field value `:value` is invalid. It should be an integer superior to `0` and inferior or equals to the value defined in the field `minWordSizeFor2Typos`.",
```
bidoubiwa, line 304:
```suggestion
    "message": "`minWordSizeFor1Typo` field value `:value` is invalid. It should be an integer superior to `0` and inferior or equal to the value defined in the field `minWordSizeFor2Typos`.",
```

bidoubiwa (changes requested):

text/0061-error-format-and-definitions.md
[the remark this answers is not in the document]
gmourier: Thanks @MarinPostma I'm introducing it into the specification

text/0118-search-api.md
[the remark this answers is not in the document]
brunoocasali: Yes! :D Since this spec talks about just the typo tolerance, and this change is related to search :)

gmourier:
@alallema, @bidoubiwa, @MarinPostma

While integrating the idea of nested fields for `minWordSizeForTypos`, I want to have your opinion on which is the best format.

```json
{
    "minWordSizeForTypos": {
        "1Typo": 5,
        "2Typos": 9
    }
}
```

or 

{
    "minWordSizeForTypos": {
        "1": 5,
        "2": 9
    }
}

Thanks 🙏

bidoubiwa:
If it is a string I don't see an issue in my scopes. But I don't know about the second one as it is not clear if the attribute is the word or the typo numbers.

alallema:
> If it is a string I don't see an issue in my scopes. But I don't know about the second one as it is not clear if the attribute is the word or the typo numbers.

Same I don't see any issue for the `JSON` but just the numbers I find not very clear either no? Can I ask why not in letter:
```
{
    "minWordSizeForTypos": {
        "oneTypo": 5,
        "twoTypos": 9
    }
}
```
I feel it more clear and also it's avoid mistake like `1typo`, `1Typo`.

text/0034-telemetry-policies.md
```diff
@@ -317,6 +329,17 @@ This property allows us to gather essential information to better understand on
...
 | searchable_attributes.total   | Number of searchable attributes. | `3` |
 
+
+## `TypoTolerance Updated`
+
+| Property name | Description | Example |
+|---------------|-------------|---------|
+| typo.enabled        | Whether the typo tolerance is enable.d | `true` |
+| typo.disabled_attributes | `true` if at least one value is defined for `disabledAttributes` property. | `false` |
+| typo.disabled__words    | `true` if at least one value is defined for `disabledWords` property. | `false` |
```
MarinPostma, line 339:
```suggestion
| typo.disabled_words    | `true` if at least one value is defined for `disabledWords` property. | `false` |
```

text/0117-typo-setting-api.md
```diff
@@ -0,0 +1,295 @@
...
+
+The size of a query term triggers the typo tolerance feature.
+
+By default, a query term made of `5` characters can tolerate `1` typo, while a query term made of `9` characters can tolerate `2` typos.
+
+If a query term can be derived from a distance of one or two typos to be found in a document, it will be selected as a candidate. The number of typos found has an impact on the relevancy.
+
+#### 2.1.1 Typo On The First Character
+
+A query term having the first character considered as a typo will not count as `1` typo but as `2` typos.
```
maryamsulemani97, line 276: - Just to be clear: if I search for `qorld` instead of `world`, the `q` will be two typos? 
- How will this work if, e.g., my settings only allow for 1 typo per 5 characters, will I not get any results?

text/0117-typo-setting-api.md
```diff
@@ -0,0 +1,295 @@
...
+
+It permits to drastically reduce the search space in this case, thus speed-up the search response.
+
+#### 2.1.2. Clamping Typo On Concatened Query Terms
+
+The concatenation of two query terms is equal to 1 typo.
+
+e.g. If `Le tableau` is typed,  the concatened string `Letableau` made of `9` letters can only have `1` typo instead of `2` since `1` typo is already counted by the concatenation.
```
maryamsulemani97, line 284: Will this rule apply if the query was `Le-tableau` or `Le,tableau`?

text/0117-typo-setting-api.md
```diff
@@ -0,0 +1,295 @@
...
+| Field                      | Type            | Required |
+|----------------------------|-----------------|----------|
+| [oneTypo](#3142-onetypo)   | Integer         | False    |
+| [twoTypos](#3143-twotypos) | Integer         | False    |
+
+#### 3.1.4.2. `oneTypo`
+
+- Type: Integer
+- Required: False
+- Default: `5`
```
maryamsulemani97, line 95: Is the default only when `enabled` is set to `true?`
Or does it apply even when typo tolerance is disabled?

text/0117-typo-setting-api.md
```diff
@@ -0,0 +1,295 @@
...
+
+#### 3.1.1. `enabled`
+
+- Type: Boolean
+- Required: False
+- Default: `true`
+
+Whether the typo tolerance feature is enabled.
+
+##### 3.1.1.1. Impacts on the `typo` ranking rule
```
maryamsulemani97, line 32: 1. As long as the `typo` rule is part of the `rankingRules` array, it will be applied regardless of if typo tolerance is enabled or not, correct? 
2. Not sure this even makes sense, but how is the `typo` ranking rule applied if you enable typo tolerance? Will it work the same way but match fewer documents?

text/0117-typo-setting-api.md
Kerollmops: Yes, it will as `-` and `,` are stop-words like ` `.

text/0117-typo-setting-api.md
Kerollmops: > 1. Just to be clear: if I search for `qorld` instead of `world`, the `q` will be two typos?

Indeed, the engine will only return documents that contains `world` after documents containing `qoyld` for example (1 typo).

> 2. How will this work if, e.g., my settings only allow for 1 typo per 5 characters, will I not get any results?

Exactly, the engine will not try to find worlds with 2 typos on a word that is smaller than 5 letters, a typo on the first letter is considered 2 therefore you will never find documents with `world` for example.

text/0117-typo-setting-api.md
gmourier: Thanks, @maryamsulemani97 and @Kerollmops ! I will precise it in the specification.

text/0117-typo-setting-api.md
gmourier: When typo tolerance is not enabled, the other properties are modifiable but this will have no impact on the search since typo tolerance is globally disabled.

So yes, `disabledAttributes`, `disabledWords`, `minWordSizeForTypos` fields will be taken into account at search when `enabled` is `true`.

text/0117-typo-setting-api.md
gmourier: The presence of `typo` in the `rankingRules` does not enable or disable the typo tolerance feature.

If `enabled` is false (the typo-tolerance feature is not used to match documents), the `typo` ranking rules have no impact.
if `enabled` is true (the typo-tolerance feature is used to match documents containing typos), the `typo` ranking rules will rank documents according to the number of typos found (`0` the better). If the `typo` rule is not defined in the `rankingRules` settings, documents containing typos will be matched but the `typo` ranking rule will not be used to sort them.

Does it help? Don't hesitate to tell me if I can be more precise. 😇

maryamsulemani97:
Naming suggestions:

1. `typo` object
**Suggestion**: `typoTolerance`
- Users won’t confuse it with the `typo` ranking rule

2. `disabledWords`
**Suggestions**:
- `typoFreeWords`: Sounds more intuitive.  `disabledWords` made me think these words won't be included in the search
- `reservedWords`: I don’t like this because this makes it sound like you can’t use these words for anything. They’re reserved for Meilisearch’s use or something

3. `disabledAttributes`
**Suggestions**: 
- `typoFreeAttributes`: Sounds more intuitive and goes with `typoFreeWords`
- `reservedAttributes`

Lmk what you think

@meilisearch/docs-team In case you guys have any suggestions

bidoubiwa:
> typo object
> Suggestion: typoTolerance

I agree !

> typoFreeWords

I'm not sure I understood what `typoFreeWords` does when reading it. It seems like you are giving a list of words that never have typos? Maybe it is a good suggestion though, I may need to let the idea marinate.  
Same for `typeFreeAttributes`.

MarinPostma:
internally I call them `exact_words` and `exact_attributes`, as a more positive naming

maryamsulemani97:
>I'm not sure I understood what typoFreeWords does when reading it. It seems like you are giving a list of words that never have typos?

`typoFreeWords`: You're saying I don't want any typos matching these words. In a way, you never want to make any typos when searching for them. Does that make sense?

If we want something super specific, we could go for:
- `disableTypoToleranceOnWords` /  `disableTypoToleranceOnAttributes`
- `wordsWithoutTypoTolerance` / `atributesWithoutTypoTolerance`

bidoubiwa:
`disableTypoToleranceOnWords` is very explicit 😂 While it might be a bit long it might compensate with the very clear usage it offers. 
To win a few chars `wordsWithoutTypoTolerance / atributesWithoutTypoTolerance` is good as well. 
Now, if the setting is called `typo tolerance` instead of `typo` it brings context to the variables. So it might be overkill (it might). 

I have a really hard time deciding which one I prefer. @gmourier do you have a preference? 

I order of preference I would say: 

- wordsWithoutTypoTolerance / atributesWithoutTypoTolerance
- disableWords / disableAttributes (if the name of the setting is `typo tolerance`. if it is just `typo` then no)
- typoFreeAttributes / typoFreeAttributes
- exactWords / exactAttributes

text/0117-typo-setting-api.md
```diff
@@ -0,0 +1,295 @@
...
+Given `3` for `oneTypo` and `5` as `twoTypos` and the following document
+
+```json
+{
+    "id": 0,
+    "title": "Hey World"
+}
+```
+
+- Typing `World` with 2 typos, e.g. `Warrld` will match `World`. It accepts 1 typo up to 2 typos since `World` size is made of `5` chars.
```
dichotommy, line 140:
```suggestion
- Typing `World` with 2 typos, e.g. `Warrld` will match `World`. It accepts up to 2 typos since `World` size is made of `5` chars.
```

text/0117-typo-setting-api.md
```diff
@@ -0,0 +1,295 @@
...
+
+### 2.2. Typo Ranking Rule
+
+The `typo` ranking rule favors candidates with the least typos. That is, if a document is found with two typos, it will be ranked as less important than a document found with 1 typo. 0 typo is the most favorable case.
+
+## 3. Future Possibilities
+
+- Expose `typo` resource as a search parameter to override index settings.
+- Add the possibility to disable the typo tolerance feature on all numeric fields.
+- Add different modes of result matching for the typo feature. e.g. `default`/`min`/`strict`
```
dichotommy, line 294: > Add different modes of result matching for the typo feature. e.g. `default`/`min`/`strict`

Great suggestion

gmourier:
Signal @meilisearch/docs-team and @meilisearch/integration-team 📡

Dedicated error `invalid_typo_{fieldName}` for type checking I've been removed to be replaced by a `bad_request` error. This is the current behavior for all other settings.

We kept the asynchronous `invalid_typo_minWordSizeForTypos`.

This is noted as a future possibility for the simple reason that we should explore the possibility of having a reusable validation component within the core codebase. I naturally thought that it would be easier than expected because it is already implemented for API Keys but it requires industrializing the code to be reusable everywhere.

text/0117-typo-setting-api.md
```diff
@@ -0,0 +1,296 @@
...
+| Field                      | Type            | Required |
+|----------------------------|-----------------|----------|
+| [oneTypo](#3142-onetypo)   | Integer         | False    |
+| [twoTypos](#3143-twotypos) | Integer         | False    |
+
+#### 3.1.4.2. `oneTypo`
+
+- Type: Integer
+- Required: False
+- Default: `5`
```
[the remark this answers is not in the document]
gmourier: That's a good point @brunoocasali, I asked Marin's opinion on this. 

The error message is like this today:

`minWordSizeForTypos` setting is invalid. `oneTypo` and `twoTypos` fields should be between `0` and `255`, and `twoTypos` should be greater or equals to `oneTypo` but found `oneTypo: :oneTypo` and `twoTypos: twoTypos`.

I come back to you when I get the behavior of what happens when `0` is set, I naturally imagine that it will disable tolerating 1 typo. If this is the case it's a bit strange since the engine could only accept 2 typos for example but I don't think it makes sense and can be done internally since to reach 2 typos you first need to reach 1 typo.

dichotommy:
Hello! If you'll forgive the last-minute intrusion, I wanted to say that I agree w/ @maryamsulemani97's suggestion to change the name of the object from `typo` to `typoTolerance`. 

The main reason is that this name is clearer about the purpose of the object. On a purely aesthetic level, `typoTolerance` also follows better the naming convention for Settings, all of which are two words except `synonyms`.

> We were afraid that it can be confusing to have 2 separate terms and we preferred to have the same term for the typo ranking rules and the settings because they are related thus using typo instead of typoTolerance to not rename the ranking rule as well and avoid a breaking change.

I do not think the setting should have the same name as the ranking rule as they do different things: one is selecting search results, the other is ordering them. Yes they are related, but I think that’s clear even if one is called `typoTolerance` and the other is called `typo`.

text/0117-typo-setting-api.md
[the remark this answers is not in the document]
gmourier: @brunoocasali, after consulting @MarinPostma we decided to not force `oneTypo` to be superior to `1` since there is no direct use case to block it. In reality, a word of `1` does not exist as a word of `0` it is, therefore, equivalent in the engine.

However, as having too small or too large values can impact relevancy and performance we would like the @meilisearch/docs-team to mention this type of information: 

For `oneTypo`:
A very small value means you'll get a lot of false-positive results.
A very high value means the typo-tolerance feature of the search engine will rarely be used.

For  `twoTypos`:
A very small value means you'll get a lot of false-positive results.
A very high value means the typo-tolerance feature of the search engine will rarely match two typos.

Algolia does this in its interface which seems smart enough to advise users without blocking them.

Also pinging the @meilisearch/cloud-team that can integrate it into the UI on their side at some point.

gmourier:
Thanks, @dichotommy!

We will follow this recommendation.

Also we choose `disableOnWords` / `disableOnAttributes` since after a discussion with @dichotommy it makes more sense than `disabledWords`/ `disabledAttributes`.

📡  Signal to @MarinPostma

**Naming is now frozen until the release.**

bidoubiwa (dismissed):
LGTM 🍍🍍 Great Spec ✨

text/0034-telemetry-policies.md
```diff
@@ -116,7 +117,12 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
...
 | `filterable_attributes.has_geo`         | `true` if `_geo` is set as a filterable attribute, otherwise `false` | false | `Settings Updated`, `FilterableAttributes Updated`|
-| `searchable_attributes.total`           | Number of searchable attributes | 4 | `Settings Updated`, `SearchableAttributes Update` |
+| `searchable_attributes.total`           | Number of searchable attributes | 4 | `Settings Updated`, `SearchableAttributes Updated` |
+| `typo.enabled`                          | Whether the typo tolerance is enabled | `true` | `Settings Updated`, `TypoTolerance Updated` |
+| `typo.disabled_attributes`              | `true` if at least one value is defined | `false` | `Settings Updated`, `TypoTolerance Updated` |
+| `typo.disabled_words`                   | `true` if at least one value is defined | `false` | `Settings Updated`,  `TypoTolerance Updated` |
+| `typo.min_word_size_for_typos.one_typo` | The defined value for `minWordSizeForTypos.oneTypo` property | `5` | `Settings Updated`, `TypoTolerance Updated` |
+| `typo.min_word_size_for_typos.two_typos`| The defined value for `minWordSizeForTypos.twoTypos` property | `9` | `Settings Updated`, `TypoTolerance Updated` |
```
bidoubiwa, line 125: is this still `typo` or does it become `typoTolerance.enabled` (etc..) ?

gmourier: I do not understand how it can happens since the last https://github.com/meilisearch/specifications/pull/117/commits/276cab452733def6db5c868f9e25fa0d5dc802dd change that.

open-api.yaml
```diff
@@ -405,6 +405,42 @@ components:
...
+          default: []
+          nullable: false
+        disableOnWords:
+          description: Disable the typo tolerance feature for a set of query terms given for a search query.
+          type: array
+          items:
+            type: string
+          default: []
+          nullable: false
+        minWordSizeForTypos:
```
maryamsulemani97, line 431: @gmourier,  

I tried the v0.27rc and I see `minWordLengthForTypo`. Wanted to confirm what name we're going with

Screen Shot 2022-04-15 at 00 51 01

MarinPostma: Uhm no you're right, needs to be changed

MarinPostma: I'll open a pr

text/0034-telemetry-policies.md
```diff
@@ -54,6 +54,7 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
 | RankingRules Updated | Occurs when ranking rules are updated via `POST` - `/indexes/:indexUid/settings/ranking-rules`. |
 | FilterableAttributes Updated | Occurs when filterable attributes are updated via `POST` - `/indexes/:indexUid/settings/filterable-attributes`. |
 | SortableAttributes Updated | Occurs when sortable attributes are updated via `POST` - `/indexes/:indexUid/settings/sortable-attributes`. |
+| TypoTolerance Updated | Occurs when typo tolerance settings are updated via `POST` - `/indexes/:indexUid/settings/typo`. |
```
gmourier, line 57:
```suggestion
| TypoTolerance Updated | Occurs when typo tolerance settings are updated via `POST` - `/indexes/:indexUid/settings/typo-tolerance`. |
```

PR #120, opened 2022-03-01
Search API - Formatting Search Results

gmourier:
Naming proposals are open until **2022-03-31**
The definitive specification is due at **2022-04-01**

---

# Summary

This specification describes the parameters that can be used to format search results with Meilisearch in a native way. 

_It also deletes an old specification dedicated to the `_formatted` field and reorganizes the sections of the specification concerning search endpoints._

🤖  [API Diff](https://github.com/meilisearch/specifications/pull/120#issuecomment-1062865154)
📄 [Rendered](https://github.com/meilisearch/specifications/blob/formatting-search-results/text/0118-search-api.md)

---

# Key Points

### Customize Highlight Tags

- `highlightPreTag`: (Type: String, Required: `false`, Default: `"<em>"`)
- `highlightPostTag`: (Type: String, Required: `false`, Default: `"</em>"`)

### Customize Default Crop Marker

- `cropMarker`: (Type: String, Required: `false`, Default: `…`) 

### Crop around **words** instead of **characters**

- `cropLength` is now a **number of words**.
- The `cropMarker` is affixed "smartly" around the cropped part of an attribute.

### Cropped part selection

The cropping algorithm tries to match the window with the highest density of query terms within the `cropLength` limit. Then it will pick the window that contains the more ordered query terms.

### Unchanged

- `matches` and `_matchesInfo` remain unchanged.

### Future Possibilities

- All parameters related to the formatting of search results could be under a `formatter` object.
- `formatter` could be exposed as an index setting to be overridable at search time.

---

# Informations To Reviewers

### Naming

Consulting @meilisearch/integration-team and @meilisearch/docs-team about the naming for the new parameters.

- `highlightPreTag`
- `highlightPostTag`
- `cropMarker`

---

## Misc

- [x] Describes future possibilities
- [x] Merge filter-and-facet-behavior spec into dedicated field description of the search spec


irevoire (dismissed):

text/0118-search-api.md
```diff
@@ -108,156 +111,273 @@ It returns the number of documents matching the current search query for each gi
...
+Specifying `cropMarker` to `""` or `null` implies that no marker will be applied to the cropped part, if any.
 
-##### 1.2.1.11 `matches`
+- 🔴 Sending a value with a different type than `String` or `null` for `cropMarker` returns a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
+
+##### 3.1.1.13.1. Applying `cropMarker`
+
+###### 3.1.1.13.1.1. Matched Part To Be Cropped
+
+The cropping algorithm tries to match the window with the highest density of query terms within the `cropLength` limit. Then it will pick the window that contains the more ordered query terms.
```
curquiza, line 269: Not sure to understand, especially the "Then it will pick the window that contains the more ordered query terms". Maybe an example would help :)

text/0118-search-api.md
```diff
@@ -108,156 +111,273 @@ It returns the number of documents matching the current search query for each gi
...
+If the cropped part has been matched against query terms and contains the end of the attribute to be cropped, the `cropMarker` is not placed to the right of the cropped part.
+
+#### 3.1.1.14. `matches`
 
 - Type: Boolean
 - Required: False
 - Default: `false`
 
-Adds a `_matchesInfo` object to the search response that contains the location of each occurrence of queried terms across all fields. This is useful when more control is needed than offered by the built-in highlighting/cropping features.
+Adds a `_matchesInfo` object to the search response that contains the location of each occurrence of queried terms across all fields. It's useful when more control is needed than offered by the built-in highlighting/cropping features.
```
curquiza, line 287: Maybe precise this is the number of bytes

text/0118-search-api.md
```diff
@@ -108,156 +111,273 @@ It returns the number of documents matching the current search query for each gi
...
-| hits                    | Array[Hit]                   | True     |
-| limit                   | Integer                      | True     |
-| offset                  | Integer                      | True     |
-| nbHits                  | Integer                      | True     |
-| exhaustiveNbHits        | Boolean                      | True     |
-| facetsDistribution      | Object                       | False    |
-| exhaustiveFacetsCount   | Boolean                      | False    |
-| processingTimeMs        | Integer                      | True     |
-| query                   | String                       | True     |
+| Field                                                 | Type                         | Required |
```
curquiza, line 293: What does it mean "required" for a search response?
Do you mean "present by default"?

text/0118-search-api.md
curquiza: Is it something we would consider to change?

text/0118-search-api.md
```diff
@@ -108,156 +111,273 @@ It returns the number of documents matching the current search query for each gi
...
+Search queries using `_geoPoint` returns a `_geoDistance` field containing the distance in meters between the document `_geo` coordinates and the specified `_geoPoint`.
+
+> See [GeoSearch](0059-geo-search.md)
+
+###### 3.2.1.1.2. `_formatted`
+
+- Type: Object
+- Required: False
+
+`_formatted` returns highlighted and cropped attributes specified in `attributesToHighlight` and/or `attributesToCrop` of a search result.
```
curquiza, line 340: If I understand correctly, only the `attributesToHighlight` and `attributesToCrop` are put in `_formatted`. If `*` is set as `attributesToHighlight` or `attributesToCrop`, _formatted will return `attributesToRetrieve`?

text/0118-search-api.md
```diff
@@ -108,156 +111,273 @@ It returns the number of documents matching the current search query for each gi
...
-- Default: `200`
+- Default: `10`
 
-Configures the number of characters to keep on each side of the matching query term when using the `attributesToCrop` parameter.
+Sets the total number of **words** to keep around the matched part of an attribute specified in the `attributesToCrop` parameter.
 
 If `attributesToCrop` is not configured, `cropLength` has no effect on the returned results.
 
-- 🔴 Sending a value with a different type than `Integer` or `null` for `cropLength` will return a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
+- 🔴 Sending a value with a different type than `Integer` or `null` for `cropLength` returns a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
```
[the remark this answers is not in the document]
gmourier: If there is no error, we must assume that nothing happens.

text/0118-search-api.md
gmourier: Yes, that's right!

text/0118-search-api.md
```diff
@@ -108,156 +109,344 @@ It returns the number of documents matching the current search query for each gi
...
+
+- 🔴 Sending a value with a different type than `Integer` or `null` for `cropLength` returns a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
+
+##### 3.1.12.1 Examples
+
+###### 3.1.12.1.1. Extending around
+
+Given an attribute defined in `attributesToCrop` containing:
+
+`"In his ravenous hatred he found no peace, and with boiling blood he scoured the umbral plains, seeking vengence afgainst the dark lords who had robbed him."`
```
bidoubiwa, line 269: 😨

text/0118-search-api.md
```diff
@@ -108,156 +109,344 @@ It returns the number of documents matching the current search query for each gi
...
+
+With `croplength` defined as `5` and `q` defined as `boiling blood`, the cropped value will be:
+
+`"…and with boiling blood he…"`
+
+Croppped query terms are counted as a word regarding `cropLength`.
```
bidoubiwa, line 275:
```suggestion
Cropped query terms are counted as a word regarding `cropLength`.
```

text/0118-search-api.md
bidoubiwa: From what I remember, attributesToRetrieves only impacts the fields at the root of the hit. AttributesToHighlight and attributesToCrop applies only in `_formatted`.

bidoubiwa (changes requested):
🔥 Good job

text/0118-search-api.md
```diff
@@ -108,156 +111,273 @@ It returns the number of documents matching the current search query for each gi
...
+
+#### 3.1.1.12. `cropLength`
 
 - Type: Integer
 - Required: False
-- Default: `200`
+- Default: `10`
 
-Configures the number of characters to keep on each side of the matching query term when using the `attributesToCrop` parameter.
+Sets the total number of **words** to keep around the matched part of an attribute specified in the `attributesToCrop` parameter.
```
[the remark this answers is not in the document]
gmourier: I have clarified the sentence (so that it is clearer and does not leave room for doubts) and you can find examples further in the specification. See 3.1.12.1.1. Extending around section.

text/0118-search-api.md
gmourier: I have rewritten this section "Matched Part To Be Cropped" with a detail of the density calculation. I hope it is clearer. 🙏

text/0118-search-api.md
```diff
@@ -108,156 +109,344 @@ It returns the number of documents matching the current search query for each gi
...
+
+- 🔴 Sending a value with a different type than `Array[String]`(POST), `String`(GET) or `null` for `attributesToHighlight` returns a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
+
+##### 3.1.8.1. searchableAttributes
+
+Attributes not defined in the `searchableAttributes` index setting are also highlighted if assigned to `attributesToHighlight`.
+
+##### 3.1.8.2. stopWords
+
+Attributes defined in the `stopWords` index setting are also highlighted if matched.
```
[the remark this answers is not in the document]
gmourier: Yes, I don't think it is a problem to highlight the terms written in the query terms whether they are stop-words or not.

The formatting of the search results is there to help the reading of the search results by highlighting the query terms contained in a search result. If a query term typed by the user can be highlighted, it should help the user read the results.

Stop-words are used to avoid degrading relevancy when they populate an attribute densely. I'm currently not convinced that removing them helps more than anything else regarding the reading of search results.

If you have arguments that help me to understand why it does not improve the reading of the search results if they are there, I am ready to reconsider the question!  

Good point about synonyms, I didn't know about this current behavior, I will specify it because it's nice.

text/0118-search-api.md
gmourier: The behavior is unchanged from today. I have added some important instructions about this (extracted from the deleted _formatted-field-behavior specification).

- If `attributesToHighlight` and `attributesToCrop` are not set, `_formatted` is not returned.
- If cumulated fields in `attributesToHighlight` and `attributesToCrop` resolve to only having non-existent fields, `_formatted` is not returned.
- If `attributesToRetrieve` is equal to `*` and `attributesToHighlight` or `attributesToCrop` are equals to `*`, `_formatted` is returned and contains `attributesToRetrieve` fields then compute highlights and crops on each received fields.
- If `attributesToRetrieve` is equal to `*` and `attributesToHighlight` or `attributesToCrop` contains a set of fields, `_formatted` is returned and contains `attributesToRetrieve` fields but only compute highlights and crops on fields declared in `attributesToHighlight` or `attributesToCrop`.

text/0118-search-api.md
```diff
@@ -108,156 +109,344 @@ It returns the number of documents matching the current search query for each gi
...
+
+###### 3.1.12.1.2. Length Reassignment to the right.
+
+Given an attribute defined in `attributesToCrop` containing:
+
+`"In his ravenous hatred he found no peace, and with boiling blood he scoured the umbral plains, seeking vengence afgainst the dark lords who had robbed him."`
+
+With `croplength` defined as `5` and `q` defined as `hatred`, the cropped value will be:
+
+`"In his ravenous hatred he…"`
```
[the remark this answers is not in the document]
bidoubiwa: > "In his ravenous hatred he found no peace, and with boiling blood he scoured the umbral plains, seeking vengence afgainst the dark lords who had robbed him."

if the word is `he` I feel like with your examples the words around it are more often in front of the word than behind.

For example: 
```
 With boiling blood he scoured...
```

Instead of for example

```
blood he scoured the umbrella...
```

its a fake example just. In the first one 3 words are in front of `he` and in the second 3 word are behind `he`. 
What is the rule for this?

text/0118-search-api.md
```diff
@@ -108,156 +109,355 @@ It returns the number of documents matching the current search query for each gi
...
+
+- 🔴 Sending a value with a different type than `Integer` or `null` for `cropLength` returns a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
+
+##### 3.1.12.1 Examples
+
+###### 3.1.12.1.1. Extending around
+
+Given an attribute defined in `attributesToCrop` containing:
+
+`"In his ravenous hatred he found no peace, and with boiling blood he scoured the umbral plains, seeking vengence afgainst the dark lords who had robbed him."`
```
mdubus, line 273: What happens if I search for `he found no peace` but I set `cropLength` to 2 ? 
And what if I set it to 0 ? Is it considered as null ?

open-api.yaml
```diff
@@ -1476,17 +1503,31 @@ paths:
...
                     default: '[]'
+                cropMarker:
+                  type: string
+                  description: Sets the crop marker to apply before and/or after cropped part selected within an attribute defined in `attributesToCrop` parameter.
+                  default: '…'
                 cropLength:
                   type: number
-                  description: Length used to crop field values.
-                  default: 200
+                  description: Sets the total number of **words** to keep around the matched part of an attribute specified in the `attributesToCrop` parameter.
```
mdubus, line 1529: By reading this, I understand that we will have XX words before and XX words after the match, not that the number of words returned will be XX.
Example : 
- Consider this phrase: `In his ravenous hatred he found no peace`
- Set `cropLength` to 2.
- I search for `hatred`

This description makes me think I will receive `his ravenous hatred he found` because I will have 2 words before and after `hatred` , which is in contradiction with what I've read below in this PR.

guimachiavelli (commented):
## Naming

Although part of me likes the `highlightPreTag`, I think it can be confusing for front-end devs. Whenever I read it, the `<pre>` HTML tag comes immediately to mind and it takes me a few seconds to adjust. It's also not mandatory people use an actual tag, right?

A few ideas:
- `highlightPrefix`, `highlightSuffix`
- `highlightBefore`, `highlightAfter`
- `beforeHighlight`, `afterHighlight`

I think `cropMarker` is fine. A few of the text truncation libraries I looked at used something like `ellipsis`, but that's terrible. `cropIndicator` could work, but not sure it's an improvement.

bidoubiwa:
To come back on your comment @guimachiavelli 

- `highlightPrefix`, `highlightSuffix`: I like this one a lot!  
- `highlightBefore`, `highlightAfter`: This one is not very clear to be
- `beforeHighlight`, `afterHighlight`: Same as above. It gives me a sense of `before the war` `after the war` vibe. I don't know how to express it. But it feels like it is in another context than the actual one. 

> I think `cropMarker` is fine. A few of the text truncation libraries I looked at used something like `ellipsis`, but that's terrible. `cropIndicator` could work, but not sure it's an improvement.

I also like `cropMarker`, I might have a small preference for `cropIndicator` though. And yes ellipsis is just terrible.

text/0118-search-api.md
```diff
@@ -108,156 +109,355 @@ It returns the number of documents matching the current search query for each gi
...
+Synonyms are also highlighted.
+
+#### 3.1.9. `highlightPreTag`
+
+- Type: String
+- Required: False
+- Default: `"<em>"`
 
-##### 1.2.1.9 `attributesToCrop`
+Specifies the tag, or sequence, to put **before** every highlighted query terms.
```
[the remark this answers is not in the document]
gmourier: Yes @guimachiavelli, this is what "sequence" allowed me to express, the simplest would be perhaps to write string. There is no particular check to verify whether it is an HTML tag or not.

text/0118-search-api.md
gmourier: Yes!

text/0118-search-api.md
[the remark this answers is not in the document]
gmourier: @ManyTheFish You may have a more precise way of describing the behavior than me. Can you suggest what you implemented regarding the cropping algorithm? It will really help me here 🙏 🤯

text/0118-search-api.md
```diff
@@ -108,156 +109,355 @@ It returns the number of documents matching the current search query for each gi
...
+
+In this example:
+
+- `"hatred"` count for `1` word.
+- The left part contains `3` words. e.g. `"In his ravenous`.
+- The cropped part will be filled by `1` words on the right to reach `cropLength`. e.g. `"he"`.
+
+#### 3.1.13. `cropMarker`
+
+- Type: String
```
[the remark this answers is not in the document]
gmourier: They can use a string of any length, I don't see why we should deprive them of it. I will precise that!

text/0118-search-api.md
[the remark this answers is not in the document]
gmourier: Your thoughts are interesting, I think these are topics that can be addressed later if they need to be. Here is my thought to give more context:

- It describes the current behavior of the engine (for stop-words and synonyms). If this was a problem for a user at one time, it is not important enough to be important now. Changing this behavior could impact other users, it would involve adding an option to allow choosing but would add complexity and time to hit the v0.27.0 while we want to mostly reduce pain points regarding the customizable highlight tag / poor cropper that forces users to use `_matchesInfo` to do this work on the frontend-side.
- I also checked this behavior which is present by default in Algolia. With so little feedback and this behavior already in place, I think we can base our choices on what a well-placed and well-rated actor does.

I can give my feedback that goes in the opposite direction of your idea about stop words to give the change but again it's just to confront it.

If a query contains several stop-words, e.g. `The son was on the path of redemption`.  In this case, doesn't removing the stop-words complicate the readability if the exact phrase can be highlighted?

text/0118-search-api.md
```diff
@@ -108,156 +109,355 @@ It returns the number of documents matching the current search query for each gi
...
 
-- 🔴 Sending a value with a different type than `Array[String]`(POST), `String`(GET) or `null` for `attributesToHighlight` will return a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
+##### 3.1.8.2. stopWords
 
-> See [_Formatted Field Behavior](0039-_formatted-field-behavior_.md)
+Attributes defined in the `stopWords` index setting are also highlighted if matched.
+
+##### 3.1.8.3. Tokenizer Separators
+
+Tokenizer separators are not highlighted.
```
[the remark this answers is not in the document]
gmourier: @guimachiavelli Good point! If I remember, we discussed it with Many during our discovery phase but we made the choice not to change this behavior in order to keep a consistent behavior on how to apply the highlight tags. I note it as a future possibility.

text/0118-search-api.md
```diff
@@ -108,156 +109,355 @@ It returns the number of documents matching the current search query for each gi
...
+
+Sets the total number of **words** to keep for the cropped part of an attribute specified in the `attributesToCrop` parameter. It means that if `10` is set for `cropLength`, the cropped part returned in `_formatted` will only be 10 words long.
+
+This parameter is applied to the fields from `attributesToCrop`. If there are none, this parameter has no effect. See [3.1.11. `attributesToCrop`](#3111-attributestocrop) section.
+
+- 🔴 Sending a value with a different type than `Integer` or `null` for `cropLength` returns a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
+
+##### 3.1.12.1 Examples
+
+###### 3.1.12.1.1. Extending around
```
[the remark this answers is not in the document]
gmourier: @ManyTheFish Regarding https://github.com/meilisearch/specifications/pull/120#discussion_r837419461

text/0118-search-api.md
```diff
@@ -108,156 +109,355 @@ It returns the number of documents matching the current search query for each gi
...
+
+If no part is found when selecting a part to be cropped. The returned value in `_formatted` will start at the beginning of the attribute until the distance of `cropLength` is reached.
+
+###### 3.1.13.1.2. Positioning Markers
+
+If the cropped part has been matched against query terms and contains the beginning of the attribute to be cropped, the `cropMarker` is not placed to the left of the cropped part.
+
+If the cropped part has been matched against query terms and contains the end of the attribute to be cropped, the `cropMarker` is not placed to the right of the cropped part.
+
+#### 3.1.14. `matches`
```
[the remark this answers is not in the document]
gmourier: Yes, `_matchesInfo` contains the positions or query terms matched in the document, regardless of whether they are highlighted or cropped.

Stop-words, if encountered are included in `_matchesInfo` as are synonyms.

text/0118-search-api.md
```diff
@@ -108,156 +109,355 @@ It returns the number of documents matching the current search query for each gi
...
 
 - Type: Boolean
 - Required: False
 - Default: `false`
 
-Adds a `_matchesInfo` object to the search response that contains the location of each occurrence of queried terms across all fields. This is useful when more control is needed than offered by the built-in highlighting/cropping features.
+Adds a `_matchesInfo` object to the search response that contains the location of each occurrence of queried terms across all fields. The given positions are in bytes.
```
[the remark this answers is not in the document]
gmourier: It is an arbitrary choice that does not fit everyone. I don't have the exact explanation in my head, but it seems to me that it comes from the way the tokenizer gets some characters from the Unicode table that are several bytes long. Making this character is not perfect, as well as taking the number of bytes. 

There are two sides to this, but during the discovery phase, we realize that people use `_matchesInfo` to put their own highlight tag, cropping marker, or rework the cropping which is not very useful nowadays.

The best solution might be to offer both `bytes`/`chars` modes in the future if people want to do what they want. What's interesting is that our competitors don't offer this feature and it doesn't seem to bother their users more than that.

gmourier:
Thanks for your feedback! 

Thanks for the feedback on the naming. Personally, I don't have a particular preference. I chose the terms used by Algolia to facilitate the migration of a user coming from this product to Meilisearch and thus accelerate its handling. Do you see anything really critical not to use these? Particularly asking this to you @guimachiavelli since you mentionned the `pre` thing. 

Concerning `cropMarker` if it's nice for everyone, I suggest that we keep it since it's shorter than `cropIndicator`.

text/0118-search-api.md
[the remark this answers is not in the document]
gmourier: @ManyTheFish Regarding https://github.com/meilisearch/specifications/pull/120#discussion_r837419461

text/0118-search-api.md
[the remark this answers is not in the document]
ManyTheFish: Hey @guimachiavelli and @gmourier!
From my side, the behavior is to crop around words in the same context that matching words. This means that we will keep words that are farther from the matching words but are in the same phrase, than words that are nearer but separated by a dot.
For instance, for the matching word `Split` the text:
`Natalie risk her future. Split The World is a book written by Emily Henry. I never read it.`
will be cropped like:
`…Split The World is a book written by Emily Henry. …`
and  not like:
`Natalie risk her future. Split The World is a book …`

text/0118-search-api.md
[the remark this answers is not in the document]
ManyTheFish: Hey @guimachiavelli and @gmourier!
About how do we crop around match [I wrote a response here](https://github.com/meilisearch/specifications/pull/120#discussion_r837495765).

About how do we find the "best" matches, we try to find the crop interval:
1) that have the highest count of unique matches
> for example, if we have a query `split the world`, then the interval `the split the split the` has 4 matches but only 2 unique matches (1 for `split` and 1 for `the`) where the interval `split of the world` has 3 matches and 3 unique matches. So the interval `split of the world` is considered better.
2) that have the minimum distance between matches
> for example, if we have a query `split the world`, then the interval `split of the world` has a distance of 3 (2 between `split` and `the`, and 1 between `the` and `world`) where the interval `split the world` has a distance of 2. So the interval `split the world` is considered better.
3) that have the highest count of ordered matches
> for example, if we have a query `split the world`, then the interval `the world split` has 2 ordered words where the interval `split the world` has 3. So the interval `split the world` is considered better.

brunoocasali (commented):
Just left some comments because the feeling is stronger than me! 😄 

Awesome work team!

open-api.yaml
```diff
@@ -681,6 +681,22 @@ components:
...
             - ''
         example: 'title,description'
+    highlightPreTag:
+      name: highlightPreTag
+      in: query
+      required: false
+      description: Specify the tag to put before the highlighted query terms.
+      schema:
+        type: string
+        default: '<em>'
```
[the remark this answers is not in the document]
gmourier: @brunoocasali We did not change this value because it's already like that and the users seem to be used to it. Did'nt saw users reporting direct feedback about the default value.

Changing it could change all the frontend code that is already using `em` and could have applied CSS rules to it. As the settings now allow to customize this, it seemed to me to have little impact.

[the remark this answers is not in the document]
brunoocasali: Yeah, I can see your point thanks for it! While researching it, I saw Elasticsearch and Solr use the same element (`<em>`) as the default one.

Even if I disagree, I'm sold regarding preventing an expectation "breaking" for the current users, so let's keep it!

- https://solr.apache.org/guide/8_11/highlighting.html
- https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html

text/0118-search-api.md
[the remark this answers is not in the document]
guimachiavelli: Ok, that makes sense. Thanks!

guimachiavelli:
Regarding `pre`/`post` highlight: trying to think as a developer unfamiliar with Algolia, I think I'd find the `pre` thing a bit annoying, but not very important in the larger scheme of things. The benefits for someone using Algolia are clear and significant.

Based on @bidoubiwa's feedback, the only two reasonable options are `highlightPrefix` and `highlightPreTag`. I think @gmourier is better equipped to make the final call, as in the end it all seems to come to "how important is maintaining name parity with Algolia for our product?"

Regarding the crop marker: I also prefer `cropMarker` because it's a bit shorter but otherwise equivalent.

text/0118-search-api.md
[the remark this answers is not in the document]
guimachiavelli: Great, thanks for clarifying, @ManyTheFish!

open-api.yaml
```diff
@@ -1476,17 +1503,31 @@ paths:
                     type: string
                     example: '["title", "overview"]'
                     default: '[]'
+                highlightPreTag:
```
brunoocasali, line 1506: I'm questioning myself about the usage of `highlightPreTag` or `highlightPostTag` alone. 
Imagine a user going to highlight their search.

The dev called the `search` method using a different highlighter tag, but it did not close or vice-versa, it will have a not-expected result.

```rb
client.index('movies').search("batman", highlightPreTag: "<mark>")
#=> [{..., text: "<mark>batman</em>" }]

client.index('movies').search("batman", highlightPostTag: "</mark>")
#=> [{..., text: "<em>batman</mark>" }]

client.index('movies').search("batman", highlightPostTag: "$$$$")
#=> [{..., text: "<em>batman$$$$" }]
```

My question is, it is OK to allow users to make this kind of request (using just one or another)?
I'm thinking about the possibility to raise an exception if the user wants to use one or another alone.

But I would like to grab some opinions on this subject.
FYI: @gmourier

gmourier: I don't see a direct problem with letting them do this for now as I imagine users will specify both fields most of the time.

I see a case for not forcing both to be set for the future. If we add an index setting that allows custom default values to be set, a frontend client might only want to change one or the other at search time so as not to override one of the two already set defaults.

Also, today if a user wants to add a custom class to `<em>` or something afterward without modifying the `highlightPostTag` it will not be possible if we force both to be defined.

I don't think it's a problem to leave it the way it is for now but if you are against it and think it's better for the UX, I'm with you. 👍  

@ManyTheFish You may have an opinion on that since you are on it

gmourier:
📡 @meilisearch/docs-team, @ManyTheFish f0265bfe9c2350d01aa6973c8a90decee2369ada Adds the new telemetry concerning the use of parameters allowing to customize the behavior of the crop and the highlight.

bidoubiwa:
Hey @gmourier, which is the final decision for the naming ?

- `highlightPreTag` and `highlightPostTag`
- `highlightPrefix` and `highlightPreTag`

gmourier:
@bidoubiwa Yes thanks for reminding it, we keep `highlightPreTag` and `highlightPostTag`!

text/0118-search-api.md
```diff
@@ -282,37 +282,21 @@ Cropped query terms are counted as a word regarding `cropLength`.
...
 
-###### 3.1.12.1.3. Length Reassignment to the left.
+###### 3.1.12.1.2. Keeping a phrase context
 
-Given an attribute defined in `attributesToCrop` containing:
-
-`"In his ravenous hatred he found no peace, and with boiling blood he scoured the umbral plains, seeking vengence afgainst the dark lords who had robbed him."`
-
-With `croplength` defined as `5` and `q` defined as `had robbed`, the cropped value will be:
+Meilisearch keep words that are farther from the matching words but are in the same phrase, than words that are nearer but separated by a dot.
```
bidoubiwa, line 287:
```suggestion
When determining where to crop, Meilisearch searches for matches that are maybe further from each other but that are present in the same sentence instead of matches that are close but separated by a dot (end of sentence).
```

text/0118-search-api.md
```diff
@@ -336,12 +320,19 @@ This parameter is applied to the fields configured in `attributesToCrop`. If the
 
 The cropping algorithm tries to match the window with the highest density of query terms within the `cropLength` limit.
 
-Here is how the density is calculated:
+The cropping algorithm find the "best" matches, it tries to find the crop interval:
```
bidoubiwa, line 323: not sure I understand this sentence. Could you rephrase it?

text/0118-search-api.md
```diff
@@ -336,12 +320,19 @@ This parameter is applied to the fields configured in `attributesToCrop`. If the
...
+
+1. That have the highest count of unique matches
```
bidoubiwa, line 325:
```suggestion
1. That has the highest count of unique matches
```

bidoubiwa (changes requested):

ManyTheFish:
the commit that precise the [behavior for _formatted](https://github.com/meilisearch/specifications/pull/120/commits/d10f231006a689f13d4b50d81b822f40dcf893cb), in fact, changes the behavior of formated relying on `displayedAtrribute` instead of `AttributeToRetrieve`, when Today's implementation rely on `AttributeToRetrieve`. 🤔

text/0118-search-api.md
dichotommy: Or is it just what the example below would suggest, that if it makes no difference to criteria such as number of unique matches, Meilisearch will try to start or end the crop at the beginning or end of the sentence?

I think the example could be improved if you changed the sentence to something like

`"Eventually, Natalie and her split up. Split The World is a book written by Emily Henry. The world never fully recognized her talent."`

I think that would provide a clearer view of the edge cases

text/0118-search-api.md
```diff
@@ -108,156 +109,350 @@ It returns the number of documents matching the current search query for each gi
...
+> See [GeoSearch](0059-geo-search.md)
+
+###### 3.2.1.1.2. `_formatted`
+
+- Type: Object
+- Required: False
+
+`_formatted` returns highlighted and cropped attributes specified in `attributesToHighlight` and/or `attributesToCrop` of a search result.
+
+- If `attributesToHighlight` and `attributesToCrop` are not set, `_formatted` is not returned.
```
dichotommy, line 410:
```suggestion
- `_formatted` is only returned if one or both of `attributesToHighlight` and `attributesToCrop` are set.
```

text/0118-search-api.md
```diff
@@ -108,156 +109,350 @@ It returns the number of documents matching the current search query for each gi
...
+- If cumulated fields in `attributesToHighlight` and `attributesToCrop` resolve to only having non-existent fields, `_formatted` is not returned.
```
dichotommy, line 411: If I understand this correctly, it could be rewritten as:

```suggestion
- If none of the fields contained in `attributesToHighlight` and `attributesToCrop` are present in any of the matched documents, `_formatted` is not returned.
```

dichotommy (commented):
Hey, I heard there was some confusion about `_formatted` in this spec so I wanted to take a crack at it and help clarify if I can. However I didn't make too many writing suggestions in this review, just concentrated on reading the text and asking some questions. May come back to make a second pass if I can find some time tomorrow.

gmourier:
> the commit that precise the [behavior for _formatted](https://github.com/meilisearch/specifications/pull/120/commits/d10f231006a689f13d4b50d81b822f40dcf893cb), in fact, changes the behavior of formated relying on `displayedAtrribute` instead of `AttributeToRetrieve`, when Today's implementation rely on `AttributeToRetrieve`. 🤔

There is no change in behavior compared to v0.26. This was to try to explain more clearly the behavior of _formatted.

It's being rewritten, thanks to @dichotommy and @curquiza.

text/0118-search-api.md
bidoubiwa: In this case, do you have an idea on how to rephrase it @dichotommy

text/0118-search-api.md
```diff
@@ -108,156 +109,350 @@ It returns the number of documents matching the current search query for each gi
...
 - Type: Integer
 - Required: False
-- Default: `200`
+- Default: `10`
+
+Sets the total number of **words** to keep for the cropped part of an attribute specified in the `attributesToCrop` parameter. It means that if `10` is set for `cropLength`, the cropped part returned in `_formatted` will only be 10 words long.
+
+This parameter is applied to the fields from `attributesToCrop`. If there are none, this parameter has no effect. See [3.1.11. `attributesToCrop`](#3111-attributestocrop) section.
+
+Sending a `0` value is equivalent to `null` and thus deactivates the cropping. It works the same for a custom crop length per attribute in `attributesToCrop`, except it deactivates it for the related field.
```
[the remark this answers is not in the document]
gmourier: You are right @dichotommy, it is not clear. This statement "deactivates the cropping" only works if there is no overriding of the crop length directly in `attributesToCrop`, I will try to rephrase that.

text/0118-search-api.md
```diff
@@ -108,156 +109,350 @@ It returns the number of documents matching the current search query for each gi
...
+Specifying `cropMarker` to `""` or `null` implies that no marker will be applied to the cropped part.
 
-Configures the number of characters to keep on each side of the matching query term when using the `attributesToCrop` parameter.
+This parameter is applied to the fields configured in `attributesToCrop`. If there are none, this parameter has no effect. See [3.1.11. `attributesToCrop`](#3111-attributestocrop) section.
 
-If `attributesToCrop` is not configured, `cropLength` has no effect on the returned results.
+- 🔴 Sending a value with a different type than `String` or `null` for `cropMarker` returns a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
 
-- 🔴 Sending a value with a different type than `Integer` or `null` for `cropLength` will return a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
+##### 3.1.13.1. Applying `cropMarker`
```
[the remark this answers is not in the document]
gmourier: What do you mean by`?` @dichotommy?

text/0118-search-api.md
```diff
@@ -108,156 +109,350 @@ It returns the number of documents matching the current search query for each gi
...
+
+###### 3.1.12.1.1. Extending around
+
+Given an attribute defined in `attributesToCrop` containing:
+
+`"In his ravenous hatred he found no peace, and with boiling blood he scoured the umbral plains, seeking vengence afgainst the dark lords who had robbed him."`
+
+With `croplength` defined as `5` and `q` defined as `boiling blood`, the cropped value will be:
+
+`"…and with boiling blood he…"`
```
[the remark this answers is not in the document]
gmourier: In this case yes but not necessarily, it will depend on other factors of the algorithm that seeks to select the best match.

text/0118-search-api.md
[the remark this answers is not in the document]
dichotommy: There's a heading here but no content... or am I going crazy? 😱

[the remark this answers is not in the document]
dichotommy: Ah, is it just because you've passed the number of nesting allowed by HTML headers? I understand now, please ignore me 🐒

text/0118-search-api.md
bidoubiwa: I prefer the first one!

text/0118-search-api.md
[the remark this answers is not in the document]
ManyTheFish: the contract for a croplenght of 10 is: "always 10 words returned or all the attribute".

text/0118-search-api.md
```diff
@@ -108,156 +109,350 @@ It returns the number of documents matching the current search query for each gi
...
+A custom crop length set in this way has priority over the `cropLength` parameter.
 
-##### 1.2.1.10 `cropLength`
+##### 3.1.11.2. searchableAttributes
+
+Attributes configured in `attributesToCrop` are cropped even if not present in the `searchableAttributes` index setting.
+
+##### 3.1.11.3. stopWords
+
+Attributes defined in the `stopWords` index setting are counted as words regarding `cropLength`.
```
guimachiavelli, line 249: Unless I'm very confused (always a possibility), stop-words don't accept attributes as values.

```suggestion
Terms defined in the `stopWords` index setting are counted as words regarding `cropLength`.
```

text/0118-search-api.md
```diff
@@ -108,156 +109,350 @@ It returns the number of documents matching the current search query for each gi
...
+
+For example, for the query terms `split the world`, then the interval `split of the world` has a distance of `3` (`2` between `split` and `the`, and `1` between `the` and `world`) where the interval `split the world` has a distance of `2`. So the interval `split the world` is considered better.
+
+3. That have the highest count of ordered matches
+
+For example, for the query terms `split the world`, then the interval `the world split` has `2` ordered words where the interval `split the world` has `3`. So the interval `split the world` is considered better.
+
+Only one cropped part from an attribute is returned.
+
+If no part is found when selecting a part to be cropped, the returned value in `_formatted` will start at the beginning of the attribute include a number of words equal to `cropLength`.
```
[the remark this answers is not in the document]
gmourier: Indeed it is the case, if there is only one word the algorithm will take the one the most left of the attribute. Here, the algorithm describes the case where there are several words that can be matched and in several places (the most _complex_ case).

text/0118-search-api.md
```diff
@@ -468,6 +523,8 @@ Whether `nbHits` is exhaustive.
 
 Added to the search response when `facetsDistribution` is set for a search query. It contains the number of remaining candidates for each specified facet in the `facetsDistribution` search parameter.
 
+If a field distributed as a facet contains no value, it is returned as a `facetDistribution` field with an empty object as value.
```
bidoubiwa, line 526: What do you mean with this?

gmourier: Trying to explain the expected behavior here https://github.com/meilisearch/meilisearch/issues/2352#issue-1216056948  !

gmourier: I will rephrase it soon to be more precise with our last tests

bidoubiwa: Ah I see!

gmourier:
Need last approvals to merge this specification and publish the Open API specification regarding v0.27.

PR #127, opened 2022-03-11
Update spec template according to our current usage

curquiza:
I update the template of the specs according to our current usage
- I remove the Meilisearch tracking issue part since sometimes multiples issues are related (with the improvements of the spec for example). However, if this is something you want to keep back in the process, please let me know
- I simplified the titles so that it's more adapted to what we do at the moment.

Since we are still learning about managing specs, I think this is important to provide a user-friendly format first adapted to our current usage, and make it improve as we go along. Complexity is not our friend at our stage from my POV 🤓 

Let me know if you want me to add something else 😇

### DIFF
--- a/text/0000-specification-template.md
+++ b/text/0000-specification-template.md
@@ -6,6 +6,6 @@
 
 ## 3. Functional Specification
 
-## 4. Technical Aspects
+## 4. Technical Details
 
 ## 5. Future Possibilities
--- a/text/0000-specification-template.md
+++ b/text/0000-specification-template.md
@@ -1,5 +1,3 @@
-- Title:
-
 # Title
 
 ## 1. Summary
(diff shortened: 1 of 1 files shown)


text/0000-specification-template.md
```diff
@@ -1,21 +1,15 @@
 - Title:
 - Start Date:
 - Specification PR:
```
gmourier, line 3:
```suggestion
```

gmourier (changes requested):
Thanks, @curquiza, I already started a branch on my side a few days ago to rewrite the template and the readme!

We can keep this one for the template. I think we can remove the start date and PR Meilisearch, it seems useless right now.

gmourier (dismissed):

gmourier (dismissed):

gmourier (approved):

PR #136, opened 2022-03-30
Add missing "v" in the user-agent data

brunoocasali:
Following the issue https://github.com/meilisearch/integration-guides/issues/150 there is a "v" before the version of the package.

I'm not sure now if this change is valid or the [integrations team issue #150](https://github.com/meilisearch/integration-guides/issues/150) is wrong 😅


gmourier:
Thanks @brunoocasali !!! Anyway, it's better to change the spec than all the SDKs. 😄

gmourier (approved):

PR #138, opened 2022-03-31
API Guideline - Return list of API resources under a `results` array

gmourier:
🤖 [API Diff](https://github.com/meilisearch/specifications/pull/138#issuecomment-1090115742)

## Why?

In order to stabilize the APIs around a convention, we add resources under a `results` array when they can be listed. 

This is already the case for the `tasks` and `API Keys` resources which have made this change along the way.

The `/search` endpoints are not affected by these changes but this can be reconsidered with caution. `hits` seem to be effective and changing it would cause a big breaking change. `hits` is already well established in the market e.g. elasticsearch, algolia, typesense.

The stats API is not concerned for the moment and maybe concerned later given feedback and what it could become. The idea behind it is that this resource does not only return index statistics but also more global statistics which makes the use of `results` less straightforward.

## TODO

- [x] Indexes API
- [x] Documents API
- [x] limit/offset pagination for Indexes API
- [x] limit/offset pagination for Keys API
- [x] Update OpenAPI

## Changes

- Returns `index` API resources on `GET - /indexes` under a `results` array + offset/limit pagination.
    - Add `limit` (Default: `20`) and `offset`(Default: `0`) query parameters.
    - Add `limit`, `offset` and `total` in the response body.
- Returns `document` API resources on `/documents` under a `results` array.
    - Add `limit`, `offset` and `total` in the response body.
- Add an offset/limit pagination on `GET - /keys`
    - Add `limit` (Default: `20`) and `offset`(Default: `0`) query parameters.
    - Add `limit`, `offset` and `total` in the response body.

### DIFF
--- a/text/0085-api-keys.md
+++ b/text/0085-api-keys.md
@@ -367,8 +367,67 @@ Only the master key allows managing the API keys.
 ```
 "Authorization: Bearer :masterKey"
 ```
+##### Query Parameters
 
-##### Response
+| Field                    | Type                     | Required |
+|--------------------------|--------------------------|----------|
+| `offset`                 | Integer / `null`         | false    |
+| `limit`                  | Integer / `null`         | false    |
+
+###### `offset`
+
+- Type: Integer
+- Required: False
+- Default: `0`
+
+Sets the starting point in the results, effectively skipping over a given number of API keys.
+
+###### `limit`
+
+- Type: Integer
+- Required: False
+- Default: `20`
+
+Sets the maximum number of documents to be returned by the current request.
+
+##### Response Definition
+
+| Field                    | Type                     | Required |
+|--------------------------|--------------------------|----------|
+| `result`                 | Array[Key]               | true     |
+| `offset`                 | Integer                  | true     |
+| `limit`                  | Integer                  | true     |
+| `total`                  | Integer                  | true     |
+
+###### `results`
+
+- Type: Array[Key]
+- Required: True
+
+An array containing the fetched API keys.
+
+###### `offset`
+
+- Type: Integer
+- Required: True
+
+Gives the `offset` parameter used for the query.
+
+###### `limit`
+
+- Type: Integer
+- Required: True
+
+Gives the `limit` parameter used for the query.
+
+###### `total`
+
+- Type: Integer
+- Required: True
+
+Gives the total number of API keys that can be browsed.
+
+###### Example
 
 `200 Success`
 
@@ -416,7 +475,10 @@ Only the master key allows managing the API keys.
             "createdAt": "2021-08-11T10:00:00Z",
             "updatedAt": "2021-08-11T10:00:00Z"
         }
-    ]
+    ],
+    "offset": 0,
+    "limit": 20,
+    "total": 3
 }
 ```
 
--- a/text/0124-documents-api.md
+++ b/text/0124-documents-api.md
@@ -53,7 +53,7 @@ Unique identifier of an index.
 | Field                    | Type                     | Required |
 |--------------------------|--------------------------|----------|
 | `offset`                 | Integer / `null`         | false    |
-| `limit`                  | String / `null`          | false    |
+| `limit`                  | Integer / `null`         | false    |
 | `attributesToRetrieve`   | String / `null`          | false    |
 
 ###### 3.1.1.2.1. `offset`
@@ -95,6 +95,7 @@ A `results` array representing documents as JSON objects.
 | `results`                | Array[Document]          | true     |
 | `offset`                 | Integer                  | true     |
 | `limit`                  | Integer                  | true     |
+| `total`                  | Integer                  | true     |
 
 ###### 3.1.1.3.1. `results`
 
@@ -121,6 +122,12 @@ Gives the `limit` parameter used for the query.
 
 > See [3.1.1.2.2. `limit`](#31122-limit) section.
 
+###### 3.1.1.3.3. `total`
+
+- Type: Integer
+- Required: True
+
+Gives the total number of documents that can be browsed in the related index.
 
 ###### 3.1.1.3.4. Example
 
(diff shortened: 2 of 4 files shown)


bidoubiwa (changes requested):

brunoocasali (approved):
Awesome spec @gmourier :)

> The /search endpoints are not affected by these changes but this can be reconsidered with caution. hits seem to be effective and changing it would cause a big breaking change.

I really like standards, but in this case, it does not seem so appropriate to change from `hits` to `results`, because this word is already well established in the market, elasticsearch, algolia, typesense.

gmourier:
Thanks, @brunoocasali. Absolutely, this is also a good reason!

text/0085-api-keys.md
```diff
@@ -368,8 +368,67 @@ Only the master key allows managing the API keys.
 ```
 "Authorization: Bearer :masterKey"
 ```
+##### Query Parameters
 
-##### Response
+| Field                    | Type                     | Required |
+|--------------------------|--------------------------|----------|
+| `offset`                 | Integer / `null`         | false    |
+| `limit`                  | Integer / `null`         | false    |
```
guimachiavelli, line 376: So, if I understand it correctly, we're basically enabling pagination of API key GET requests, right?

Obviously not opposed in any way, but, out of curiosity, do we expect people to use so many keys they'll need pagination?

text/0085-api-keys.md
```diff
@@ -368,8 +368,67 @@ Only the master key allows managing the API keys.
...
+- Required: True
+
+Gives the `limit` parameter used for the query.
+
+###### `total`
+
+- Type: Integer
+- Required: True
+
+Gives the total number of API keys that can be browsed.
```
guimachiavelli, line 429: Is this an actual total or an estimated total? I'm guessing actual, but just to be sure.

text/0124-documents-api.md
```diff
@@ -121,6 +122,12 @@ Gives the `limit` parameter used for the query.
 
 > See [3.1.1.2.2. `limit`](#31122-limit) section.
 
+###### 3.1.1.3.3. `total`
+
+- Type: Integer
+- Required: True
+
+Gives the total number of documents that can be browsed in the related index.
```
guimachiavelli, line 130: Same question as in the other total: actual or estimated?

text/0124-documents-api.md
```diff
@@ -53,7 +53,7 @@ Unique identifier of an index.
 | Field                    | Type                     | Required |
 |--------------------------|--------------------------|----------|
 | `offset`                 | Integer / `null`         | false    |
-| `limit`                  | String / `null`          | false    |
+| `limit`                  | Integer / `null`         | false    |
```
guimachiavelli, line 56: huh, that's interesting. Why was the type string?

text/0132-indexes-api.md
```diff
@@ -98,9 +96,73 @@ Manipulate indexes of a Meilisearch instance.
...
+- Required: True
+
+An array containing the fetched indexes.
+
+###### 3.2.1.2.2. `offset`
+
+- Type: Integer
+- Required: True
+
+Gives the `offset` parameter used for the query.
```
bidoubiwa, line 147:
```suggestion
The `offset` contains the number of offset documents.
```

text/0132-indexes-api.md
```diff
@@ -98,9 +96,73 @@ Manipulate indexes of a Meilisearch instance.
...
+
+> See [3.2.1.1.1. `offset`](#32111-offset) section.
+
+###### 3.2.1.2.3. `limit`
+
+- Type: Integer
+- Required: True
+
+Gives the `limit` parameter used for the query.
```
bidoubiwa, line 156:
```suggestion
The `limit` parameter contains the max number of indexes returned in `results`.
```

text/0132-indexes-api.md
```diff
@@ -98,9 +96,73 @@ Manipulate indexes of a Meilisearch instance.
...
+
+> See [3.2.1.1.2. `limit`](#32112-limit) section.
+
+###### 3.2.1.2.3. `total`
+
+- Type: Integer
+- Required: True
 
-Returns an array made of multiple index API resource objects. See [3.1.2.1. Response Definition](#3121-response-definition) section.
+Gives the total number of indexes that can be browsed.
```
bidoubiwa, line 165:
```suggestion
The `total` contains the total number of indexes that can be browsed.
```

bidoubiwa (commented):
I made some suggestions in wording in the indexes response, if you agree, can you apply them to all other occurrences ?

text/0085-api-keys.md
gmourier: Actual total. There is no notion of approximation here. 🤓

text/0124-documents-api.md
gmourier: https://github.com/meilisearch/specifications/pull/138#discussion_r852920472

text/0085-api-keys.md
gmourier: Hi @guimachiavelli 👋

Exactly, it is to avoid a future breaking if we change later the number of results by adding the pagination, it also to be consistent by having pagination on all the lists of resources of the API from v0.28

It's possible that this could happen, especially for a company that would decide to have several indexes with specific keys per operation for each index for some security reasons / internal organization. 

We haven't had any expression of this need but the cloud team might also want to paginate to improve the UX of the dashboard at some point.

A User who creates no more than 20 keys shouldn't be directly bothered. The API keys are still in the `results` array

text/0124-documents-api.md
gmourier: It's a quick copy/paste jutsu that went through the reviews like a super ninja. 🥲

curquiza (approved):
👌

text/0085-api-keys.md
gmourier: Yes!

bidoubiwa (approved):
LGTM 🔥

PR #139, opened 2022-04-04
Dumps API - Make dump creation an asynchronous task

gmourier:
🤖 [API Diff](https://github.com/meilisearch/specifications/pull/139#issuecomment-1096491430)

## Motivation

Today, the route to retrieve the status of a dump is not consistent with other APIs that retrieve the details of a resource and it does not share the same values for the `status` enum comparing to a `task` `status`.

Rather than reworking the dump API endpoints, we propose to make dumps an asynchronous task like the others.

This allows to have a history of dumps created for free and to unify the status directly as well as to provide a way to see the status of a dump creation from its associated `task`.

## Changes

- Dump creation becomes an asynchronous task. The `type` associated is `dumpCreation`.
- The generated uid of a dump can be found in the `details` object of a task under the `dumpUid` field.
- Removes the 409 conflict `dump_already_processing` error when a dump is already processed since they can now be enqueued.
- `POST` on `/dumps` return a `202 Accepted` representing a summarized `task`. 
- The `202 Accepted` (Summarized task response) and the related `task` object define the `indexUid` to `null` since dump are related to all indexes.
- `dumpCreation` tasks are only visible on GET `/tasks` and `/task/task_uid` since they are not related to a specific index.
- A dump creation is now a visible asynchronous task. The dump creation task takes priority over other enqueued tasks exactly as today. Its position in the list of tasks is not changed.
- `/dumps/:dump_uid/status` is removed.
- Remove `dumps.get` API Key action.

## Misc

- Add forgotten task `type` for `202 Accepted` response in the open-api.yml

## 🚨

- [x] https://github.com/meilisearch/specifications/pull/144#issuecomment-1105107600


open-api.yaml
```diff
@@ -892,59 +860,16 @@ paths:
...
-                $ref: '#/components/schemas/dumps'
+                $ref: '#/components/responses/202'
               examples:
-                Created:
+                '202':
                   value:
-                    uid: 20210719-114144097
-                    status: in_progress
-                    startedAt: '2021-07-19T14:31:16.920473Z'
+                    uid: 0
```
maryamsulemani97, line 867: This should have a `type`, right?

gmourier: Good point Maryam. I'll change it to the appropriate place! Thank you very much.

text/0105-dumps-api.md
```diff
@@ -131,18 +99,20 @@ The following table describes which version of the dump correspond to which vers
...
 
-### 2.2. Importing a dump
+### 4.2. Importing a dump
 
 When a dump is being imported, the http API is not available. Meilisearch can't receive read or write requests.
 
-## 3. Future Possibilities
-- Make dump creation a `task`
+## 5. Future Possibilities
+
```
irevoire, line 117: As a future possibility, I was thinking; that maybe, at some point, we could create a “fake” task called `DumpImport` that would only be created when you import a dump.
That would allow people to see things like; which updates have been processed since they imported their dump or what was already indexed.

irevoire (approved):
Everything looks clear and good to me :+1: 

If he has the time I would love to see a review from @nicolasvienot though because that's quite a big breaking for the cloud :grimacing:

text/0105-dumps-api.md
gmourier: I've added it as a future possibility! Thanks @irevoire

text/0060-tasks-api.md
```diff
@@ -58,10 +58,10 @@ The motivation is to stabilize the current `update` resource to a version that c
 | field   | type    | description                     |
 |---------|---------|---------------------------------|
 | uid      | integer | Unique sequential identifier           |
-| indexUid | string | Unique index identifier |
+| indexUid | string | Unique index identifier. This field is defined to `null` for a `dumpCreation`. |
```
bidoubiwa, line 61:
```suggestion
| indexUid | string | Unique index identifier. This field is `null` when the task type is `dumpCreation`. |
```

Should we say `string or null` in the type column?

text/0060-tasks-api.md
```diff
@@ -76,9 +76,10 @@ The motivation is to stabilize the current `update` resource to a version that c
 | field      | type    | description                     |
 |------------|---------|---------------------------------|
 | uid        | integer | Unique sequential identifier           |
-| indexUid   | string | Unique index identifier |
+| indexUid   | string  | Unique index identifier. This field is defined to `null` for a `dumpCreation`. |
```
bidoubiwa, line 79:
```suggestion
| indexUid   | string  | Unique index identifier. This field is `null` when the task type is `dumpCreation`. |
```

text/0060-tasks-api.md
```diff
@@ -176,6 +178,16 @@ The motivation is to stabilize the current `update` resource to a version that c
 | distinctAttribute | `distrinctAttribute` payload string |
 | displayedAttributes | `displayedAttributes` payload array |
 
+##### dumpCreation
+
+| name    | description  |
+| -----   | ------------ |
+| dumpUid | The generated uid of the dump |
+
+Since the creation of a dump is not a task associated with a particular index, it will only be visible on the `GET` - `/tasks` and `GET` - `tasks/:task_uid` endpoints.
```
bidoubiwa, line 187:
```suggestion
Since the creation of a dump is not a task associated with a particular index, it is only present on the `GET` - `/tasks` and `GET` - `tasks/:task_uid` endpoints.
```

text/0060-tasks-api.md
```diff
@@ -176,6 +178,16 @@ The motivation is to stabilize the current `update` resource to a version that c
...
+
+Fully qualified and summarized task objects related to a dump creation display a `null` `indexUid` field.
```
bidoubiwa, line 189: This sentence is a bit confusing :s

text/0060-tasks-api.md
```diff
@@ -586,4 +598,4 @@ New task types are also added for these operations. `indexCreation`, `indexUpdat
 - Add dedicated task type names modifying a sub-setting. e.g. `SearchableAttributesUpdate`.
 - Reconsider existence of `/indexes/:indexUid/tasks/:taskUid` route since it is similar to `/tasks/:taskUid`.
 - Add an archived state for old `tasks`.
-- Indicate the `API Key` identity that added a `task`. It should not permits to
\ No newline at end of file
+- Indicate the `API Key` identity that added a `task`.
```
bidoubiwa, line 601:
```suggestion
- Add the `API Key` identity that added a `task`.
```

bidoubiwa (changes requested):

text/0060-tasks-api.md
gmourier: Good point will mention that!

open-api.yaml
```diff
@@ -892,59 +869,17 @@ paths:
...
+                    indexUid: null
+                    status: enqueued
+                    type: dumpCreation
+                    enqueuedAt: '2021-07-19T14:31:16.920473Z'
         '401':
           $ref: '#/components/responses/401'
-  '/dumps/{dumpUid}/status':
```
curquiza, line 904: I think the PR description saying "/dumps/{dumpUid}/status is removed" is missing in the PR description
The PR description helps me a lot when creating issues for the Meilisearch milestones, since it gathers all the breaking changes and the main enhancements 😇

bidoubiwa:
My only concern is that a user asking for a dump will potentially wait hours before having the dump if a lot of tasks are in front in the dump. 

I don't know if this is problematic. 

Apart for that, great spec!

bidoubiwa (dismissed):

nicolasvienot (commented):
The spec is clear, thanks @gmourier!

> A dump creation is now an asynchronous task. This means that the dump creation task does not take priority over other tasks and that the dump will be created when the task store will not have any other processing task or when other enqueud tasks are not positioned before.

As @irevoire and @bidoubiwa noted, the fact that we can't prioritize a dump anymore and need to wait until all the previous tasks are processed is quite an issue as it means that we won't be able to launch an update directly in the cloud.

What's your opinion on this @qdequele ?

qdequele:
Hello @gmourier, after a discussion with @nicolasvienot about implementing the update feature (the ability to update a meilisearch version on the cloud manually) by the cloud team, we have some comments on this specification. 

**tl;dr:**
We concluded that it would be much simpler, and the update would be much smoother for the user if the dump creation task is prioritized and not pushed at the end of the task queue. 

**Explanation:**
When the cloud needs to update a meilisearch instance, the simplified process is the following. 
- We modify the proxy to prevent users from sending new documents. We will basically send errors to every call made in `POST`/`PUT`/`PATCH`/`DELETE` on the `/indexes/x/documents` or `/indexes/x/settings/*`
- We trigger a new dump creation on the old engine.
- Wait until the dump creation task is finished. 
- Download the dump and use it to launch a new server.
- Wait until the new server has indexed the dump and is ready to be used. 
- Redirect the proxy traffic to the new server. 

The goal is to update the server as fast as possible because the user will not be able to index new documents during the entire process. In a way, we don't want to wait until all tasks are finished because this time could be very long.

gmourier:
Hey @meilisearch/cloud-team 👋

I updated the specification to mention that, like today, the dumps task will be prioritized before the other enqueued tasks.

It will run as soon as the current processing task finishes, even if others have been enqueued before.

@kero, maybe in the future, we could bypass the current processing task and have a global read transaction to read the current state of the B+Trees of the different LMDBs (indexes) of a Meilisearch instance even while a write transaction is in progress. What do you think about it?

bidoubiwa (dismissed):
LGTM 👌

PR #140, opened 2022-04-06
Indexes API - Remove `name` field

gmourier:


text/0132-indexes-api.md
```diff
@@ -111,13 +102,10 @@ Fetch an index of a Meilisearch instance.
 | Field                    | Type                     | Required |
 |--------------------------|--------------------------|----------|
 | `uid`                    | string                   | true     |
```
maryamsulemani97, line 104: Is it possible to rename the `uid`s so you know what they're for? E.g., `index_uid`, `task_uid`, `dump_uid`, etc.

gmourier: I don't see much point in prefixing the variable name when you are literally in the context of a resource you are retrieving it would force us to prefix all fields of a resource to be consistent.

On the other hand, when you retrieve an object from another context it is interesting to give the information. Typically on this [issue ](https://github.com/meilisearch/product/issues/368)on 202 Summarized Task Response where `uid` could be misleading and `taskUid` permits to contextualize the given response.

PR #142, opened 2022-04-12
API Keys - `expiresAt` can't be null

gmourier:
🤖 [API Diff](https://github.com/meilisearch/specifications/pull/142#issuecomment-1127596975)

## Motivation

Following this initial discussion https://github.com/meilisearch/product/discussions/403, we tried with @bidoubiwa and @brunoocasali to find the best solution to handle `null` value on partial updates with strongly typed languages that omit them. 

Today, `null` is mostly used to reset fields to default values. E.g. For a partial update on `/indexes/:index_uid/settings`, `null` value can be omitted safely (this is already the case for strongly typed languages like C#) since we propose dedicated endpoints to reset a specific setting or the global settings to the default.

**We stated that SDK built with a strongly typed language will omit `null` value for partial updates to be consistent in all cases**

---

Since the initial issue concerns the `expiresAt` field that needs a `null` value as a "real" value to mark no expiration date it was problematic for strongly types languages during a partial update.

More than that, it was mentioned that it could be a security issue to not have to set an expiration date which is an important point to consider. We decided to make this expiration date to be set to a valid date.

The counterpoint to this is that it will force users to set an expiration date during the setup/development cycle and thus make them lose some time.

We evaluated several solutions and this one seemed to be the most obvious since it doesn't complicate the API much, doesn't require many changes, and adds security.

## Changes

- API Key must have an expiration date defined
- Default API Keys are set to `9999-12-31T00:00:00Z` when generated

---

## Migrate API Keys with `null` `expiresAt` from v0.27 to v0.28

We could migrate the API keys that do not have an expiration date by setting the same value as for the default API Keys like `9999-12-31T00:00:00Z`.


open-api.yaml
```diff
@@ -588,8 +588,7 @@ components:
         expiresAt:
           type:
             - string
-            - 'null'
-          description: Represent the expiration date and time as RFC 3339 format. null equals to no expiration time.
+          description: Represent the expiration date and time as RFC 3339 format.
```
brunoocasali, line 591:
```suggestion
          description: Represent the expiration date and time in RFC 3339 format.
```

bidoubiwa (approved):
LGTM 🔥

brunoocasali (commented):
I've checked most of our languages to verify if they will support the `9999-12-31`, and I think we are ok with that. 

Curiosity time: Python supports only [until 9999](https://docs.python.org/3/library/datetime.html). Citizens living during these years will have to update some legacy systems 😝

MarinPostma:
👌

MarinPostma (dismissed):

Kerollmops:
After a recent discussion with @gmourier, I think that we should disallow any update apart from the description of the keys, not being able to change the expiration date or the actions is a good security behavior.

gmourier:
It would also allow not to make `expirationAt` have a date value at creation, `null` could still be used to say that this key has no expiration date. The fact of not being able to update the fields of an API key except the `description` would not be a problem for typed languages. @brunoocasali  what do you think of this suggestion?

open-api.yaml
```diff
@@ -588,8 +588,7 @@ components:
         expiresAt:
           type:
             - string
-            - 'null'
```
curquiza, line 591: For me removing the `null` possibility is really painful for the users. And the current way of handling our keys does not solve the security issue. It means the users have to update the `expireAt` value -> the key is not changed and the security issue is still there (since the key does not change). It does not enforce the user at all to change the key, so to remove the security issue

For example, on GitHub side, with the API keys (called PAT on GitHub side) you can set an infinite expiration date. A warning is displayed, but I can
However, if I put an expiration date, I can never update the information once I created a key. Meaning I'm enforced to re-create a key if it expires, but I cannot change the expiration date.

brunoocasali: Removing the UPDATE request in the Meilisearch API will make the Integrations team's life easier. 
Still, on the other side, we will prevent the users from using keys more suitably for them like [**stripe**](https://stripe.com/docs/keys#limit-access) does, for example, they allow updates in particular fields. 

I could be wrong, but since we don't know exactly how the users will interact with our API's removing this capability seems to be not the best choice.

If we came up with a good solution for the key-rotation problem (see https://github.com/meilisearch/product/discussions/438), I think we can start thinking to remove the update method.

To sum up, I had a similar idea like @Kerollmops considered ([more here](https://github.com/meilisearch/product/discussions/403#discussioncomment-2503704)), and it seems to be the best in class because is able to solve the PATCH issue and to support the `null` value like Clem mentioned at the same time.

FYI: @gmourier

gmourier:
Closed in favor of #148 since `indexes`/`actions`/`expires_at` cannot be updated for an API Key. It should permit fixing the issue strongly typed language has with a `null` value during partial update of a resource without adding a weird default value to `9999-12-31T23:59:59` for the `expiresAt ` field.

PR #143, opened 2022-04-14
Documents API - `displayedAttributes` should not impact the documents API / Rename `attributesToRetrieve` to `fields`

gmourier:
🤖 [API Diff](url)

## Why?

The documents API is used to manage documents within an index.

`displayedAttributes` allows limiting the attributes that can be retrieved in the search but also impacts the endpoints of the documents API.

My proposal is to remove this behavior from the documents API. Indeed, in order to manage a Meilisearch instance, it is good to have all the information about it.

Today, it implies the user to modify `displayedAttributes` to access all the information which will also impact the search. This is not desirable for the developer experience.

---

Following this [discussion ](https://github.com/meilisearch/product/discussions/440)

`attributesToRetrieve` query parameter is renamed to `fields` on this API. Using a more generic name that does not collide with `attributesToRetrieve` from the `search` endpoint could help consistency if we add this feature on another API endpoints in the future.

Rejecting a field from a document in the given response with a negative operator is added as a future possibility. e.g. `?fields=-createdAt`

## Changes

- `displayedAttributes` do not impact the documents API.
- Rename `attributesToRetrieve` on the document API to `fields`. It's more generic and thus can be used on all routes other than search in the future as a query parameter component.

### DIFF
--- a/open-api.yaml
+++ b/open-api.yaml
@@ -156,7 +156,7 @@ components:
           type:
             - string
             - number
-          description: Retrieve attributes of the document. `attributesToRetrieve` controls these fields.
+          description: Retrieve attributes of a search result. `attributesToRetrieve` controls these fields.
         _geoDistance:
           type: number
           description: 'Using _geoPoint({lat}, {lng}) built-in sort rule at search leads the engine to return a _geoDistance within the search results. This field represents the distance in meters of the document from the specified _geoPoint.'
@@ -815,6 +815,16 @@ components:
 
         > warn
         > Attribute(s) used in `filter` should be declared as filterable attributes. See [Filtering and Faceted Search](https://docs.meilisearch.com/reference/features/filtering_and_faceted_search.html).
+    fields:
+      name: fields
+      in: query
+      description: 'Comma-separated list of fields to display for an API resource. By default it contains all fields of an API resource.'
+      schema:
+        type: string
+        items:
+          type: string
+        example: 'uid,createdAt'
+        default: '*'
     Content-Type:
       name: Content-Type
       in: header
@@ -1255,7 +1265,7 @@ paths:
       parameters:
         - $ref: '#/components/parameters/limit'
         - $ref: '#/components/parameters/offset'
-        - $ref: '#/components/parameters/attributesToRetrieve'
+        - $ref: '#/components/parameters/fields'
     post:
       operationId: indexes.documents.create
       summary: Add or replace documents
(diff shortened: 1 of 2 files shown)


text/0124-documents-api.md
```diff
@@ -451,4 +453,5 @@ The auth layer can return the following errors if Meilisearch is secured (a mast
 N/A
 
 ## 5. Future Possibilities
-N/A
\ No newline at end of file
+
+- Introduce a way to reject fields from a document in the response. e.g. `?fields=-createdAt`
```
brunoocasali, line 457: This option is really interesting, but I think it could add a lot of complexity,
The jsonapi spec specifies this behavior as sparse fieldsets: https://jsonapi.org/format/#fetching-sparse-fieldsets but I did not find a specific part about having the negation behavior of this option, so I think if they don't have it we don't need it too :)

open-api.yaml
```diff
@@ -748,6 +748,16 @@ components:
 
         > warn
         > Attribute(s) used in `filter` should be declared as filterable attributes. See [Filtering and Faceted Search](https://docs.meilisearch.com/reference/features/filtering_and_faceted_search.html).
+    fields:
+      name: fields
+      in: query
+      description: 'Comma-separated list of fields to display for an API resource. By default it contains all fields of an API resource.'
```
brunoocasali, line 754: The fields list is case-sensitive?

open-api.yaml
```diff
@@ -748,6 +748,16 @@ components:
...
+      schema:
+        type: string
+        items:
+          type: string
+        example: 'uid,createdAt'
```
brunoocasali, line 759: Just checking:

`?fields=` => 200 OK, full document object.
`?fields=wrongName` => 200 OK, full document object.

This is correct?

open-api.yaml
gmourier: That's a good question, and I think it should ease the usage.

@Kerollmops It is already the case natively? I know that query parameter are not so easy to handle due to some limitations.

open-api.yaml
gmourier: Good one @brunoocasali; I think it should be equivalent to the current attributesToRetrieve behavior.

- Not sending the `fields` parameter = `fields=*`
- Sending `fields=` returns empty documents
- Sending `fields=wrongField` returns empty documents
- Sending `fields=wrongField,fieldThatExists` return documents with fieldThatExists displayed

WDYT?

text/0124-documents-api.md
gmourier: Yes, this is a bonus that may not be useful. I see it in case you have a lot of fields, and instead of specifying the ones you want, it becomes easier to remove the ones you don't want.

open-api.yaml
brunoocasali: Perfect for me!

PR #149, opened 2022-05-05
Geosearch - Enhance lat/lng format error messages

irevoire:
It’s also important to note that the first error message could not stay the same (and show the whole object) as before because we lost this information with the v0.27 and the nested fields 😢


gmourier:
Do things need to change regarding this due to other enhancements we have made? "as before because we lost this information with the v0.27 and the nested fields"

gmourier (approved):

PR #145, opened 2022-04-26
Telemetry - Add `x-meilisearch-client` query parameter

gmourier:
## Why?

Following this [comment](https://github.com/meilisearch/integration-guides/issues/150#issuecomment-966360488) pointing out that some browsers forbid the customization of the User-Agent header, the analytics stack can now take the value of a `x-meilisearch-client` query parameter if found instead of the user-agent header to track the usage of our SDKs/Integrations.

PR #150, opened 2022-05-19
Telemetry - Replace `x-meilisearch-client` query parameter by `X-Meilisearch-Client` header

gmourier:
## Why?

Following #145, the core team had some technical difficulties implementing it.

We decided to replace the `x-meilisearch-client` query parameter with a custom header.

According to our exploration, a custom header is rarely stripped by a proxy by default (_tested on nginx_). If it still happens, the analytics will just not be collected in this case.

We found this solution to be the least impactful regarding the user experience while ensuring that it can be implemented quickly without rewriting how the API passes an `x-meilisearch-client` query parameter for a non-business field to the end-user POV.

### DIFF
--- a/text/0034-telemetry-policies.md
+++ b/text/0034-telemetry-policies.md
@@ -416,11 +416,11 @@ The `User-Agent` header is tracked on the events listed below. Our official SDKs
 
 Each endpoint API tracked sends the `User-Agent` as a `user_agent` event property as an array. If several values are contained in the `User-Agent` header, they are split by the `;` character.
 
-##### `x-meilisearch-client` Query Parameter
+##### `X-MEILISEARCH-CLIENT` Header
 
-Some browser engines prevent overloading the User-Agent header. To track the calls made by some clients concerned by this fact, e.g. the JavaScript SDK, it is possible to use the `x-meilisearch-client` query parameter.
+Some browser engines prevent overloading the User-Agent header. To track the calls made by some clients concerned by this fact, e.g. the JavaScript SDK, it is possible to use the `X-MEILISEARCH-CLIENT` custom header.
 
-If the query parameter is encountered, it overrides the presence of the `User-Agent` header.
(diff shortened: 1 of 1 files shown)


text/0034-telemetry-policies.md
```diff
@@ -417,11 +417,11 @@ The `User-Agent` header is tracked on the events listed below. Our official SDKs
 
 Each endpoint API tracked sends the `User-Agent` as a `user_agent` event property as an array. If several values are contained in the `User-Agent` header, they are split by the `;` character.
 
-##### `x-meilisearch-client` Query Parameter
```
brunoocasali, line 420: Disclaimer, I like the name `X-MEILISEARCH-CLIENT`.

But it is not mandatory to have the `X-` prefix https://stackoverflow.com/a/3561399/2649707

brunoocasali (dismissed):
Let's do this! 👯

text/0034-telemetry-policies.md
gmourier: I think it's safe to say that `X-MEILISEARCH-CLIENT` will not become a standard.

I like the idea of `X-` because like most people I don't know the RFC in every detail and I know right away that it is often a custom header introduced by the API used.

Nevertheless, this is only my opinion, and let technical folks decide with a vote.

👍: To keep X-
👎  To remove it

irevoire: WHY IS IT WRITTEN IS CAPS LOCK THOUGH?
SHOULDN'T IT BE SOMETHING LIKE `X-Meilisearch-Client`? :joy:

gmourier: I DON'T KNOW 🥲, I THINK YOU ARE RIGHT

irevoire (approved):
Thanks!

brunoocasali (approved):
🎉

MarinPostma:
guys, http headers are case insensitive https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive

PR #155, opened 2022-05-31
Make the /documents/:uid route accept the fields query parameter

irevoire:


text/0124-documents-api.md
```diff
@@ -165,6 +184,7 @@ A document represented as a JSON object.
 
 - 🔴 If the requested `index_uid` does not exist, the API returns an [index_not_found](0061-error-format-and-definitions.md#index_not_found) error.
 - 🔴 If the requested `document_id` does not exist, the API returns an [document_not_found](0061-error-format-and-definitions.md#document_not_found) error.
+- 🔴 Sending a value with a different type than `String` or `null` for `attributesToRetrieve` will return a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
```
gmourier, line 187:
```suggestion
- 🔴 Sending a value with a different type than `String` or `null` for `fields` will return a [bad_request](0061-error-format-and-definitions.md#bad_request) error.
```

gmourier (changes requested):

PR #157, opened 2022-06-08
Settings API - Customize the hard limits for `pagination` and `faceting`

gmourier:
🤖  [API Diff](https://github.com/meilisearch/specifications/pull/157#issuecomment-1150004644)

---

# Summary

This pull request introduces two new settings API to customize the current hard limits that are blocking users to adjust the search engine behavior to their custom needs.

## `pagination` settings API

Before v0.27, we didn't have a limit on the number of reachable documents for a query. A handful of users complained that this made their database scrappable or subject to DDoS attack, so we decided to introduce a limit of `1000` for v0.27, non-customizable.

Our hypothesis (Some users might need to customize it at a superior range) proved to be true, some users need to customize this limit. 

The main argument is that some users need to make the search results scrappable by search engines for their SEO.

Some users also simply want to go further in the pagination despite the relevancy of the first pages.

No one has yet asked to customize it to a lower value, but we imagine this is possible, especially for those who want to reduce the “scrappability” of their dataset.

## `faceting` settings API

Since v0.21, there is no more limit on the number of values returned per facet. This was the case before v0.21 but some users found the default value too low, so it was decided to remove this limit.

The fact that it is unlimited today poses performance problems for users who have many possible values for a facet.

Rather than blocking one or the other, we make this value customizable according to their needs.

## Important

Increasing the default values of these properties can degrade performance and more easily expose an instance to DDoS attacks or scrapping by a malicious user.

The default values should be a good compromise between performance and usage for most users.

---

# Changes

## API Endpoints

- Introduce a `pagination` settings API resource. `GET/PATCH/DELETE` - `/indexes/:indexName/settings/pagination`
    - `maxTotalHits` property - Type: Integer - Default to `1000`
- Introduce a `faceting` settings API resource. `GET/PATCH/DELETE` - `/indexes/:indexName/settings/faceting`
    - `maxValuesPerFacet` property - Type: Integer - Default to `100`
- `pagination` and `faceting` objects are accessible from the global settings endpoints `/indexes/:indexName/settings`

## Telemetry Datapoints

- `pagination.max_total_hits` and `faceting.max_values_per_facet` are as a property of the `Settings Updated` event.
- `Pagination Updated` event is introduced with the `max_total_hits` property.
- `Faceting Updated`event is introduced with the `max_values_per_facet` property

# Out Of Scope

n/a

---

# Attention To Reviewers

Introduced fields like `maxTotalHits` and `maxValuesPerFacet` are open to suggestion.

---

## Misc


text/0034-telemetry-policies.md
```diff
@@ -349,12 +353,24 @@ This property allows us to gather essential information to better understand on
...
+
+| Property name | Description | Example |
+|---------------|-------------|---------|
+| pagination.limited_to       | The defined value for `limitedTo` property | `1000` |
+
+## `Faceting Updated`
+
+| Property name | Description | Example |
+|---------------|-------------|---------|
+| pagination.max_values_per_facet | The defined value for `maxValuesPerFacet` property | `100` |
```
Kerollmops, line 372:
```suggestion
| faceting.max_values_per_facet | The defined value for `maxValuesPerFacet` property | `100` |
```

Kerollmops (changes requested):

open-api.yaml
```diff
@@ -2227,6 +2263,142 @@ paths:
...
+          description: Ok
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/faceting'
+        '401':
+          $ref: '#/components/responses/401'
+        '404':
+          description: Not Found
+    patch:
```
brunoocasali, line 2355: Following the [spec #152](https://github.com/meilisearch/specifications/pull/152) we should use `PUT` for these routes, correct?

text/0034-telemetry-policies.md
```diff
@@ -55,6 +55,8 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
 | FilterableAttributes Updated | Occurs when filterable attributes are updated via `POST` - `/indexes/:indexUid/settings/filterable-attributes`. |
 | SortableAttributes Updated | Occurs when sortable attributes are updated via `POST` - `/indexes/:indexUid/settings/sortable-attributes`. |
 | TypoTolerance Updated | Occurs when typo tolerance settings are updated via `POST` - `/indexes/:indexUid/settings/typo-tolerance`. |
+| Pagination Updated | Occurs when pagination settings are updated via `PATCH` — `/indexes/:indexUid/settings/pagination`. |
```
brunoocasali, line 58: Same here related to the PUT/PATCH.

text/0034-telemetry-policies.md
```diff
@@ -55,6 +55,8 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
...
+| Faceting Updated | Occurs when faceting settings are updated via `PATCH` — `/indexes/:indexUid/settings/faceting`. |
```
brunoocasali, line 59: Same here related to the PUT/PATCH.

brunoocasali (changes requested):
@gmourier give a look at the PATCH/PUT definition here I think it is not compatible with this one https://github.com/meilisearch/specifications/pull/152 😄 .

open-api.yaml
gmourier: I had a doubt about using PATCH/PUT for the following reason.

There is a very high chance that we will add parameters to the `pagination` and `facetting` objects in the future, by setting PATCH I wanted to avoid a future breaking and thus directly allow to edit each of the properties partially.

What do you think Bruno? I have alternated between PUT and PATCH to choose PATCH.

bidoubiwa:
> limited_to

What do you think of `max_total_hits` or `total_hit_limit`.

It goes in line with `estimatedTotalHits` 

> maxValuesPerFacet

Makes me think we can customize this value for every facet. Like `genres: 20`, `color: 100` 

If this is not the case, what about something more generic like: `facet_values_limit` or `max_facet_values` 

?

dichotommy:
> limited_to
> What do you think of max_total_hits or total_hit_limit.

Strong agree with `max_total_hits` 👍🏻 

> maxValuesPerFacet
> Makes me think we can customize this value for every facet. Like genres: 20, color: 100

For me, the name `maxValuesPerFacet` is really clear. When I read `maxValuesPerFacet: x`, I think that the engine will show only the `x` most common facet values for each attribute defined in `filterableAttributes`.

open-api.yaml
brunoocasali: I understand your point now, and it seems a good choice from my side. Making decisions with data that does not exist (yet) is hard because you never know when you are doing overengineering. 
I don't know if these new attributes will appear in the future but I trust you! 🤝

text/157-faceting-setting-api.md
```diff
@@ -0,0 +1,146 @@
...
+- Default: `100`
+
+Define maximum number of value returned for a distributed facet.
+
+e.g. It means that with the default value of `100`, it is not possible to have `101` colors if the color field is defined as a facet at search time.
+
+The value of 100 ensures good performance and prevents malicious users from scraping data from a Meilisearch instance.
+
+Increasing this value can degrade performance as well as expose the data of an instance to scrapping.
+
```
Kerollmops, line 34:
```suggestion

The facets that are returned are not specified, they are not sorted in a particular way.

```

Kerollmops (changes requested):

gmourier:
Thanks for your comments on the names! Do you have a better name than `pagination` knowing that the pain points about this limit have been expressed for this use case and that we will probably add more parameters about it? Unfortunately, I don't have a better suggestion to make.

gmourier:
Waiting for the last naming changes if needed on https://github.com/meilisearch/specifications/pull/157 before giving a final review. Other than that, everything looks great to me!

dichotommy:
Hi @gmourier , if you know that you will add more pagination-related settings in the future and wish to keep this one within that category, then I completely understand and support the naming decision.

On the other hand, I think this setting can be useful and interesting to people besides those who need it for pagination. One example that you've already brought up is security, to prevent scraping of the database. There may be other cases we haven't considered where users want to enforce a base `limit` on all search queries without using the parameter. My concern would be that users could miss this setting if it's in an object called `pagination`.

I think this setting would be easier to find as an int at the root level of the settings resource with a name like `resultsLimit`, `maxResults`, `maxHits` or similar. Just my two cents.

bidoubiwa:
Poll time! 

## Limit the number of hits on which we can paginate

Little context, currently it would be inside a `pagination` object. For example: 

```js
{
   pagination: {
      limited_to: 100
   }
}
```

This might change depending on another poll.

- `limited_to` 👍
- `max_total_hits` 🎉
- `total_hit_limit` 🚀
- other 👀

bidoubiwa:
## Maximum values returned by each facet

Little context, currently it would be inside a pagination object. For example:
```js
{
   faceting: {
      maxValuesPerFacet: 100
   }
}
```

- `maxValuesPerFacet` 👍
- `facetValuesLimit` 🚀
- `maxFacetValues` 🎉

gmourier:
Thanks to all who voted 🙌

The polls are now closed.

Here are the final changes based on the votes:

- `limitedTo` is renamed `maxTotalHits`.

PR #161, opened 2022-07-13
Geosearch - Add error variants for `invalid_geo_field` error

irevoire:
# Summary

Add variants to help users when the `_geo` field is invalid.

- Add a variant in the case the provided `_geo` is not an object.
- Add a variant in the case `_geo.lat` **and** `_geo.lng` fields are missing. 

---

# Changes

- Update `text/0061-error-format-and-definitions.md` to add two new variants for `invalid_geo_field` error.

# Out Of Scope

N/A

---

# Attention To Reviewers

N/A

---

## Misc

Implemented in https://github.com/meilisearch/milli/pull/561

### DIFF
--- a/text/0061-error-format-and-definitions.md
+++ b/text/0061-error-format-and-definitions.md
@@ -731,7 +731,29 @@ These errors occurs when the `_geo` field of a document payload is not valid. Ei
 
 ### Error Definition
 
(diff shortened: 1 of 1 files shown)


gmourier (approved):
LGTM! 🧱

PR #165, opened 2022-08-10
Remove `name` from indexes resource definition

bidoubiwa:
Remove the `name` from the index type definition as the field does not exist anymore.


gmourier (approved):
Thanks @bidoubiwa !!!

PR #172, opened 2022-08-18
Health API — Catch up

gmourier:
# Summary

Catch up on missing spec for the `/health` endpoint.

---

# Changes

- Add `0172-health-api.md` specification file

# Out Of Scope

n/a

---

# Attention To Reviewers

n/a

### DIFF
--- a/text/0172-health-api.md
+++ b/text/0172-health-api.md
@@ -1,0 +1,53 @@
+# Health API
+
+## 1. Summary
+
+This specification describes the health API endpoint. The `/health` route allows to verify the status and availability of a Meilisearch instance.
+
+## 2. Motivation
+
+To know the status of a Meilisearch instance, `/health` is a public route that can be long pooled with simple monitoring tools.
+
+## 3. Functional Specification
+
+### 3.1. `health` API resource properties
+
+| Field                                            | Type            | Required |
+|--------------------------------------------------|-----------------|----------|
+| [status](#311-status)                            | String          | True     |
+
+#### 3.1.1. `status`
+
+- Type: String
+- Required: True
+- Default: `available`
+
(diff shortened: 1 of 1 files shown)


text/0172-health-api.md
```diff
@@ -0,0 +1,53 @@
...
+
+To know the status of a Meilisearch instance, `/health` is a public route that can be long pooled with simple monitoring tools.
+
+## 3. Functional Specification
+
+### 3.1. `health` API resource properties
+
+| Field                                            | Type            | Required |
+|--------------------------------------------------|-----------------|----------|
+| [status](#311-status)                            | String          | True     |
```
irevoire, line 17: `Required`?
It's not a parameter, you don't have a choice, we should remove this last column right?

text/0172-health-api.md
```diff
@@ -0,0 +1,53 @@
...
+
+#### 3.1.1. `status`
+
+- Type: String
+- Required: True
+- Default: `available`
```
irevoire, line 23:
```suggestion
- Value: `available`
```

I don't know what it's worth since you say it literally one line later, but we could be even more explicit by giving its value instead of its « default » value.

irevoire (changes requested):
I don't understand why there is the last column in the table.
Everything else is great thanks 👍

text/0172-health-api.md
gmourier: The boolean required is also mentioned in the response definition of resources because we sometimes have conditional response fields (for example in the search).

text/0172-health-api.md
gmourier: Yes, this is a special case, there can be no other value than `available` so I put it in default to reuse an existing format in our specifications. It seemed appropriate but it can be changed!

PR #173, opened 2022-08-22
Search API — Add `matchingStrategy` parameter with `last` / `all` strategies

gmourier:
🤖  [API Diff](https://github.com/meilisearch/specifications/pull/173#issuecomment-1222297686)

---

# Summary

Implement the ability to choose a word matching strategy at search time between `last` and `all` strategies.

We know that some users are still stuck in pre-v0.21 versions because of the hard implementation of the `last` mode introduced with v0.21. We are confident that adding the `all` strategy will permit those users to move to the newer versions while giving more versatility to Meilisearch regarding relevancy tweaking.

---

# Changes

- Add a `matchingStrategy` parameter to POST and GET `search` endpoints
      - `last` is the default strategy if the field is not specified
      - 2 strategies are available: `last` and `all`.
- `most_used_matching_strategy` is added as a telemetry node on POST and GET `/search` events.

# Out Of Scope

Proposing and introducing other modes than `all` and `last` is out-of-scope. A [product discussion](https://github.com/meilisearch/product/discussions/255) exists to collect and discuss other strategies.

---

# Attention To Reviewers

- ~~@meilisearch/docs-team `wordMatchingStrategy` parameter can change. We also proposed `termMatchingStrategy`. Please validate/correct this proposal or propose another one that seems more suited if you have a better idea 💡~~ **(addressed in the following comments)**
- ~~@meilisearch/core-team @meilisearch/integration-team Should we put this parameter in a parent object at search to accommodate other fields related to this topic in the future? There is a **good chance** we will propose this parameter as an index setting in the future. Storing this parameter and other related parameters in a parent-object avoids generating a new API route per parameters on the settings side, which could impact the search endpoints regarding consistency and thus add a breaking change if we don't do it now.~~ **(addressed in the following comments)**

Example:

`wordMatching` is just an example; there may be a better name for that parent object concept.

At search:

```jsonc
{
      "wordMatching": {
            "wordMatchingStrategy": "all",
            "disablePrefixSearchOnAttributes": ["title", ...] //exemple of a potential future,
            "...": ...
      }
}
```

If needed, SDKs could still implement facade methods to obfuscate the `wordMatching` parent.

Doing so would permit us to easily add new fields in the future while keeping a well-organized API matching the index settings.

In index settings API: 

`GET - /indexes/:indexUid/settings/wordMatching`
`PATCH - /indexes/:indexUid/settings/wordMatching` 
`DELETE - /indexes/:indexUid/settings/wordMatching`

Instead of: 

`GET - /indexes/:indexUid/settings/wordMatchingStrategy`
`GET - /indexes/:indexUid/settings/disabledPrefixSearchOnAttributes`
`PUT - /indexes/:indexUid/settings/wordMatchingStrategy`
`PUT- /indexes/:indexUid/settings/disabledPrefixSearchOnAttributes`
`DELETE - /indexes/:indexUid/settings/wordMatchingStrategy`
`DELETE - /indexes/:indexUid/settings/disabledPrefixSearchOnAttributes`
...

and so on for every field that may be added in the future that could have been organized under a parent object scoping the same "layer".

WDYT?

---

## Misc

### DIFF
--- a/open-api.yaml
+++ b/open-api.yaml
@@ -847,6 +847,17 @@ components:
         type: boolean
         default: 'false'
       description: Defines whether an `_matchesPosition` object that contains information about the matches should be returned or not.
+    wordMatchingStrategy:
+      name: wordMatchingStrategy
+      in: query
+      required: false
+      schema:
+        type: string
+        enum:
+          - last
+          - all
+        default: 'last'
+      description: Defines which strategy to use to match the query terms within the documents as search results. Two different strategies are available, `last` and `all`. By default, the `last` strategy is chosen.
     sort:
       name: sort
       in: query
@@ -1620,6 +1631,7 @@ paths:
         - $ref: '#/components/parameters/sort'
         - $ref: '#/components/parameters/limit'
         - $ref: '#/components/parameters/showMatchesPosition'
+        - $ref: '#/components/parameters/wordMatchingStrategy'
       responses:
         '200':
           description: Ok
@@ -1724,6 +1736,10 @@ paths:
                   type: boolean
                   description: Defines whether an `_matchesPosition` object that contains information about the matches should be returned or not.
                   default: false
+                wordMatchingStrategy:
+                  type: string
+                  description: Defines which strategy to use to match the query terms within the documents as search results. Two different strategies are available, `last` and `all`. By default, the `last` strategy is chosen.
+                  default: 'last'
                 filter:
                   $ref: '#/components/schemas/filter'
                 facets:
@@ -1762,6 +1778,7 @@ paths:
                   attributesToHighlight:
                     - overview
                   showMatchesPosition: true
+                  wordsMatchingStrategy: all
       responses:
         '200':
           description: Ok
--- a/open-api.yaml
+++ b/open-api.yaml
@@ -847,8 +847,8 @@ components:
         type: boolean
         default: 'false'
       description: Defines whether an `_matchesPosition` object that contains information about the matches should be returned or not.
-    wordMatchingStrategy:
-      name: wordMatchingStrategy
+    matchingStrategy:
+      name: matchingStrategy
       in: query
       required: false
       schema:
@@ -1631,7 +1631,7 @@ paths:
         - $ref: '#/components/parameters/sort'
         - $ref: '#/components/parameters/limit'
         - $ref: '#/components/parameters/showMatchesPosition'
-        - $ref: '#/components/parameters/wordMatchingStrategy'
+        - $ref: '#/components/parameters/matchingStrategy'
       responses:
         '200':
           description: Ok
@@ -1736,7 +1736,7 @@ paths:
                   type: boolean
                   description: Defines whether an `_matchesPosition` object that contains information about the matches should be returned or not.
                   default: false
-                wordMatchingStrategy:
+                matchingStrategy:
                   type: string
                   description: Defines which strategy to use to match the query terms within the documents as search results. Two different strategies are available, `last` and `all`. By default, the `last` strategy is chosen.
                   default: 'last'
--- a/text/0034-telemetry-policies.md
+++ b/text/0034-telemetry-policies.md
@@ -135,7 +135,7 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
 | `filtered_by_type`                      | `true` if `GET /tasks` endpoint is filered by `type`, otherwise `false` | false | `Tasks Seen` |
 | `filtered_by_status`                    | `true` if `GET /tasks` endpoint is filered by `status`, otherwise `false` | false | `Tasks Seen` |
 | `per_index_uid` | `true` if an uid is used to fetch an index stat resource, otherwise `false` | false | `Stats Seen` |
-| `most_used_word_matching_strategy`      | Most used word matching strategy among all search requests in this batch | `last` | `Documents Searched POST`, `Documents Searched GET` |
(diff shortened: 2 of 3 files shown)


open-api.yaml
```diff
@@ -813,6 +813,17 @@ components:
         type: boolean
         default: 'false'
       description: Defines whether an `_matchesPosition` object that contains information about the matches should be returned or not.
+    wordMatchingStrategy:
+      name: wordMatchingStrategy
```
curquiza, line 817:
```suggestion
      name: matchingStrategy
```

curquiza: What about a smaller name like this?

gmourier: In my opinion, it works too!

bidoubiwa (commented):
## Nested object

> Should we put this parameter in a parent object at search to accommodate other fields related to this topic in the future? 

Going into this direction opens the door to many more sub-objects in the future (ex: pagination, filtering, ...).  Adding this as a normal and expected behavior, we can also imagine in the future objects whose keys are fields: 

```json
{
   "myParameter": { "title": "something", "description": "something" }
}
```

Another argument in favor of nesting is that we can create a bundle of required parameter. For example, my object `parameter_1` requires the field `x` and `y`. It is easier to raise an error when the query is formatted like this:

```json
{
   "my_parameter": {
      "x": true,
    }
}
```
=> `ERROR: y is missing in parameter_1`

Instead of 

=> `ERROR: when using myParameterX, you need to also defined myParameterY`

Not sure this would every happen though, but it is clearer.

BUT! 
The GET route becomes harder to use, but it is barely used so maybe asking them to URL encode their parameter is a price we are willing to pay (and the JS SDK which is the only one allowing get can compensate with its own API).

Screenshot 2022-08-22 at 16 55 51

Also, It introduces nesting complexity, but for me, if we stay at a one deepness level of complexity it might still be okay. 

Nesting is not a strong opinion so if you prefer not, I'm okey with that.

## Setting

>  There is a good chance we will propose this parameter as an index setting in the future.

Since nesting are already a thing in the settings, we can imagine using the same naming in both which would make it easier for the user to recognize that these are the same.

>  which could impact the search endpoints regarding consistency and thus add a breaking change if we don't do it now

Not sure I understand this part. Do you mean if later on we decide to put them in a sub-object it becomes a breaking?

## Naming

`wordMatchingStrategy` works for me! It may be longer than `matchingStrategy` but it is maybe more intuitive when it is read.

Kerollmops:
I was wondering why we were using `wordsMatchingStrategy` (🚀) and not `wordsMatchingPolicy` (❤️) ?

curquiza:
I think I prefer the nested setting for many reasons

Pros
- avoid creating new GET/POST/DELETE routes for each "matching policy" setting we add. It can be really tedious for docs team and integration team.
- it brings context: you understand the settings is about the matching strategy even if you see "disablePrefix", or "optionalWords".
See below: when I see the [`advancedSyntax` setting](https://www.algolia.com/doc/api-reference/api-parameters/advancedSyntax/) of Algolia, you got nothing, but fortunately, they put the settings under the "query strategy" section. If you don't read the docs, you cannot even guess what it is about.

Capture d’écran 2022-08-23 à 16 09 01

- it allows us to have smaller setting names since the context is already described


Cons
- the GET route is indeed harder to use, but since it's not impossible, it's really ok because we have few people using it.
- if we don't add any other settings it's useless and even "dumb" but I cannot imagine the future Meilisearch without another setting in the sub-setting (ex: `disablePrefix`, `disableExact`...)

⚠️ In a world where we don't have the sub-settings routes, I'm not sure my decision will be the same, but this point make me prefer the nested settings

---

## Naming suggestions

For the root name
- `matchingStrategy`
- `queryStatregy`
- `matchingPolicy`

For the sub-setting
- `optionalWords`
- `matchingMode`
- ...

gmourier:
If we decide to go for a nested object, we should consider doing it for others for the sake of consistency at some time, e.g

```json
{
    "q": "mySearch",
     "formatting": {
          "attributesToCrop": [],
          "cropLength": 10,
          "attributesToHighlight": [],
          "highlightPreTag": "<cool>",
          "...": "etc"
     }
}
```

Here is a scenario that would allow us not to avoid breaking changes on the search endpoint until a v2.0.

We could keep all the parameters at the root on the `/search` endpoint as we do today; if we add one of the already existing search parameters as an index setting, we add it in a parent object on the settings side to avoid generating at least 3 settings endpoints for each parameter (GET, PUT, DELETE) that a parent could contextualize.

Concretely, in the future, we could add an index setting named `formatting` to handle `attributesToHighlight`, `attributesToCrop`, `cropLength`, etc... and leave them at the root level of the `search` endpoint until a major version to break the `/search` endpoint.

We could also keep this way of doing things without ever changing it. Keeping settings organized under a parent object while keeping search parameters at the root level.

Or... we could break it before the v1.0 if we firmly believe that the `search` parameters organization must match the `settings` parameters organization.

curquiza:
I really would rather we don't break the search for not really impactful changes. I said "not really impactful" because we are indeed not sure this is better for the user or whatever. It's worthy to break for a huge complaint.
I think it's less breaking to remove the sub-setting routes than to change the search. The search is the most used route, and each time we made breaking on it, it was painful for the users. Plus, we promise stabilization to the users and the cloud team. Breaking the search is the worst move to break the promise 😂

gmourier:
Ok, so the strategy is:

- Keep search parameters at the root level (not breaking)
- We can consider adding search parameters as index settings in a nested if needed (not breaking, avoid multiplying the endpoints on the `/settings`).
- If nested index settings need to be provided at search, they will be displayed at the root level. (not breaking, keep consistency)

gmourier:
🥁 `matchingStrategy` wins!

PR #181, opened 2022-10-18
Keys API - Add a variant message when an index uid format is invalid for the indexes field.

gmourier:
Use the `invalid_index_uid` error message as a variant for `invalid_api_key_indexes` error when an invalid index uid is set into the API Key `indexes` field.

### Related issue

https://github.com/meilisearch/meilisearch/issues/2158
https://github.com/meilisearch/meilisearch/issues/2924

### DIFF
--- a/text/0061-error-format-and-definitions.md
+++ b/text/0061-error-format-and-definitions.md
@@ -294,6 +294,19 @@ HTTP Code: `400 Bad Request`
 
 - The `:value` is inferred when the message is generated.
 
+#### Variant: Sending an invalid index uid format in `indexes` field.
+
+```json
+{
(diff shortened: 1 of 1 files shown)


curquiza:
I'm not sure to get the purpose of this spec? is it a catch up? Looks like it's already implemented since the issue is closed in v0.28.0?

gmourier:
The previous spec never merged, but I saw the related issue as fixed in a previous release. When I tested it, it was not solved, so with the new process, I had to rewrite it based on release-v0.30.0 to avoid a huge rebase 🤯 So it can be as a catch-up that needs to be fixed for v0.30.0

curquiza:
Where is the open related issue about this bug? Shouldn't we reopen the new one?

ManyTheFish (approved):

PR #187, opened 2022-10-18
Instance Options - Precise that `--max-index-size` is for an index and not all indexes.

gmourier:
# Summary

Precise that `--max-index-size` is for an index and not all indexes.

Related to https://github.com/meilisearch/product/issues/530

---

# Changes

See Summary Section.

# Out Of Scope

N/A

---

# Attention To Reviewers

N/A

---

## Misc
N/A


curquiza (approved):
FYI the related issue is not only about this change but makes all the CLI option documentation consistent with the official documentation https://github.com/meilisearch/meilisearch/issues/2810

PR #188, opened 2022-10-18
Tasks API - Rename `receivedDocumentsIds` by `matchedDocuments`

gmourier:
🤖  [API Diff]() _Put the link of the GitHub comment generated by bump.sh if generated; Apply the `OpenApi` label_

---

# Summary

This name will make sense when it will be possible to delete documents via filters.

`matchedDocuments` seems more appropriate regarding all the methods available for deleting documents, whereas `receivedDocumentsIds` is more limited and indicate that ids are always received, which will not be true.

---

# Changes

- open-api.yml

# Out Of Scope

N/A

---

# Attention To Reviewers

N/A

---

## Misc


open-api.yaml
```diff
@@ -323,9 +323,9 @@ components:
             indexedDocuments:
               type: integer
               description: Number of documents finally indexed for documentAdditionOrUpdate task or a documentAdditionOrUpdate batch of tasks.
-            receivedDocumentsIds:
```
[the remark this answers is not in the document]
curquiza: `matchedDocuments` is returned for `delete-batch` task, and was chosen because, in the future, we expect to allow the deletion of documents by filters when deleting documents (with `delete-batch` or not)
`receivedDocuments` here are literally the documents sent by the users so received by meilisearch

For me it's consistent. Maybe @gmourier has a different opinion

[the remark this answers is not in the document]
gmourier: This is only for the `documentDeletion` task type. Since there are multiple ways of deleting documents, by an id, by ids, and in the future by a filter, it seems better to call it `matchedDocuments`

`matchedDocuments` will be less precise in its naming than `receivedDocumentsIds` in the case of deletion by identifier but will be much more straightforward for deletion by a filter.

brunoocasali (approved):
LGTM 🥇

open-api.yaml
```diff
@@ -323,9 +323,9 @@ components:
...
+            matchedDocuments:
```
Kerollmops, line 326:
```suggestion
            providedIds:
```

As discussed on Slack, we decided that it would be more correct to use this naming as it doesn't represent the number of documents that match but raw documents ids, the number of existing documents is not known at this point.

open-api.yaml
```diff
@@ -323,9 +323,9 @@ components:
...
               type: integer
-              description: Number of document ids received for documentDeletion task.
+              description: Number of document ids matched for documentDeletion task.
```
Kerollmops, line 328:
```suggestion
              description: Number of provided document ids for the documentDeletion task.
```

text/0060-tasks-api.md
```diff
@@ -84,7 +84,7 @@ As writing is asynchronous for most of Meilisearch's operations, this API makes
 
 | name                | description                          |
 |---------------------|--------------------------------------|
-| receivedDocumentIds | Number of document ids received.     |
+| matchedDocuments    | Number of document ids matched.      |
```
Kerollmops, line 87:
```suggestion
| ProvidedIds    | Number of provided document ids.      |
```

Kerollmops (changes requested):
As discussed on Slack, we decided that it would be more correct to use this naming as it doesn't represent the number of documents that match but raw documents ids, the number of existing documents is not known at this point.

text/0060-tasks-api.md
gmourier:
```suggestion
| providedIds    | Number of provided document ids.      |
```

Kerollmops (approved):
I like it like that 👯

PR #189, opened 2022-10-18
API Keys - Introduce `*` wildcard char at the last position to express access to multiple indexes starting with the same string

gmourier:
🤖  [API Diff]()

---

# Summary

It's now possible to express access to multiple indexes starting by the same pattern in an API Key's `indexes` field by using the `*` char as a wildcard. The pattern is only considered when the `*` character is at the last position.

e.g `products_*` match all indexes starting by `products_`

Related to https://github.com/meilisearch/meilisearch/issues/2788

---

# Changes

N/A

# Out Of Scope

N/A

---

# Attention To Reviewers

N/A

---

## Misc


curquiza (dismissed):
I approve

⚠️ I cannot guarantee it will be implemented in v0.30.0 since it's done by a contributor for hacktoberfest and we are lacking of time

curquiza:
@gmourier will probably not be ready for v0.30.0

curquiza:
@gmourier: confirmation, will not be in v0.30.0 but in v1! Sorry for the confusion! 🙇

gmourier:
@Kerollmops, I think it misses the error regarding the invalid index uid format in the `POST` route of an API Key. Could you check that, please?

curquiza (approved):

PR #209, opened 2022-12-26
Reject master keys shorter than 16 bytes in `production`

dureuill:
No API change

---

# Summary

Update specs after the changes of https://github.com/meilisearch/meilisearch/issues/3272.

Related to https://github.com/meilisearch/meilisearch/pull/3274.

---

# Changes

- Specify that too short a Master Key is a cause for error

# Out Of Scope

Warnings emitted in dev mode are considered out of scope as far as the spec is regarded. This is consistent with other specifications that generally leave warnings up to the implementation.

---

# Attention To Reviewers

N/A

---

## Misc

- [] Update OpenAPI specification file _(if needed; Apply the `OpenApi` label)_
- [] Update telemetry datapoints _(if needed; Apply the `Telemetry` label)_


text/0085-api-keys.md
```diff
@@ -41,15 +41,16 @@ At the first launch of Meilisearch with a master key, Meilisearch automatically
 
 If the master key is removed at Meilisearch launch, the previously generated API keys no longer secure the Meilisearch instance.
 
-If Meilisearch is launched with the `production` value for the `MEILI_ENV` environment variable or the `--env` CLI option, a master key is mandatory. If the master key is omitted in that particular case, Meilisearch launch is aborted and displays an error:
+If Meilisearch is launched with the `production` value for the `MEILI_ENV` environment variable or the `--env` CLI option, a master key of at least 16 characters is mandatory. If the master key is omitted in that particular case, or is too short, Meilisearch launch is aborted and displays an error:
 
-`Error: In production mode, the environment variable MEILI_MASTER_KEY is mandatory`
+- `Error: In production mode, the environment variable MEILI_MASTER_KEY is mandatory` if the key is omitted.
```
gmourier, line 46:
```suggestion
- `Error: In production mode, you must provide a master key to secure your instance. It can be specified via the MEILI_MASTER_KEY environment variable or the --master-key launch option.`
```

Hey @dureuill, we had a [discussion](https://github.com/meilisearch/product/discussions/565) dedicated to that particular error. I'm suggesting the one from @dichotommy here!

Source: https://github.com/meilisearch/product/discussions/565#discussioncomment-4266375

gmourier (changes requested):

gmourier (changes requested):

text/0085-api-keys.md
```diff
@@ -41,15 +41,16 @@ At the first launch of Meilisearch with a master key, Meilisearch automatically
...
+- `Error: In production mode, the Master Key must be of at least 16 characters, but the provided key is only :numChars characters long` if the key is too short (`numChars` is inferred in the error message).
```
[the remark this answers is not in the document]
irevoire: I would stay `master key` personally

text/0085-api-keys.md
dureuill: Thank you @gmourier! I agree with the change. Unfortunately, the implementation PR for this has already landed. I will have to open another one.

text/0085-api-keys.md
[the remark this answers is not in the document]
dureuill: The capitalization was already inconsistent in the CLI. I chose in favor of `Master Key`, but maybe the other option was preferable! I will have to make another implementation PR to bring consistency to the phrase.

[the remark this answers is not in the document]
dureuill: See https://github.com/meilisearch/meilisearch/pull/3295

text/0085-api-keys.md
dureuill: I noticed that tamo already fixed that on main. Thanks, @irevoire ❤️

text/0085-api-keys.md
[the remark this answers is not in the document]
dureuill: Updated in specification too!

text/0085-api-keys.md
dureuill: Updated in the spec!

dureuill:
Clarified that the master key length is tested against its size in **bytes**, not **characters**

dureuill:
Thanks for the warning and error messages, @gmourier !

text/0119-instance-options.md
```diff
@@ -170,10 +171,48 @@ Sets the instance's master key, automatically protecting all routes except [`GET
...
+
+We generated a secure master key for you (you can safely copy this token):
+
+>> export MEILI_MASTER_KEY=`:suggestedMasterKey` <<
+```
+
+Providing a master key of at least 16 bytes is mandatory when `--env` is set to `production`; if it is given but too short then, Meilisearch will throw an error and refuse to launch.
+
+```
+Error: In production mode, the master key must be of at least 16 bytes, but the provided key is only `:numBytes` bytes long
```
gmourier, line 191:
```suggestion
Error: In production mode, the master key must be of at least 16 bytes, but the provided key is only `:numBytes` bytes long.
```

text/0119-instance-options.md
```diff
@@ -170,10 +171,48 @@ Sets the instance's master key, automatically protecting all routes except [`GET
...
+We generated a secure master key for you (you can safely copy this token):
+
+>> export MEILI_MASTER_KEY=`:suggestedMasterKey` <<
+```
+
+Providing a master key of less than 16 bytes when `--env` is set to `development` displays a warning message.
+
+```
+The provided master key is too short (< 16 bytes)
+A master key of at least 16 bytes will be required when switching to the production environment.
```
gmourier, line 202:
```suggestion
The provided master key is too short (< 16 bytes) and thus considered unsafe for production use.
A master key of at least 16 bytes will be required when switching to the production environment.

We generated a secure master key for you (you can safely copy this token):

>> export MEILI_MASTER_KEY=`:suggestedMasterKey` <<
```

text/0119-instance-options.md
```diff
@@ -170,10 +171,48 @@ Sets the instance's master key, automatically protecting all routes except [`GET
...
+```
+
+Not providing a master key when `--env` is set to `development` displays a warning message.
+
+```
+No master key found; The server will accept unidentified requests. If you need some protection in development mode, please export a key:
+
+export MEILI_MASTER_KEY=`:suggestedMasterKey`
+
+A master key of at least 16 bytes will be required when switching to the production environment.
```
gmourier, line 212: No master key found; The server will accept unidentified requests.

A master key of at least 16 bytes will be required when switching to the production environment.

If you need some protection in development mode, we generated a secure master key for you (you can safely copy this token):

>> export MEILI_MASTER_KEY=`:suggestedMasterKey` <<

gmourier, line 212:
```suggestion
No master key found; The server will accept unidentified requests.

A master key of at least 16 bytes will be required when switching to the production environment.

If you need some protection in development mode, we generated a secure master key for you (you can safely copy this token):

>> export MEILI_MASTER_KEY=`:suggestedMasterKey` <<
```

gmourier (commented):
@dureuill @brunoocasali I've suggested some changes regarding the error and warning messages.

text/0119-instance-options.md
dureuill: Concern: users could believe that Meilisearch was started with the generated key rather than the one they provided.

This condition is unique: Meilisearch accepts to start with the provided master key that is too short, but still generates a different one. A possible interpretation could be that Meilisearch used the generated key instead of the provided one.

In a test with a fellow team member, this was their interpretation.

I feel we should specify somehow that the generated key is a suggestion to use for future launch, and maybe even that meilisearch started with the provided key.

dureuill:
@gmourier: answered inline above with a concern in the case of development mode with too short a master key.

Other changes look fine to me ❤️

text/0119-instance-options.md
gmourier: Proposal 1: Not showing the master key currently being used in the message.

```
Meilisearch started with a master key considered unsafe for production use.
	
A master key of at least 16 bytes will be required when switching to the production environment.
	
We generated a secure master key for you if you want to change it now (you can safely copy this token):
	
>> export MEILI_MASTER_KEY=`:suggestedMasterKey` <<
```

Proposal 2: Showing the master key currently being used in the message.

```
Meilisearch started with the following master key:

>> `:usedMasterKey` <<
	
The provided master key is too short (< 16 bytes) and should be considered unsafe. A master key of at least 16 bytes will be required when switching to the production environment.
	
We generated a secure master key for you if you want to change it now (you can safely copy this token):
	
>> export MEILI_MASTER_KEY=`:suggestedMasterKey` <<
```

Is it better?

dureuill: I think proposal 2 is a no-no, we should never print a master key that was *user-provided* to stdout.

I like proposal 1, maybe even stressing it further:

```
Meilisearch started with a master key considered unsafe for production use.
	
A master key of at least 16 bytes will be required when switching to the production environment.
	
We generated a new secure master key for you (you can safely copy this token):
	
>> export MEILI_MASTER_KEY=`:suggestedMasterKey` <<

Restart Meilisearch after executing the command above to use this new and secure master key.
```

What do you think?

gmourier: > I think proposal 2 is a no-no, we should never print a master key that was user-provided to stdout.

Agree!

I like your proposal. 🪄 Should we add the final line to every message?

dureuill: > Should we add the final line to every message?

Hmm I'm not sure, especially in the error cases, since we already stop meilisearch for the user

brunoocasali: > Restart Meilisearch after executing the command above to use this new and secure master key.

I completely understand the message's purpose, but if the user does it literally, it may not work when using Docker.

Because if you run the `export` command, the ENV var will be applied to the host, and the user will still have to pass it through the docker command.
So it does not work if the user "just restart Meilisearch".

brunoocasali (approved):
I left one comment, and I like the @gmourier's suggestions

Amazing work @dureuill 🎉

gmourier:
> I completely understand the message's purpose, but if the user does it literally, it may not work when using Docker.

> Because if you run the export command, the ENV var will be applied to the host, and the user will still have to pass it > through the docker command.
> So it does not work if the user "just restart Meilisearch".

Should we give the `--master-key` options instead of the `export` command @brunoocasali to overcome that?

gmourier:
I agree with you @dureuill; we can favor the `--master-key` option instead of the export one. One solution could be to display both also... @brunoocasali WDYT?

If we don't have a proper answer before wednesday, we will move to the `--master-key` option @dureuill

brunoocasali:
@dureuill made excellent observations about it.

I think the `--master-key` option is the best because it will work across all the environments as Louis pointed out. And even if the user does not like the option, they will have to review the docs to check another option to use it (config file, env var, or keep the CLI).

Could we also have some opinions from the docs team? Because maybe by changing the way the message is written, we could avoid misunderstandings.

gmourier:
Hi @maryamsulemani97! Can you check if the text messages are understandable and correct from the doc team's POV? It would be super helpful 🙇‍♂️ Let me know if I should ask someone else

text/0119-instance-options.md
```diff
@@ -170,10 +171,59 @@ Sets the instance's master key, automatically protecting all routes except [`GET
...
+
+>> export MEILI_MASTER_KEY=`:suggestedMasterKey` <<
+
+Restart Meilisearch after executing the command above to use this new and secure master key.
+```
+
+Not providing a master key when `--env` is set to `development` displays a warning message.
+
+```
+No master key found; The server will accept unidentified requests.
```
maryamsulemani97, line 215:
```suggestion
No master key was found. The server will accept unidentified requests.
```

text/0119-instance-options.md
```diff
@@ -170,10 +171,59 @@ Sets the instance's master key, automatically protecting all routes except [`GET
...
+
+A master key of at least 16 bytes will be required when switching to the production environment.
```
maryamsulemani97, line 217:
```suggestion
A master key of at least 16 bytes will be required when switching to production mode.
```

maryamsulemani97 (commented):
Made a few minor suggestions 🪴

text/0119-instance-options.md
```diff
@@ -170,10 +171,59 @@ Sets the instance's master key, automatically protecting all routes except [`GET
...
+
+If you need some protection in development mode, we generated a secure master key for you (you can safely copy this token):
```
maryamsulemani97, line 219:
```suggestion
If you need protection in a development environment, we generated a secure master key for you (you can safely copy this token):
```

gmourier:
Thank you, @maryamsulemani97, for your review! I've integrated all the suggested changes.

@dureuill, @brunoocasali I've changed the `export MEILI_MASTER_KEY=` by `--master-key`

text/0119-instance-options.md
```diff
@@ -170,10 +171,59 @@ Sets the instance's master key, automatically protecting all routes except [`GET
...
+```
+Meilisearch started with a master key considered unsafe for use in a production environment.
+
+A master key of at least 16 bytes will be required when switching to a production environment.
+
+We generated a new secure master key for you (you can safely use this token):
+
+>> --master-key `:suggestedMasterKey` <<
+
+Restart Meilisearch after executing the command above to use this new and secure master key.
```
dureuill, line 209: Late change notice ...

since this is an additional flag and not a command to execute this sentence should change.

```suggestion
Restart Meilisearch with the argument above to use this new and secure master key.
```

text/0119-instance-options.md
```diff
@@ -170,10 +171,59 @@ Sets the instance's master key, automatically protecting all routes except [`GET
...
+```
+No master key was found. The server will accept unidentified requests.
+
+A master key of at least 16 bytes will be required when switching to a production environment.
+
+If you need protection in a development environment, we generated a secure master key for you (you can safely use this token):
+
+>> --master-key `:suggestedMasterKey` <<
+
+Restart Meilisearch after executing the command above to use this new and secure master key.
```
dureuill, line 223:
```suggestion
Restart Meilisearch with the argument above to use this new and secure master key.
```

gmourier (approved):
LGTM! Thank you!

PR #210, opened 2023-01-02
Introduce the `--generate-master-key` cli-flag

irevoire:
Associated issue; https://github.com/meilisearch/meilisearch/issues/3287
Associated PR; https://github.com/meilisearch/specifications/pull/209

---

# Summary

Introduce the `--generate-master-key` cli-flag.
It generates a cryptographically safe to use master-key.

# Changes

Introduce the `--generate-master-key` cli-flag

# Attention To Reviewers

Should we really introduce an env-var + allow this parameter to be specified with the config file?

### DIFF
--- a/text/0119-instance-options.md
+++ b/text/0119-instance-options.md
@@ -468,6 +468,17 @@ Define the config file to load at Meilisearch launch.
 
 See [Configuration File](0185-configuration-file.md) specification details.
 
+
+#### 3.3.31. Generate master-key
+
+**Environment variable**: `MEILI_GENERATE_MASTER_KEY`
+**CLI option**: `--generate-master-key`
+**Default**: Disabled
+
+⚠️ This command-line option does not take any values. Assigning a value will throw an error.
+
+Generate a safe to use master key and exit without error.
+
 ## 4. Technical Aspects
 
 N/A


gmourier (approved):
LGTM!

gmourier (changes requested):
Could we indicate the message being shown when `generate-master-key` is used?

About the Attention To Reviewers questions, it doesn't shock me to keep something consistent (i.e. it can be used as env var but also in the configuration file). 

The only concern I have is that a user may be stuck without knowing how to disable the generation of a master key. One idea could be to indicate in the message that the master-key is given without launching the instance because `generate-master-key` is used. WDYT?

irevoire:
Currently, only the master-key is outputted with no text at all. This allows you to run something like `meilisearch --generate-master-key | xargs meilisearch --db-path truc --master-key`.
Or more generally, `export MEILI_MASTER_KEY=$(meilisearch --generate-master-key)`.
I don't know if it's really important.

> One idea could be to indicate in the message that the master-key is given without launching the instance because generate-master-key is used.

It's ok-ish to me.

Another idea could be to make the `--generate-master-key` incompatible with settings your master-key so people would get a better error message once they start to run their meilisearch :thinking: 

------

The last solution I can think of is to output the master-key in the logs and thus when you use the `--generate-master-key` it starts meilisearch.
But I think there is a security concern with this one, people would not be able to share their logs anymore and even storing them would get complicated :weary:

gmourier:
I've been thinking about it again, and it's perfectly ok for me to go as that for v1.0

1. This flag is a bonus. The user flow already indicates pre-generated master keys when possible.
2. We'll see if keeping it as such (just outputting a master key) is a problem depending on user feedback.

dureuill:
I'd like to call your attention on the fact that there are currently [discussions](https://meilisearch.slack.com/archives/C01K0CZKASZ/p1672769479169809) (internal link) about removing this flag.

gmourier:
Closing following https://github.com/meilisearch/meilisearch/pull/3308

Removal of this option proceeds of the following reasoning:

1. It is the only option that starts meilisearch and then immediately exits
2. We are unsure if we want to keep it under this form in the future or switch to a subcommand.
3. Releasing this option in v1 would make it insta-stable.
4. The option is only marginally useful, as users will be presented with freshly generated key directly in the error messages if their master key is absent/too short.
5. If we remove this option now, we can still add it back in a future v1 release. If we add it now, we won't be able to remove it in any future v1 version.

PR #211, opened 2023-01-03
Add `OFF` log level to spec

curquiza:
Add the missing `OFF` value to the logger. 

It already was present before v1, so I don't know where we should merge this. (into `main` or into `release-v1.0.0`)


gmourier:
Thank you @curquiza! Let's merge it to `release-v1.0.0`; I'm marking that spec as a catch-up.

loiclec (approved):

gmourier (approved):
Thanks!

PR #218, opened 2023-01-12
Specify the effect of the `*` value when used in the task filters

irevoire:
# Summary
https://github.com/meilisearch/meilisearch/issues/3342

Specify the effect of the `*` value when used in the task filters.

# Changes

Specify the effect of the `*` value when used in the task filters.

# Attention To Reviewers

I specified the effect of the `*` value when used in the task filters.


gmourier (approved):
LGTM! Thank you for the precision and the catch-up @irevoire! Can you link the spec to the related issue on the Meilisearch repository, please?

PR #222, opened 2023-02-20
Geosearch — Accept `null` for the `_geo` field

irevoire:
# Summary

Add a simple sentence on the fact that `null` is a valid geo field value and won't throw an error.

Implemented in https://github.com/meilisearch/meilisearch/pull/3515


gmourier (approved):
LGTM 🎉 Thanks @irevoire

PR #226, opened 2023-03-02
Add small blurb about index count / max size

dureuill:
# Summary

- In v1.1, changes were made to how we open indexes so that we can open an arbitrary number of indexes of almost arbitrary size.
- This PR specifies the new behavior of index wrt size/count.

---

# Changes

- Augment indexes API text with indication on the limits and behavior of index wrt size/count in the technical details

# Out Of Scope

N/A

---

# Attention To Reviewers

N/A

---


gmourier (approved):
Thank you, @dureuill! 🎉 Well done for lifting the limits 🧱

PR #232, opened 2023-04-27
Filter operators - `IS NULL` and `IS EMPTY` filter operators

Kerollmops:
This PR is updating the search API specification and, specifically the filter field. It explains the support of the `IS NULL` and `IS EMPTY` operators. [The original PR](https://github.com/meilisearch/meilisearch/pull/3571) is available on GitHub, and [the original product discussion](https://github.com/meilisearch/product/discussions/539) too!

# Changes

- Explain the `IS EMPTY` operator on the filter field of  the search API
- Explain the `IS NULL` operator on the filter field of  the search API

# Attention To Reviewers

_Explain clearly what reviewers should specifically look for to facilitate the review phase. Also, mention the sections that will not necessarily be taken into account if reviewed._

---

## Misc

- [] ~Update OpenAPI specification file _(if needed; Apply the `OpenApi` label)_~
- [] ~Update telemetry datapoints _(if needed; Apply the `Telemetry` label)_~

### DIFF
--- a/text/0118-search-api.md
+++ b/text/0118-search-api.md
@@ -360,7 +360,7 @@ attribute != value1 AND attribute != value2 AND ...
 - The `_geoRadius` operator selects the documents whose geographical coordinates fall within a certain range of a given coordinate. See [GeoSearch](0059-geo-search.md) for more information.
 - The `_geoBoundingBox` operator selects the documents whose geographical coordinates fall within a square described by the given coordinates. See [GeoSearch](0059-geo-search.md) for more information.
 
-###### 3.1.2.1.13 IS EMPTY
+###### 3.1.2.1.12 IS EMPTY
 
 The `IS EMPTY` operator selects the documents for which the filterable attribute exists and is empty. If the attribute doesn't exists then it is not empty and the document will not be returned. It is a postfix operator that takes an attribute name as argument.
 
--- a/text/0118-search-api.md
+++ b/text/0118-search-api.md
@@ -159,6 +159,12 @@ The grammar for the value of a filterable attribute is the same as the grammar f
 - In:
     * `attribute IN[value, value, etc.]`
     * `attribute NOT IN[value, value, etc.]`
+- IS NULL:
+    * `attribute IS NULL`
+    * `attribute IS NOT NULL`
+- IS EMPTY:
+    * `attribute IS EMPTY`
+    * `attribute IS NOT EMPTY`
 - AND: `filter AND filter`
 - OR: `filter OR filter`
 - NOT: `NOT filter`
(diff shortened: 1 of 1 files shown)


gmourier:
Thank you @Kerollmops I see an `OpenAPI` label, but I don't see any changes on the open-api.yml spec. I don't think it should since we are not describing every possible operator on the open API. We could do that in a future iteration if needed!

ahmednfwela:
One thing to note, is that currently there is no way to use the new `IS` filter with the `IN` filter
but you can do
`tag IS NULL OR tag IN ["hello", "world"]`

text/0118-search-api.md
```diff
@@ -374,6 +380,76 @@ is equivalent to:
 }
 ```
 
+###### 3.1.2.1.13 Is Empty
```
gmourier, line 383: The section should be moved after `3.1.2.1.12 Geo Search` and before `3.1.2.2. Array Syntax`

text/0118-search-api.md
```diff
@@ -374,6 +380,76 @@ is equivalent to:
...
+    "id": 3,
+    "colour": {}
+},
+{
+    "id": 4
+}]
+```
+Then the filter `colour IS EMPTY` selects the document ids `[0,2,3]` while the filter `colour IS NOT EMPTY` or `NOT colour IS EMPTY` selects the document ids `[1,4]`.
+
+###### 3.1.2.1.13 Is Null
```
gmourier, line 418:
```suggestion
###### 3.1.2.1.14 IS NULL
```

text/0118-search-api.md
gmourier, line 383:
```suggestion
###### 3.1.2.1.13 IS EMPTY
```

text/0118-search-api.md
```diff
@@ -374,6 +380,76 @@ is equivalent to:
...
+###### 3.1.2.1.13 Is Empty
+
+The `IS EMPTY` operator selects the documents for which the filterable attribute exists and is empty. If the attribute doesn't exists then it is not empty and the document will not be returned. It is a postfix operator that takes an attribute name as argument.
+
+The negated form of `IS EMPTY` can be written in two ways:
+```
+attribute IS NOT EMPTY
+NOT attribute IS EMPTY
+```
+Both forms are equivalent. They select the documents for which the attribute is not empty.
```
gmourier, line 392: Could we add a bullet points list explaining what is considered empty for an attribute?

e.g

- `""`
- `[]`
- `{}`

gmourier (changes requested):
Thank you @Kerollmops

Some comments about the document structures (section and naming about introduced parts)

text/0118-search-api.md
Kerollmops: Yup, I did a mistake here 🙊

text/0118-search-api.md
```diff
@@ -159,6 +159,12 @@ The grammar for the value of a filterable attribute is the same as the grammar f
 - In:
```
gmourier, line 159:
```suggestion
- IN:
```

text/0118-search-api.md
```diff
@@ -354,6 +360,82 @@ attribute != value1 AND attribute != value2 AND ...
 - The `_geoRadius` operator selects the documents whose geographical coordinates fall within a certain range of a given coordinate. See [GeoSearch](0059-geo-search.md) for more information.
 - The `_geoBoundingBox` operator selects the documents whose geographical coordinates fall within a square described by the given coordinates. See [GeoSearch](0059-geo-search.md) for more information.
 
+###### 3.1.2.1.13 IS EMPTY
```
gmourier, line 363:
```suggestion
###### 3.1.2.1.12 IS EMPTY
```

gmourier (changes requested):

gmourier:
Hey @ahmednfwela 👋

What do you mean by "One thing to note, is that currently there is no way to use the new IS filter with the IN filter"?

Thanks!

ahmednfwela:
@gmourier you can't use `tags IN [IS NULL, "action"]` for example which might be a common query that people expect,so I thought it would be cool to document it

gmourier:
Thank you for the precision, @ahmednfwela !

gmourier (approved):
Thank you @Kerollmops ! 🚂

macraig (approved):

PR #234, opened 2023-05-04
Specify the new fetch documents route

irevoire:
🤖  [API Diff](https://github.com/meilisearch/specifications/pull/234#issuecomment-1534727560)

---

Implemented in https://github.com/meilisearch/meilisearch/pull/3570
Associated issue: https://github.com/meilisearch/meilisearch/issues/3477

# Summary

Specify the new fetch documents route

---

# Changes

- The `GET /indexes/:uid/documents` route has been updated with a new query parameter `filter`
- A new `POST /indexes/:uid/documents/fetch` route has been created with all the parameters of the GET version  above (`limit`, `offset`, `fields`, and `filter`)
- A new error code has been introduced in case an error happens on the new `filter` fields: `invalid_document_filter`

## Misc


text/0124-documents-api.md
```diff
@@ -20,21 +20,25 @@ Documents are stored inside indexes.
 
 Manipulate documents of a Meilisearch index.
 
-- [3.1.1. `GET` - `indexes/:index_uid/documents`](#311-get---indexesindexuiddocuments)
+- [3.1.1. `GET` - `indexes/:index_uid/documents`](#311-get-indexesindexuiddocuments-and-post-indexesindexuiddocumentsfetch)
```
gmourier, line 23: Can we separate `GET - indexes/:index_uid/documents` and `POST - indexes/:index_uid/documents/fetch`? Both are mixed under the same title described by the HTTP verb `GET`, which is inaccurate

text/0124-documents-api.md
```diff
@@ -20,21 +20,25 @@ Documents are stored inside indexes.
...
+Meilisearch exposes 2 routes to get the documents:
+- GET `indexes/:index_uid/documents`
+- POST `indexes/:index_uid/documents/fetch`
 
 List all documents of a Meilisearch index.
 
 The query parameters `offset` and `limit` permit browsing through all documents of an index.
 
-Meilisearch orders documents depending on the hash of their id.
+Meilisearch orders documents depending on the order they were inserted in the db.
```
gmourier, line 41: Was the previous statement false?

text/0124-documents-api.md
```diff
@@ -75,7 +80,7 @@ Sets the maximum number of documents to be returned by the current request.
 
 ###### 3.1.1.2.3. `fields`
 
-- Type: String
+- Type: Array of Strings
```
gmourier, line 83: Not for the `GET` endpoint AFAIK. That's why separating the description of `GET indexes/:index_uid/documents` and `POST indexes/:index_uid/documents/fetch` can be helpful to handle every aspect of each endpoint.

text/0124-documents-api.md
```diff
@@ -161,6 +177,7 @@ Gives the total number of documents that can be browsed in the related index.
 - 🔴 Sending a value with a different type than `Integer` or `null` for `offset` will return a [invalid_document_offset](0061-error-format-and-definitions.md#invalid_document_offset) error.
 - 🔴 Sending a value with a different type than `Integer` or `null` for `limit` will return a [invalid_document_limit](0061-error-format-and-definitions.md#invalid_document_limit) error.
 - 🔴 Sending a value with a different type than `String` or `null` for `fields` will return a [invalid_document_fields](0061-error-format-and-definitions.md#invalid_document_fields) error.
+- 🔴 Sending a value with a different type than `filter` or `null` for `filter` will return a [invalid_document_filter](0061-error-format-and-definitions.md#invalid_document_filter) error.
```
gmourier, line 180: What does means "Sending a value with a different type than `filter`?"

gmourier (changes requested):

gmourier:
It seems that there is no telemetry added with this feature. Is this the case?

text/0124-documents-api.md
irevoire: Yep, it was the old behaviour before the v0.27

text/0124-documents-api.md
irevoire: Actually, since both routes serve the exact same purpose with the same parameter and error code, I wanted to do something similar to what we do with [the search route](https://github.com/meilisearch/specifications/blob/main/text/0118-search-api.md#3-functional-specification) and group both routes in the title.
Don't you think it would be easier to read the spec if I just updated the title? :thinking:

text/0124-documents-api.md
irevoire: Huum, yes, it is; it's not the same syntax, but the get route also accepts an array of strings.
I'm not sure I see the point of duplicating ~150 lines of specification if both routes are pretty much the same.
I'm really in favor of following what we did with the two search routes we have.

text/0124-documents-api.md
irevoire: Oops, I forgot to edit this one, I wanted to copy-paste the super long `string or array of string or array of array of string` thingy, and I wrote `filter` as a placeholder in the meantime :grimacing: 

It could be interesting to define somewhere in the spec what a filter is as a query parameter or a json body and then use it everywhere because it doesn't help the readability to copy-paste this long series of types everywhere.

irevoire:
> There is no telemetry added with this feature. Is this the case? (Asking for double checking, it's easy to forget from my experience face_exhaling)

Yes! We didn't have the time to implement it (especially since there was no telemetry initially on the GET route either).
Either we wait, and I can add it here at some point during the next week, or I'll do a second spec later.

text/0124-documents-api.md
```diff
@@ -20,21 +20,25 @@ Documents are stored inside indexes.
...
+- [3.1.1. `GET` - `indexes/:index_uid/documents`](#311-get-indexesindexuiddocuments-and-post-indexesindexuiddocumentsfetch)
 - [3.1.2. `GET` - `indexes/:index_uid/documents/:document_id`](#312-get---indexesindexuiddocumentsdocumentid)
 - [3.1.3. `POST` - `indexes/:index_uid/documents`](#313-post---indexesindexuiddocuments)
 - [3.1.4. `PUT` - `indexes/:index_uid/documents`](#314-put---indexesindexuiddocuments)
 - [3.1.5. `DELETE` - `indexes/:index_uid/documents`](#315-delete---indexesindexuiddocuments)
 - [3.1.6. `DELETE` - `indexes/:index_uid/documents/:document_id`](#316-delete---indexesindexuiddocumentsdocumentid)
 - [3.1.7. `POST` - `indexes/:index_uid/documents/delete-batch`](#317-post---indexesindexuiddocumentsdelete-batch)
 
-#### 3.1.1. `GET` - `indexes/:index_uid/documents`
+#### 3.1.1. `GET` - `indexes/:index_uid/documents` and `POST` - `indexes/:index_uid/documents/fetch`
```
guimachiavelli, line 31: An unsolicited comment/feedback/question: is it wise to have two endpoints that do almost exactly the same thing, but in slightly different ways? 

From a documentation perspective, it makes it harder for us to provide clear guidance on recommended practices and best use cases: "you can use A to do X, but you can also use B. B is slightly better for Y, but A works fine as well." This often results in users having to spend more time than they would want to on making a decision that might not be that important.

irevoire: Yeah, personally, I think we should deprecate the `GET` version of the route, same for the search.
To me, basically;
- The get route can be slightly easier to use with `curl`
- The post route is better on absolutely everything else (like sending big filters or using the array syntax)

guimachiavelli: Deprecating `GET` would be fine for us. In case this is not possible (e.g. API stability and versioning), would be to get clear confirmation from product and/or engine teams that we can recommend users to prefer `POST` and discourage `GET` usage.

IMHO, with this change `GET` becomes a net negative. Being slightly easier to use in a relatively small use-case is outweighed by the effort required in choosing between similar options.

macraig: @guimachiavelli totally fine on our side to recommend users to use `POST` instead of `GET`. We can even go so far as to warn users that `GET` will be deprecated in the near future.

guimachiavelli: Brilliant, @macraig, thanks! We'll keep this in mind when documenting these changes and add you as a one of the reviewers to ensure our recommendations are aligned with the API design's intent 👍

gmourier: @macraig I think we could deprecate it for v1.3, given the feedback we receive until then. I like the current solution proposed here, indicating users to favor the new one and that we will deprecate it in the future.

Deprecating the GET search endpoint should be treated as another project, we first need to conduct some technical research to verify if the post one can be safely used for caching mechanism

text/0034-telemetry-policies.md
```diff
@@ -179,7 +181,11 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
...
 | `matching_strategy.most_used_strategy`      | Most used word matching strategy among all search requests in this batch | `last` | `Documents Searched POST`, `Documents Searched GET` |
-| `per_document_id` | `true` if `DELETE /indexes/:indexUid/documents/:documentUid` endpoint was used in this batch, otherwise `false` | false | `Documents Deleted` |
+| `per_document_id` | `true` if `DELETE /indexes/:indexUid/documents/:documentUid` or `GET /indexes/:indexUid/documents/:doc_id` endpoint was used in this batch, otherwise `false` | false | `Documents Deleted`, `Documents Fetched GET` |
+| `per_filter` | `true` if `POST /indexes/:indexUid/documents/fetch` or `GET /indexes/:indexUid/documents/` endpoint was used with a filter in this batch, otherwise `false` | false | `Documents Fetched POST`, `Documents Fetched GET` |
+| `total_received` | Total number of request received in this batch | 325 | `Documents Fetched GET`, `Documents Fetched POST` |
+| `max_limit` | Highest value given for the `limit` parameter in this batch | 60 | `Documents Fetched POST`, `Documents Fetched GET` |
```
gmourier, line 187: You can reuse the `pagination.max_limit` that already exists for the search endpoints

text/0034-telemetry-policies.md
```diff
@@ -179,7 +181,11 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
...
+| `max_offset` | Highest value given for the `offset` parameter in this batch | 1000 | `Documents Fetched POST`, `Documents Fetched GET` |
```
gmourier, line 188: You can reuse the `pagination.offset` that already exists for the search endpoints

text/0034-telemetry-policies.md
```diff
@@ -179,7 +181,11 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
...
 | `swap_operation_number`            | The number of swap operation given in `POST /swap-indexes` API call | 2 | `Indexes Swapped` |
 | `matching_strategy.most_used_strategy`      | Most used word matching strategy among all search requests in this batch | `last` | `Documents Searched POST`, `Documents Searched GET` |
-| `per_document_id` | `true` if `DELETE /indexes/:indexUid/documents/:documentUid` endpoint was used in this batch, otherwise `false` | false | `Documents Deleted` |
+| `per_document_id` | `true` if `DELETE /indexes/:indexUid/documents/:documentUid` or `GET /indexes/:indexUid/documents/:doc_id` endpoint was used in this batch, otherwise `false` | false | `Documents Deleted`, `Documents Fetched GET` |
+| `per_filter` | `true` if `POST /indexes/:indexUid/documents/fetch` or `GET /indexes/:indexUid/documents/` endpoint was used with a filter in this batch, otherwise `false` | false | `Documents Fetched POST`, `Documents Fetched GET` |
+| `total_received` | Total number of request received in this batch | 325 | `Documents Fetched GET`, `Documents Fetched POST` |
```
gmourier, line 186: You can reuse the `requests.total_received` that already exists for some other aggregated events

text/0124-documents-api.md
irevoire: Why doesn't the same concern apply to the get documents route?

Also, since we had no analytics on the usage of the get route, are we not afraid of deprecating it _that early_ without having any idea of its usage?

(I'm just throwing some ideas; personally, I would like to deprecate both routes asap so users have the time to update their code until the v2 :joy:)

text/0034-telemetry-policies.md
```diff
@@ -44,6 +44,8 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
 | Documents Added | Aggregated event on all received requests via the `POST` - `/indexes/:indexUid/documents` route during one hour or until a batch size reaches `500Kb`. |
 | Documents Updated | Aggregated event on all received requests via the `PUT` - `/indexes/:indexUid/documents` route during one hour or until a batch size reaches `500Kb`. |
 | Documents Deleted | Aggregated event on all received requests via the `DELETE` - `/indexes/:indexUid/documents`, `DELETE` - `/indexes/:indexUid/documents/:documentId`, `POST` - `indexes/:indexUid/documents/delete-batch` routes during one hour or until a batch size reaches `500Kb`.  |
+| Documents Fetched GET | Aggregated event on all received requests via the `GET` - `/indexes/:indexUid/documents` or `GET` - `/indexes/:indexUid/documents/:doc_id` routes during one hour or until a batch size reaches `500Kb`. |
```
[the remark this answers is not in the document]
gmourier: It's indeed easier to create graphs directly from the event!

text/0124-documents-api.md
irevoire: What do you think of grouping both routes under a common « human-friendly » name like that; https://github.com/meilisearch/specifications/pull/234/commits/9af7b723789988aac793df988bac3b353e8cb665

text/0124-documents-api.md
gmourier: There is no urgency to depreciate it indeed, it is far from being a priority today, we can give ourselves time. What matters now, as pointed out by @guimachiavelli, is to eliminate the doubt among users between using GET or POST to fetch documents and thus simplify their use of the documentation and ultimately the product.

guimachiavelli: For what it's worth, we'd also be very happy if `GET` `/search` were to be deprecated.

brunoocasali: So, just to be sure are we going to deprecate the `GET index/{index_uid}/documents` right away or are we waiting until the v1.3?

gmourier: @brunoocasali We won't mark it as deprecated for v1.2, we are waiting for v1.3

open-api.yaml
```diff
@@ -1805,6 +1806,65 @@ paths:
...
+                    total: 2
+        '401':
+          $ref: '#/components/responses/401'
+        '404':
+          description: Not Found
+      parameters:
+        - $ref: '#/components/parameters/limit'
+        - $ref: '#/components/parameters/offset'
+        - $ref: '#/components/parameters/fields'
+        - $ref: '#/components/parameters/filter'
```
gmourier, line 1867: If filter has also been added on `GET /indexes/:indexUid/documents` don't forget to specify it in the route description in this file and in the markdown specification

text/0124-documents-api.md
irevoire: How would you call it, then?

text/0124-documents-api.md
brunoocasali: Should we (integrations) follow that? Cause I don't see a world where waiting til v1.3 will make much difference for the users, a deprecation notice is not the same as "this method does not work anymore. Take this exception in your face instead".

open-api.yaml
irevoire: In this sentence?

>         Get [documents](https://docs.meilisearch.com/learn/core_concepts/documents.html) by batch.

Like « get documents by batch with a filter »?

Is this really a good idea, though? It's just another way to get documents; we don't talk about the `fields` parameter we can use either, so I would say we don't need to write all the means you can get a document in the little description of the route; what do you think? :thinking:

text/0124-documents-api.md
irevoire: Once we deprecate it, it'll continue to work until the v2 of meilisearch, right?
The user will never get an exception, I think? (but maybe you could deprecate it in the integration and get rid of the method _before_ meilisearch v2 :thinking:)

text/0124-documents-api.md
```diff
@@ -49,13 +53,20 @@ Meilisearch orders documents depending on the hash of their id.
...
+##### 3.1.1.2. Parameters
 
-| Field                    | Type                     | Required |
-|--------------------------|--------------------------|----------|
-| `offset`                 | Integer / `null`         | false    |
-| `limit`                  | String / `null`          | false    |
-| `fields`                 | String / `null`          | false    |
+The following parameters need to be send as:
+- Query parameter for the `GET` - `indexes/:index_uid/documents` route.
+- Json body for the `POST `indexes/:index_uid/documents/fetch` route.
```
bidoubiwa, line 60:
```suggestion
- JSON body for the `POST `indexes/:index_uid/documents/fetch` route.
```

text/0124-documents-api.md
```diff
@@ -49,13 +53,20 @@ Meilisearch orders documents depending on the hash of their id.
...
+
+| Field                    | Type                      | Required |
+|--------------------------|---------------------------|----------|
+| `offset`                 | Integer                   | false    |
+| `limit`                  | Integer                   | false    |
+| `fields`                 | Array of Strings / `null` | false    |
+| `filter`                 | filter / `null`           | false    |
+
+In the case of the query parameter, as always, the `filter` can only be a string, while it can be either a string, an array of strings, or an array of array of strings for the json body.
```
bidoubiwa, line 69:
```suggestion
In the case of the query parameter, as always, the `filter` can only be a string, while it can be either a string, an array of strings, or an array of array of strings for the JSON body.
```

text/0124-documents-api.md
```diff
@@ -20,21 +20,25 @@ Documents are stored inside indexes.
...
 - [3.1.5. `DELETE` - `indexes/:index_uid/documents`](#315-delete---indexesindexuiddocuments)
 - [3.1.6. `DELETE` - `indexes/:index_uid/documents/:document_id`](#316-delete---indexesindexuiddocumentsdocumentid)
 - [3.1.7. `POST` - `indexes/:index_uid/documents/delete-batch`](#317-post---indexesindexuiddocumentsdelete-batch)
 
-#### 3.1.1. `GET` - `indexes/:index_uid/documents`
+#### 3.1.1. (Fetch Documents) - `GET` - `indexes/:index_uid/documents` and `POST` - `indexes/:index_uid/documents/fetch
+
+Meilisearch exposes 2 routes to get the documents:
+- GET `indexes/:index_uid/documents`, which gets its parameters as query parameters.
+- POST `indexes/:index_uid/documents/fetch`, which gets its parameters in a json payload.
```
bidoubiwa, line 35:
```suggestion
- POST `indexes/:index_uid/documents/fetch`, which gets its parameters in a JSON payload.
```

open-api.yaml
gmourier: My bad @irevoire I missed the addition of the `filter` as a parameter for the `GET /indexes/:indexUid/documents` route specification 🙇‍♂️

irevoire:
Seen it with @gmourier.
Instead of merging the get and the delete PR, we’re going to close this PR and only merge https://github.com/meilisearch/specifications/pull/236, which contains the specification for both features.

From what I see, all comments have been addressed on this PR. If I missed something or you want to add something, please do it on #236.

PR #238, opened 2023-05-10
Auto deletion of tasks

irevoire:
# Summary

Specify the auto-deletion of tasks that was introduced in meilisearch v1.2

Associated issue: https://github.com/meilisearch/meilisearch/issues/3622
Implemented in: https://github.com/meilisearch/meilisearch/pull/3693

### DIFF
--- a/text/0060-tasks-api.md
+++ b/text/0060-tasks-api.md
@@ -1010,16 +1010,9 @@ The task types are listed in decreasing order of priority:
 
 ### 2.2. Auto deletion of tasks
 
-Since Meilisearch can't store tasks forever, at some point, it needs to free up some space in its queue.
-The engine will try to delete the 100k last finished tasks upon reaching 1M total tasks stored.
+Since Meilisearch can't store tasks forever, at some point, it needs to free up some space in its queue. The engine will try to delete the 100k last finished tasks upon reaching 1M total tasks stored.
 
-That means after a batch finishes processing and right before processing the following enqueued tasks,
-Meilisearch will check the number of tasks currently written in its queue.
-If this number is more than 1M, the engine will enqueue a new task that automatically deletes
-the last 100k **finished** tasks.
-That means if there are only 2k finished tasks, only these ones will be deleted.
-And if all the tasks in the queue are still enqueued, then nothing will be deleted, and the engine
-will continue to process the next enqueued tasks.
+That means after a batch finishes processing and right before processing the following enqueued tasks, Meilisearch will check the number of tasks currently written in its queue. If this number is more than 1M, the engine will enqueue a new task that automatically deletes the last 100k **finished** tasks. That means if there are only 2k finished tasks, only these ones will be deleted. And if all the tasks in the queue are still enqueued, then nothing will be deleted, and the engine will continue to process the next enqueued tasks.
 
 ## 3. Future Possibilities
 
--- a/text/0060-tasks-api.md
+++ b/text/0060-tasks-api.md
@@ -4,7 +4,7 @@
 
 ### I. Summary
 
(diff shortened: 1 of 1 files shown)


gmourier (commented):
Thanks, @irevoire! Should we also mention the behavior of blocking the write to the task queue when Meilisearch can't delete the task and has reached it's maximum size for the task db?

irevoire:
Huum, I think I would say no since it’s completely hidden from the end user 🤔 
What do you think? Would you like to get it specified?

text/0060-tasks-api.md
```diff
@@ -1008,10 +1008,22 @@ The task types are listed in decreasing order of priority:
...
+Since Meilisearch can't store tasks forever, at some point, it needs to free up some space in its queue.
+The engine will try to delete the 100k last finished tasks upon reaching 1M total tasks stored.
+
+That means after a batch finishes processing and right before processing the following enqueued tasks,
+Meilisearch will check the number of tasks currently written in its queue.
+If this number is more than 1M, the engine will enqueue a new task that automatically deletes
+the last 100k **finished** tasks.
+That means if there are only 2k finished tasks, only these ones will be deleted.
+And if all the tasks in the queue are still enqueued, then nothing will be deleted, and the engine
+will continue to process the next enqueued tasks.
```
guimachiavelli, line 1022: It's usually best not to have hard line breaks in markdown, as these often translate to `<br />`s. Not super important, but it can cause problems if we ever want to generate HTML pages, PDFs, or etc for the specs.

```suggestion
Since Meilisearch can't store tasks forever, at some point, it needs to free up some space in its queue. The engine will try to delete the 100k last finished tasks upon reaching 1M total tasks stored.

That means after a batch finishes processing and right before processing the following enqueued tasks, Meilisearch will check the number of tasks currently written in its queue. If this number is more than 1M, the engine will enqueue a new task that automatically deletes the last 100k **finished** tasks. That means if there are only 2k finished tasks, only these ones will be deleted. And if all the tasks in the queue are still enqueued, then nothing will be deleted, and the engine will continue to process the next enqueued tasks.
```

text/0060-tasks-api.md
```diff
@@ -4,7 +4,7 @@
 
 ### I. Summary
 
-This specification describes the API endpoints for handling asynchronous tasks.
+This specification describes the behaviour of the task queue and the API endpoints for handling asynchronous operations.
```
guimachiavelli, line 7: Best to stick to US spelling for consistency. 

```suggestion
This specification describes the behavior of the task queue and the API endpoints for handling asynchronous operations.
```

text/0060-tasks-api.md
irevoire: Isn't the double `\n` supposed to be changed as a `<br>`? I usually just put a newline when it gets hard to read on my screen because it doesn't wrap in a nice way :sweat_smile:

guimachiavelli: Markdown doesn't have an unambiguous and universally agreed upon spec, so it's hard to describe what will happen in every parser. In my experience, though:

```
lorem ipsum

dolor sit amet
```

Generates two paragraphs:

```html
<p>lorem ipsum</p>
<p>dolor sit amet</p>
```

While:

```
lorem ipsum
dolor sit amet
```

Becomes:

```html
<p>lorem ipsum <br /> dolor sit amet</p>
```

That `<br />` tends to create issues both because it forces line breaks that look awkward depending on screen size, and because it introduces ambiguity. Like, if two sentences are in the same paragraph, they relate to the same subject; if two sentences are in different paragraphs, they're speaking about different things. With a forced line break it's sort of in between, neither here nor there.

gmourier (approved):
Thanks @irevoire

macraig (approved):

PR #251, opened 2023-07-03
Define fields to search on at runtime 

ManyTheFish:
# Summary

Update specifications related to the [Define fields to search on at runtime](https://github.com/meilisearch/meilisearch/issues/3772) feature.

---

# Changes

- Add the new search parameter definition
- Add related error messages
- Update OpenAPI

### DIFF
--- a/open-api.yaml
+++ b/open-api.yaml
@@ -719,7 +719,7 @@ components:
         attributesToSearchOn:
           type: array
           description: Defines which `searchableAttributes` the query will search on.
-          default: 'null'
+          default: '["*"]'
         filter:
           $ref: '#/components/schemas/filter'
         facets:
--- a/open-api.yaml
+++ b/open-api.yaml
@@ -716,6 +716,10 @@ components:
           type: string
           description: Defines which strategy to use to match the query terms within the documents as search results. Two different strategies are available, `last` and `all`. By default, the `last` strategy is chosen.
           default: 'last'
+        attributesToSearchOn:
+          type: array
+          description: Defines which `searchableAttributes` the query will search on.
+          default: 'null'
         filter:
           $ref: '#/components/schemas/filter'
         facets:
--- a/text/0034-telemetry-policies.md
+++ b/text/0034-telemetry-policies.md
@@ -121,6 +121,7 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
 | `filter.with_geoRadius`                 | `true` if the filter rule `_geoRadius` was used in this batch, otherwise `false` | false | `Documents Searched POST`, `Documents Searched GET` |
 | `filter.with_geoBoundingBox`            | `true` if the filter rule `_geoBoundingBox` was used in this batch, otherwise `false`| false | `Documents Searched POST`, `Documents Searched GET` |
 | `filter.most_used_syntax`               | Most used filter syntax among all requests containing the `filter` parameter in this batch | string | `Documents Searched POST`, `Documents Searched GET` |
+| `attributes_to_search_on.total_number_of_uses`| Total number of queries where `attributesToSearchOn` is set | 5 | `Documents Searched POST`, `Documents Searched GET` |
 | `q.max_terms_number`                    | Highest number of terms given for the `q` parameter in this batch | 5 | `Documents Searched POST`, `Documents Searched GET` |
 | `pagination.max_limit`                  | Highest value given for the `limit` parameter in this batch | 60 | `Documents Searched POST`, `Documents Searched GET`, `Documents Fetched GET`, `Documents Fetched POST` |
 | `pagination.max_offset`                 | Highest value given for the `offset` parameter in this batch | 1000 | `Documents Searched POST`, `Documents Searched GET`, `Documents Fetched GET`, `Documents Fetched POST` |
@@ -267,6 +268,7 @@ This property allows us to gather essential information to better understand on 
 | filter.with_geoBoundingBox | Does the built-in filter rule _geoBoundingBox has been used in the aggregated event?| `false` |
 | filter.avg_criteria_number | The average number of filter criteria among all the requests containing the `filter` parameter in the aggregated event. `"filter": []` equals to `0` while not sending `filter` does not influence the average in the aggregated event. | `4` |
 | filter.most_used_syntax | The most used filter syntax among all the requests containing the requests containing the `filter` parameter in the aggregated event. `string` / `array` / `mixed` | `mixed` |
+| attributes_to_search_on.total_number_of_uses| Total number of queries where `attributesToSearchOn` is set | `5` |
 | q.max_terms_number | The maximum number of terms for the `q` parameter among all requests in the aggregated event. | `5` |
 | pagination.max_limit | The maximum limit encountered among all requests in the aggregated event. | `20` |
 | pagination.max_offset | The maximum offset encountered among all requests in the aggregated event. | `1000` |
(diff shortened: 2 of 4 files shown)


text/0118-search-api.md
```diff
@@ -912,6 +913,20 @@ The documents containing ALL the query words (i.e. in the `q` parameter) are ret
...
+
+- Type: Array of String (POST) | String (GET)
+- Required: False
+- Default: `null`
+
+Defines which `searchableAttributes` the query will search on.
+
+- If `attributesToSearchOn` is not set or set to `null`, then the query will search on all `searchableAttributes`.
+- Sending the attributes in a different order than the order set in the settings `searchableAttributes` doesn't reorder the fields' rank for the `Attributes` ranking rule
+- 🔴 Sending a value with a different type than `Array of String`(POST), `String`(GET) or `null` for `attributesToSearchOn` returns an [invalid_attributes_to_search_on](0061-error-format-and-definitions.md#invalid_search_attributes_to_search_on) error.
```
macraig, line 926: What happens if you send an empty array?

macraig (commented):
Can we add telemetry to know when a user is searching using `attributesToSearchOn`?

text/0118-search-api.md
ManyTheFish: Empty result

macraig: Should that be specified or is it common practice in the API?

gmourier:
Don't we want a telemetry point to see if the feature is used?

text/0118-search-api.md
ManyTheFish: I think that defining an empty attributesToSearchOn list means that you search in no attributes, which implies no results. Do you see any other behavior?

ManyTheFish:
@gmourier @macraig about telemetry, what about something like:
```json
"attributes_to_search_on": {
    "total_number_of_use": 60,
}
```

text/0118-search-api.md
macraig: I guess not, I was just wondering if it might be surprising for a user to leave it empty and then be confused about getting no results

macraig:
> @gmourier @macraig about telemetry, what about something like:
> 
> ```json
> "attributes_to_search_on": {
>     "total_number_of_use": 60,
> }
> ```

@ManyTheFish how does `total_number_of_use` work?

open-api.yaml
```diff
@@ -716,6 +716,10 @@ components:
           type: string
           description: Defines which strategy to use to match the query terms within the documents as search results. Two different strategies are available, `last` and `all`. By default, the `last` strategy is chosen.
           default: 'last'
+        attributesToSearchOn:
+          type: array
+          description: Defines which `searchableAttributes` the query will search on.
+          default: 'null'
```
bidoubiwa, line 722: Why is the default value here not `["*"]` ? Like attributesToRetrieve?
`null` seems counterintuitive in its usage

text/0118-search-api.md
gmourier: I'm more on the side of the error if the array is sent with an empty form, so the developer can't break the search and is aware of it; By doing this, we can catch it before they have to understand what's wrong and fix it; I see no point in returning 0 results, am I wrong?

bidoubiwa: Maybe we could return the hits as if the query was empty?

ManyTheFish: > I'm more on the side of the error if the array is sent with an empty form, so the developer can't break the search and is aware of it; By doing this, we can catch it before they have to understand what's wrong and fix it; I see no point in returning 0 results, am I wrong?

I agree about the fact that would be useless to set `[]`, however, it's a logical behavior from Meilisearch in my sense. Moreover, it's consistent with the [`searchableAttributes` that accepts an empty array](https://github.com/meilisearch/specifications/blob/main/text/0123-searchable-attributes-setting-api.md#3321-request-payload-definition)

> Maybe we could return the hits as if the query was empty?

It doesn't seems logical to me to replace the search with a placeholder one because `attributeToSearchOn` is empty. 🤔

open-api.yaml
ManyTheFish: Changed in rc1

ManyTheFish, line 722:
```suggestion
          default: '["*"]'
```

text/0118-search-api.md
```diff
@@ -912,6 +913,20 @@ The documents containing ALL the query words (i.e. in the `q` parameter) are ret
 
 Only the documents containing ALL the query words (i.e. in the `q` parameter) are returned by Meilisearch. If Meilisearch doesn't have enough documents to fit the requested `limit`, it returns the documents found without trying to match more documents.
 
+#### 3.1.17. `attributesToSearchOn`
+
+- Type: Array of String (POST) | String (GET)
+- Required: False
+- Default: `null`
```
ManyTheFish, line 920:
```suggestion
- Default: `["*"]`
```

text/0118-search-api.md
```diff
@@ -912,6 +913,20 @@ The documents containing ALL the query words (i.e. in the `q` parameter) are ret
...
+
+Defines which `searchableAttributes` the query will search on.
+
+- If `attributesToSearchOn` is not set or set to `null`, then the query will search on all `searchableAttributes`.
```
ManyTheFish, line 924:
```suggestion
- If `attributesToSearchOn` is not set, set to `["*"]` or set to `null`, then the query will search on all `searchableAttributes`.
```

text/0034-telemetry-policies.md
```diff
@@ -121,6 +121,7 @@ The collected data is sent to [Segment](https://segment.com/). Segment is a plat
 | `filter.with_geoRadius`                 | `true` if the filter rule `_geoRadius` was used in this batch, otherwise `false` | false | `Documents Searched POST`, `Documents Searched GET` |
 | `filter.with_geoBoundingBox`            | `true` if the filter rule `_geoBoundingBox` was used in this batch, otherwise `false`| false | `Documents Searched POST`, `Documents Searched GET` |
 | `filter.most_used_syntax`               | Most used filter syntax among all requests containing the `filter` parameter in this batch | string | `Documents Searched POST`, `Documents Searched GET` |
+| `attributes_to_search_on.total_number_of_uses`| Total number of queries where `attributesToSearchOn` is set | 5 | `Documents Searched POST`, `Documents Searched GET` |
```
ManyTheFish, line 124: Hello @macraig, here is the explanation of how `attributes_to_search_on.total_number_of_uses` works

text/0118-search-api.md
macraig: +1 @gmourier , let's be consistent and revisit this when we're ready for a v2

text/0061-error-format-and-definitions.md
```diff
@@ -1808,6 +1808,49 @@ HTTP Code: `400 Bad Request`
...
+```json
+{
+    "message": "Attribute `:value` is not searchable. Available searchable attributes are: `:searchableAttributes`.",
+    "code": "invalid_search_attributes_to_search_on",
+    "type": "invalid_request",
+    "link": "https://docs.meilisearch.com/errors#invalid_search_attributes_to_search_on"
+}
+```
+
+#### Variant: one of the values is not part of the settings `searchableAttributes` list and not all of the `searchableAttributes` are displayable
```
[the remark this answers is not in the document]
ManyTheFish: Hello @gmourier, I'm not sure to understand your question. 
This variant is the same as the previous one, we will **never** show an attribute name that is not displayable, but if there are some attributes that are searchable and not displayable we will elude them by printing `, <..hidden-attributes>` at the end of the list. This way, the User will not be able to see attributes that can't be shown, however, he will have the information that the list is not exhaustive.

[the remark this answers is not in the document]
gmourier: Okay! I was trying to say: won't it always be the last variant described here that will always be used, thus rendering the other unnecessary (the one line 1832) to be described in the spec?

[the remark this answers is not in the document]
gmourier: Okay, I get it thanks @ManyTheFish !

bidoubiwa (approved):

brunoocasali (approved):

PR #267, opened 2024-02-26
Add `gpu_enabled` to telemetry

dureuill:
🤖  [API Diff]() _Put the link of the GitHub comment generated by bump.sh if generated; Apply the `OpenApi` label_

---

# Summary

cf title

---

# Changes

cf title

# Out Of Scope

N/A

---

# Attention To Reviewers

N/A

---

## Misc

- [x] Update telemetry datapoints _(if needed; Apply the `Telemetry` label)_


irevoire (commented):
There is a second place where you’re supposed to add it.

Try to search for `infos.experimental_reduce_indexing_memory_usage` for example and it should appear in two places

dureuill:
nice catch @irevoire , I don't cope well with DRY violations 👀 

btw funny, the `infos.experimental_reduce_indexing_memory_usage` was also appearing only once, had to search for `infos.experimental_enable_metrics` instead.

I fixed that as well...

irevoire:
😬 
Yeah, I feel like that’s a part of the specification we should either simplify or get rid of.
Plus, the documentation already maintains a **single** list of all the telemetry we send: https://www.meilisearch.com/docs/learn/what_is_meilisearch/telemetry#exhaustive-list-of-all-collected-data

macraig (approved):

